PR #91, opened 2015-02-27
Clean-up + Bug fix for the loop nodes

### LINKED ISSUE #90: Clean-up Loops
I think loops should receive a little clean-up.

For now, there are 3 loop nodes in the add menu. But actually, only one of them is really required: **The loop caller**. Since the other nodes you need for a loop (**object loop** and **generic loop**) can be created directly from the **loop caller** node by pressing **New Loop**.

I think it would make more sense to rename *"loop caller"* to just *"loop"* in the add menu and remove **object loop** and **generic loop** from the add menu. 

I think this could decrease confusion for new users and make the add menu a little less crowded :wink: 

----------

PS: Since I heard you were not going to actively develop this add-on anymore. I am willing to look into this if you want.

azerupi:
Fixes #90 

This pull request contains commits that:

- clean up the add menu to have 1 loop node (previously *loop caller*)
- automatically change the sockets of the loop caller in function of the selected loop
- prevent the selected loop from changing when a new loop is added

I propose you test this before you accept it so that I don't introduce bugs in master :)

### DIFF
--- a/nodes/mn_node_list.py
+++ b/nodes/mn_node_list.py
@@ -1,127 +1,125 @@
 def getNodeNameDictionary():
-	nodes = []
-	
-	nodes.append(("Number", [
-		("mn_FloatInputNode", "Float"),
-		("mn_IntegerInputNode", "Integer"),
-		("mn_FloatListInputNode", "Float List"),
-		("mn_RandomNumberNode", "Random"),
-		("mn_FloatWiggle", "Wiggle"),
-		("mn_AnimateFloatNode", "Animate"),
-		("mn_FloatMathNode", "Math"),
-		("mn_FloatClamp", "Clamp") ] ))
-		
-	nodes.append(("Vector", [
-		("mn_CombineVector", "Combine Vector"),
-		("mn_SeparateVector", "Separate Vector"),
-		("mn_RandomVectorNode", "Random Vector"),
-		("mn_VectorWiggle", "Vector Wiggle"),
-		("mn_VectorLengthNode", "Vector Length"),
-		("mn_AnimateVectorNode", "Animate Vector"),
-		("mn_VectorDistanceNode", "Vector Distance"),
-		("mn_DirectionToRotation", "Direction to Rotation"),
-		("mn_TransfromVector", "Transform Vector"),
-		("mn_VectorMathNode", "Vector Math") ] ))	
-		
-	nodes.append(("Text", [
-		("mn_StringInputNode", "Text"),
-		("mn_StringListInputNode", "Text List"),
-		("mn_RandomStringNode", "Random Text"),
-		("mn_TextBlockReader", "Text Block Reader"),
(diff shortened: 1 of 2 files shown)


azerupi:
There is still a bug:
When you add a new loop caller via a loop node with that + icon. The activeLoop = the loop you created it with. But the dropdown shows another loop...

JacquesLucke:
I currently have a small wlan problem what makes some things more difficult. 
It looks like you accidently changed all tabs to whitespaces.. that makes it very hard to see the differences in your commits. Can you maybe convert them back to tabs?

azerupi:
Yes sorry for that. But my editor was drawing warnings everywhere because PEP8 says the way to indent python code is with spaces https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces So I let my editor convert everything so that he would be happy. 

The changes:

**First Commit**
```python
   nodes.append(("System", [
       ("mn_LoopCallerNode", "Loop Caller"),
       ("mn_LoopStartNode", "Generic Loop", {"preset" : repr("NONE")}),
       ("mn_LoopStartNode", "Object Loop", {"preset" : repr("OBJECT")}),
```
to
```python
   nodes.append(("System", [
        ("mn_LoopCallerNode", "Loop"),
```

**Second Commit**
```python
selectedLoop = bpy.props.EnumProperty(items = getStartLoopNodeItems, name = "Selected Loop")
```
to
```python
def updateActiveL(self, context):
        self.updateActiveLoop()

selectedLoop = bpy.props.EnumProperty(
        items=getStartLoopNodeItems,
        name="Selected Loop",
        update=updateActiveL)
```

**Third Commit**

No problem with that one it's readable :)

azerupi:
I managed to fix the bug I mentioned earlier :)

azerupi:
I think I finally got it to work the way I want :) I don't see any bugs in the loop nodes anymore

azerupi:
@JacquesLucke Have you had a chance to test it yet?

JacquesLucke:
No sorry, no chance yet. I will let you know!

JacquesLucke:
I will check it now. I'm sorry but before I can merge it you have to try to replace the whitespaces with tabs again.. :/ for consistency reasons. Or you say that I should do it after the merge...

thanks for you work btw :)

JacquesLucke:
And you should remove the code you commented out.

Sometimes I still get an weird behavior. I select a loop name from the drop down menu. the Active loop changes to the one I clicked, but the enum property still shows another name. do you have this too?

azerupi:
Ok, well what I'm going to do is delete this pull request. Start from your latest commit and just make the changes I did. It will be cleaner that way. As for the odd behaviour, I will check. Can you maybe point out after what manipulation that happened?

JacquesLucke:
sure, just add a loop call node and two object loops. then switch the name in the drop down menu for multiple times.. sometimes I get the described behavior

Issue #95, opened 2015-03-06
Some Feature Requests


LeonCheung:
(A) Split lines
1. Adding suport to Split by BlankLine, this is very useful for some imported format.
2. Allowing using list for "Split By" input, or to support splitting by multiple rules, (and also support additional rule upon the presets, for example, to split by lines AND other one or two specified characters)

(B) Expost Input for LIST nodes:
Currently we can only type values in, not input from other nodes, which limit its power a bit.

azerupi:
For (A) this is about the "*split text*" node right?
I don't know how it is implemented but it would be simple to make that node super powerful by making it possible to split using a regexp. If I have time I can look into it :)

LeonCheung:
Hi azerupi, cool! Yes, that is about rule definition especially for texts loaded by Text Block Reader.

azerupi:
So, I  changed the Custom to Regexp. The layout was and is still not very slick, but it's working. The only drawback of this feature is that users have to know regular expressions. Can you test it and give me some feedback? https://github.com/azerupi/animation-nodes/tree/split-text

**Some examples:**

- Split on blank lines: ```\n```
- Split on dots: ```\.``` (you have to escape the dot because it has a special meaning in regular expressions)
- Split on vowels: ```[aeiou]``` or ```a|e|i|o|u``` (between the letters are not "L"s but vertical lines)

If you know how to use regular expressions you can do some great stuff ;)

LeonCheung:
Yeah, it would be definitely great if regex is supported. LuckilyI did know some regex, will give it a try!  :D

azerupi:
I am also considering an option for keeping the delimiters in the final list :)

LeonCheung:
So far so good, it gives us much more possibilities for Animation Nodes! Thank you so much! Just one comment:

When leaving regex field blank, the result behaves exactly the same as when use "Character" from the list, is that make by design? If so, I bet "Character" can retire.  :). But I would suggest to keep it and try to leave the block un-split when the regex field is blank.  FYI

azerupi:
Yes, that behaviour was already there before, I just left it like that because I didn't know anything better to do when it was blank ;) But I guess returning a one item list is not a bad idea. Thank you for your feedback :)

LeonCheung:
[...]

PR #96, opened 2015-03-06
Added regexp based splitting in the Split text node

azerupi:
In the **split text** node I changed the Custom mode to Regexp mode.

Instead of splitting by plain string comparison the split is now regexp based. In the "split by" the user can input a regexp. 

I also added a checkbox to keep the delimiters in the list. 
**Example:** ```Dummy text``` split at "**u**" 

- without the option checked would give ```["D","mmy text"]```
- with the option checked gives ```["D","u","mmy text"]```

This feature was requested here #95

### DIFF
--- a/nodes/text/mn_split_text.py
+++ b/nodes/text/mn_split_text.py
@@ -18,7 +18,7 @@ class mn_SplitText(Node, AnimationNode):
 		self.setHideProperty()
 	
 	splitType = bpy.props.EnumProperty(name = "Split Type", default = "Regexp", items = splitTypes, update = splitTypeChanges)
-	keepDelimiters = bpy.props.BoolProperty(default=False)
+	keepDelimiters = bpy.props.BoolProperty(default = False)
 	
 	def init(self, context):
 		forbidCompiling()
@@ -30,7 +30,7 @@ class mn_SplitText(Node, AnimationNode):
 		
 	def draw_buttons(self, context, layout):
 		layout.prop(self, "splitType", text = "Type")
-		if self.splitType == "Regexp": layout.prop(self, "keepDelimiters", text="Keep Delimiters")
+		if self.splitType == "Regexp": layout.prop(self, "keepDelimiters", text = "Keep Delimiters")
 		
 	def setHideProperty(self):
 		self.inputs["Split By"].hide = not self.splitType == "Regexp"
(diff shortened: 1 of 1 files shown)


nodes/text/mn_split_text.py
```diff
@@ -17,7 +17,8 @@ class mn_SplitText(Node, AnimationNode):
 	def splitTypeChanges(self, context):
 		self.setHideProperty()
 	
-	splitType = bpy.props.EnumProperty(name = "Split Type", default = "Custom", items = splitTypes, update = splitTypeChanges)
+	splitType = bpy.props.EnumProperty(name = "Split Type", default = "Regexp", items = splitTypes, update = splitTypeChanges)
+	keepDelimiters = bpy.props.BoolProperty(default=False)
```
JacquesLucke: maybe add 2 whitespaces around the equal sign, just to be consistent :)

nodes/text/mn_split_text.py
```diff
@@ -29,9 +30,11 @@ def init(self, context):
 		
 	def draw_buttons(self, context, layout):
 		layout.prop(self, "splitType", text = "Type")
+		if self.splitType == "Regexp": layout.prop(self, "keepDelimiters", text="Keep Delimiters")
```
JacquesLucke: same here

nodes/text/mn_split_text.py
```diff
@@ -42,11 +45,16 @@ def getOutputSocketNames(self):
...
 		elif self.splitType == "Words": textList = text.split()
 		elif self.splitType == "Lines": textList = text.split("\n")
-		elif self.splitType == "Custom":
-			if splitBy == "": textList = list(text)
-			else: textList = text.split(splitBy)
+
+		elif self.splitType == "Regexp":
+			if splitBy == "": textList = [text]
+			else: 
+				if self.keepDelimiters == True: textList = re.split("("+splitBy+")", text)
```
JacquesLucke: you don't have to check '== True' for boolean variables.

JacquesLucke:
so I commented on a few lines ( did this the first time ) :D
maybe you can change this rather small clean code stuff and then I will merge

nodes/text/mn_split_text.py
azerupi: I agree for consistency, but it's against PEP 8 guidelines. Maybe one day you (or me or someone else) should reformat the code to adhere to PEP 8? :)

JacquesLucke: yeah maybe some day. :D I also think there are tools which can do it partially. Personally I find it cleaner, but of course this has something to do with that I'm used to it
but consistency goes first

nodes/text/mn_split_text.py
azerupi: Yeah that's true. I can remove it if you want. Personally I don't mind and sometimes find it clearer to read :)

azerupi:
I have made the tweaks you asked for :)

PR #123, opened 2015-05-07
'projected' mode in revolve 

guy-lateur:


nodes/curves/Curves.py
```diff
@@ -18,6 +18,17 @@ def IsBezierCurve(blenderObject):
         if spline.type != 'BEZIER': return False
 
     return True
+    
+deltaParameter = 0.001
+def ParameterIsZero(parameter):
+    if parameter < deltaParameter: return True
```
JacquesLucke: You could also write: return parameter < deltaParameter
to shorten it a bit

nodes/curves/Curves.py
```diff
@@ -18,6 +18,17 @@ def IsBezierCurve(blenderObject):
...
+    
+    return False
+
+def ParameterIsOne(parameter):
+    if parameter + deltaParameter > 1.0: return True
```
JacquesLucke: same here

guy-lateur: Done and done. 
I don't think I should create a new pull request for this, right? I did commit and sync my code.

JacquesLucke:
I'm testing your changes locally currently.

One question: why are these two functions needed at all?  (ParameterIsZero,etc)

JacquesLucke:
Also 'Fast' and 'Projected' mode seem to do totally different things or am I wrong?
Maybe another word for "fast" would be better here. Or you put a very short description in the description of this item.

guy-lateur:
Those functions are needed to account for possible numerical/round-off errors. I've had bad experiences in the past with comparing floating values directly (if x == 1.0). If I get back a parameter of 0.9999, I want to be able to treat that as a 1, ie, the end point of the curve. This will also be useful in future developments.

I agree 'Fast' is probably not the best name for the first/original mode of the revolve node. I'll think about a better name/description. They're not 2 totally different things, though; they just let you choose the way you want the revolve to work, geometrically. See my discussions with o.g. on the thread on BA about this.

JacquesLucke:
Yes I see. 

Maybe "Projected" and "Same Parameter" or something in this direction

JacquesLucke:
otherwise this looks fine

guy-lateur:
Modes renamed to "Same Parameter" and "Project Profile".
Changed default mode to Project Profile, because that seems to be more familiar to users.

JacquesLucke:
Yes. It is in master now :)

guy-lateur:
Great, cheers! :)

PR #129, opened 2015-05-13
curve nodes pull request

guy-lateur:
. added analytic projection
. refactored '& Morph' variants


guy-lateur:
I was able to sync to your code without reforking the whole thing.
Not sure how I did it, though.. :)

JacquesLucke:
thats good news :D
Give me some time to look over it :)

guy-lateur:
Alright, take your time.. :)

JacquesLucke:
Is there a case when the old sampled algorithm is better than the new analytic way? If not you can remove the old algorithm from the node I guess

Also I'm not sure if the '(Optional)' should be in a socket name or if we can leave it away completety

JacquesLucke:
Apart from that it looks good. 
and I'm happy that I just found a very easy way to checkout your pull request locally :D

guy-lateur:
The old sampled mode is still there to be able to compare it to the new analytic mode, for now. It may give better accuracy when using higher resolutions. It may be quite slow, though -- but still it's nice to have, as a reference measure.

The analytic way calculates the roots of a 5th order polynomial, using the numpy module. There's probably some accuracy threshold implied in there as well, although it isn't exposed, ATM. So yes, the sampled mode may disappear someday, but not until I'm satisfied about the (implied) accuracy of the analytic mode.

Yeah, I'm not sure about the '(Optional)' thing, either. Initially, I thought of having a bool property to expose the endProfile input socket, but that would make it harder to compare between having an endProfile or not. OTOH, it may be weird to have a node execute when not all input sockets are connected. So maybe having some 'optional' input sockets isn't so bad after all.. :)

I don't know. Maybe we could just go with it the way it is right now? We can always update it later.

I'm glad to hear you can now easily locally checkout my updates -- woohoo! :)

JacquesLucke:
ok, that's a good reason to leave it :)

I think I would prefer a bool property 'Use End Profile' that also influences the hide property of the socket
The problem is that every time you add/rename sockets you will break all files in which these nodes are used

PR #136, opened 2015-05-15
Update mn_sound_bake_reader.py

brookman:
Typo: Renamed frequence to frequency


JacquesLucke:
Thanks for that. I will have to check something else before merging this because there are some .blend files I don't want to break

Issue #155, opened 2015-06-15
List nodes refurbish, modifications, additions


og76:
There are 2 sections here:
**1. node modification or addition**
[I add some patches with comments. if not ok, I can post the files.] https://github.com/og76/animation_nodes
**2. code or socket issues**
[may wanna take a look there first? however these may be for future refurbish, not immediate...]

**_1.  node modification or addition_**

1.1
[add] - **trim text**  (see #150) -  have "Allow negative Index" bool 
edit: DONE

1.2
[add] **trim list**, same as text but for lists (see patch) 
https://github.com/og76/animation_nodes/commit/0bd33b0744129c842750c7962976d9b64caba702

1.3
[mod] -Negative index, same for **get element** (see patch) 
https://github.com/og76/animation_nodes/commit/c07b69e56854d1a38b20c1fe6cabda93d418a050
For set element, not.

1.4 separate **sum elements** to:
[add] - **join text list** , just that, adding start end functionality, same negative etc
https://github.com/og76/animation_nodes/commit/10b054ca6ca08011b7cf2d987849d55da8be4003
note: create list(string)+join txt looks better than current combine txt (list has management and all)

[add] -**math list elements** with enum: sum, min ,max , avg , 
https://github.com/og76/animation_nodes/commit/8725fcaf20b56c75e1e729dddc96a705cf90a483
note: is there more to add than sum, min, max, avg? 

may *keep original Sum Elements* (hidden) for backwards compatibility
it affects #64 ? 

1.5 
[add] add **sort list elements**, like the expressions we discussed #149, with inputs:
list (list sockets) // attribute (string socket?) // reverese (bool)
testing with script node example ...  I would not do it this brute, I'll test more...
*.to be discussed, this is complicated,*

1.6 
[add] **Filter list**: there is also startswith /endswith ( pattern? not quite so, it seams to be much faster this startswith, out of my reading...) 
-very usefull for string list **Text List Filter**
TO DO ...
-object list  **Object List Filter** (may add others than object)
https://drive.google.com/file/d/0BzzAoBwXO8dsOUw5a0Vjd1pBYWM/view?usp=sharing
script node example ... very brute, but useful
object filter (also with obj type, mesh etc, and has attribute?)
TO DO ...

1.7
[add] add **map lists** . with fallof, guess is better than zip. if it has start/end, we manually limit anyway testing, long discussion, if the case

1.8
[...]

PR #181, opened 2015-07-30
7 list nodes update

og76:
as per #155


JacquesLucke:
Thanks for all the work!
As you might have noticed I started a big refactor last night. Merging some of your changes may become difficult..
But there are a few things I would like to merge

I'm not sure how difficult this is but can you create another request containing only the new nodes you created?
Otherwise merging this with my refactor branch can become a bit complicated.
I could also just copy your new files into my branch. I would mention your name in the commit log of course... Do you have a better idea?

Great work!

og76:
I'll have to see how to do this. This does show all the little updates I did in the meantime.(useless)
I thought it will compare only the final ver not with all history...

-Should I make pull request for each file? (is there such a thing?)
-Should I create a new clean branch from master and just add these final 7 nodes, so that it does not show all updates? **I'll try this**
-You do have direct links to py in last post #155 

ps. combine lists is one of them, and is not new, is considered mod

PR #204, opened 2015-09-16
operations labels for math, vec math nodes Patch 2

og76:
operations labels for math, vec math

math: 
+ node op labels (also show int/float)  
+ added invert,reciproc 
vector math: 
+ node op labels (hehe, the workaround..)  
+ added project(), reflect()

### DIFF
--- a/nodes/number/float_math.py
+++ b/nodes/number/float_math.py
@@ -41,9 +41,9 @@ class FloatMathNode(bpy.types.Node, AnimationNode):
     bl_label = "Math"
 
     def operationChanged(self, context):
-        for op in operationItems:
-            if op[0] == self.operation:
-                self.labelOperation = op[2]
+        for item in operationItems:
+            if item[0] == self.operation:
+                self.labelOperation = item[2]
         self.inputs[1].hide = self.operation in singleInputOperations
         executionCodeChanged()
 
@@ -56,8 +56,8 @@ class FloatMathNode(bpy.types.Node, AnimationNode):
     outputInteger = BoolProperty(name = "Output Integer", default = False,
         update = outputIntegerChanged)
 
-    labelOperation = StringProperty(name = "labelOperation", default = "A * B", update = operationChanged)
-    labelOutputType = StringProperty(name = "labelOutputType", default = " (float)", update = outputIntegerChanged)
+    labelOperation = StringProperty(name = "Operation Label", default = "A * B", update = operationChanged)
+    labelOutputType = StringProperty(name = "Output Type Label", default = "float", update = outputIntegerChanged)
 
     def create(self):
         self.inputs.new("an_FloatSocket", "A", "a")
@@ -68,7 +68,7 @@ class FloatMathNode(bpy.types.Node, AnimationNode):
         layout.prop(self, "operation", text = "")
         
     def draw_label(self):
-        return self.labelOperation + self.labelOutputType
+        return self.labelOperation + " (" + self.labelOutputType + ")"
 
     def edit(self):
         targets = self.outputs[0].dataTargets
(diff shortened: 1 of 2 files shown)


nodes/number/float_math.py
```diff
@@ -39,6 +41,9 @@ class FloatMathNode(bpy.types.Node, AnimationNode):
     bl_label = "Math"
 
     def operationChanged(self, context):
+        for op in operationItems:
```
JacquesLucke: please don't use 'op' here. instead write "for item in operationItems"
I know that I wrote 'op' in the execution code but that was to improve readability. Here that is not the case (imo)

nodes/number/float_math.py
```diff
@@ -51,6 +56,9 @@ def outputIntegerChanged(self, context):
     outputInteger = BoolProperty(name = "Output Integer", default = False,
         update = outputIntegerChanged)
 
+    labelOperation = StringProperty(name = "labelOperation", default = "A * B", update = operationChanged)
+    labelOutputType = StringProperty(name = "labelOutputType", default = " (float)", update = outputIntegerChanged)
```
JacquesLucke: the name attribute should be good for the ui, it doesn't have to look like a variable.  "Operation Label" would be better

nodes/number/float_math.py
```diff
@@ -58,6 +66,9 @@ def create(self):
 
     def draw(self, layout):
         layout.prop(self, "operation", text = "")
+        
+    def draw_label(self):
```
JacquesLucke: uhh you're right here. In fact I didn't know about this function. We can leave it as it is now but most propably I will add here an abstraction layer as well. 
Then we will have the functions: 'draw', 'drawAdvanced' and 'drawLabel'

nodes/number/float_math.py
```diff
@@ -70,6 +81,8 @@ def edit(self):
             self.outputInteger = False
 
     def recreateOutputSocket(self):
+        self.labelOutputType = " (integer)" if self.outputInteger else " (float)" #just to be more clear
```
JacquesLucke: maybe the 'labelOutputType' variable should just hold "Integer" or "Float". The brackets and spaces should be added in the draw_label function

nodes/vector/vector_math.py
```diff
@@ -19,12 +21,17 @@ class VectorMathNode(bpy.types.Node, AnimationNode):
     bl_label = "Vector Math"
 
     def operationChanged(self, context):
+        for op in operationItems:
+            if op[0] == self.operation:
```
JacquesLucke: same with 'op' here

nodes/vector/vector_math.py
```diff
@@ -19,12 +21,17 @@ class VectorMathNode(bpy.types.Node, AnimationNode):
...
+                self.labelOperation = op[2][:11]
         self.inputs["B"].hide = self.operation in operationsWithFloat
         self.inputs["Scale"].hide = self.operation not in operationsWithFloat
         executionCodeChanged()
 
     operation = EnumProperty(name = "Operation", items = operationItems, default = "ADD", update = operationChanged)
 
+    labelOperation = StringProperty(name = "labelOperation", default = "A + B", update = operationChanged)
```
JacquesLucke: same as above

nodes/number/float_math.py
og76: do we really need an abstraction layer for this? I mean, how much should we abuse labels?
Plus, the current BL way works very well with custom labels (that prevail..), extra things to bother about.
Just wondering if it's worth the effort ..

JacquesLucke: The effort to do this is very low. 
max. 5 lines
I can never know wether we will use this more in the future. Also then this function blends better into the others. I rarely use underscores in variable and function names

PR #206, opened 2015-09-18
obj vis in / out, +filter by name (startswith..)

og76:
obj vis input:
should I delete isLinked in vis input? it's kinda too much for this little one, maybe I should let it just do straight the functions ..

ob startswith/ends:
-this is a classic simple filter used in other soft and places
-startswith/endswith are supposed to be way faster than match, so I don't include other search ways

ob vis in out
ob startswith

### DIFF
--- a/nodes/object/object_name_filter.py
+++ b/nodes/object/object_name_filter.py
@@ -1,19 +1,17 @@
 import bpy
+from bpy.props import *
 from ... events import executionCodeChanged
 from ... base_types.node import AnimationNode
+
+filterTypeItems = [("STARTS_WITH", "Starts With", "All Objects with names starting with"),
+                   ("ENDS_WITH", "Ends With", "All Objects with names ending with")]
 
 class ObjectNameFilterNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ObjectNameFilterNode"
     bl_label = "Object Name Filter"
     
-    def checkedPropertiesChanged(self, context):
-        executionCodeChanged()
-    
-    useCaseSensitive = bpy.props.BoolProperty(default = False, description = "Use Case Sensitive", update = checkedPropertiesChanged)
-    
-    filterType = bpy.props.EnumProperty(items = [("STARTS_WITH", "Starts With", "All Objects with names starting with"),
-                                                 ("ENDS_WITH", "Ends With", "All Objects with names ending with")],
-                                        update = checkedPropertiesChanged, default = "STARTS WITH")
+    useCaseSensitive = BoolProperty(default = False, description = "Use Case Sensitive", update = checkedPropertiesChanged)
+    filterType = EnumProperty(items = filterTypeItems, default = "STARTS_WITH", update = executionCodeChanged)
     
     def create(self):
         self.bl_width_default = 160
--- a/nodes/object/object_name_filter.py
+++ b/nodes/object/object_name_filter.py
@@ -2,19 +2,17 @@ import bpy
 from ... events import executionCodeChanged
 from ... base_types.node import AnimationNode
 
-class an_ObjectNameFilterNode(bpy.types.Node, AnimationNode):
+class ObjectNameFilterNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ObjectNameFilterNode"
     bl_label = "Object Name Filter"
     
     def checkedPropertiesChanged(self, context):
-        self.updateSocketVisibility()
         executionCodeChanged()
     
-    useAllInScene = bpy.props.BoolProperty(default = False, description = "Use All Objects In Scene instead of socket", update = checkedPropertiesChanged)
     useCaseSensitive = bpy.props.BoolProperty(default = False, description = "Use Case Sensitive", update = checkedPropertiesChanged)
     
-    filterType = bpy.props.EnumProperty(items = [("STARTS WITH", "Starts With", "All Objects with names starting with"),
-                                                 ("ENDS WITH", "Ends With", "All Objects with names ending with")],
+    filterType = bpy.props.EnumProperty(items = [("STARTS_WITH", "Starts With", "All Objects with names starting with"),
+                                                 ("ENDS_WITH", "Ends With", "All Objects with names ending with")],
                                         update = checkedPropertiesChanged, default = "STARTS WITH")
     
     def create(self):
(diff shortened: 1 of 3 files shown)


nodes/object/object_visibility_input.py
```diff
@@ -0,0 +1,32 @@
+import bpy
+from ... base_types.node import AnimationNode
+
+class an_ObjectVisibilityInput(bpy.types.Node, AnimationNode):
```
JacquesLucke: should be an_ObjectVisibilityInputNode 
and the same with the bl_idname

nodes/object/object_visibility_input.py
```diff
@@ -0,0 +1,32 @@
...
+    bl_idname = "an_ObjectVisibilityInput"
+    bl_label = "Object Visibility Input"
+    outputUseParameterName = "useOutput"
```
JacquesLucke: the outputUseParameterName isn't used anymore

nodes/object/object_visibility_input.py
```diff
@@ -0,0 +1,32 @@
...
+        self.outputs.new("an_BooleanSocket", "Show Name", "showName").hide = True
+        self.outputs.new("an_BooleanSocket", "Show Axis", "showAxis").hide = True
+        self.outputs.new("an_BooleanSocket", "Show Xray", "showXray").hide = True
+    
+    def getExecutionCode(self):
+
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return ""
+        lines = []
+        lines.append("hide, hideSelect, hideRender, showName, showAxis, showXray = None, None, None, None, None, None")
```
JacquesLucke: we have Booleans here. and a boolean shouldn't be None. Use False instead.
Maybe you can even write: hide = hideSelect = hideRender =...= False

nodes/object/object_visibility_input.py
```diff
@@ -0,0 +1,32 @@
...
+        lines.append("if object is not None:")
+        if isLinked["hide"]: lines.append("    hide = object.hide")
+        if isLinked["hideSelect"]: lines.append("    hideSelect = object.hide_select")
+        if isLinked["hideRender"]: lines.append("    hideRender = object.hide_render")
+        if isLinked["showName"]: lines.append("    showName = object.show_name")
+        if isLinked["showAxis"]: lines.append("    showAxis = object.show_axis")
+        if isLinked["showXray"]: lines.append("    showXray = object.show_x_ray")
+
```
JacquesLucke: The hide = hideSelect ...= False thing should be here in *else*

nodes/object/object_visibility_output.py
```diff
@@ -0,0 +1,160 @@
...
+        self.inputs.new("an_BooleanSocket", "Show Name", "showName").value = False
+        self.inputs.new("an_BooleanSocket", "Show Axis", "showAxis").value = False
+        self.inputs.new("an_BooleanSocket", "Show Xray", "showXray").value = False
+        self.outputs.new("an_ObjectSocket", "Object", "object")
+        self.updateSocketVisibility()
+        
+    def draw(self, layout):
+        rrow = layout.row()
+        
+        col = rrow.column()
```
JacquesLucke: rrow is a bad name :p
please use *row* and *subrow* in such cases

nodes/object/object_visibility_output.py
```diff
@@ -0,0 +1,160 @@
...
+        return lines
+
+
+class ShowHelp(bpy.types.Menu):
+    bl_idname = "an.show_obj_vis_help"
+    bl_label = "Object Visibility Output node | Blender - Animation Nodes"
+    bl_icon = "FORCE_TURBULENCE"
+    
+    helpText = bpy.props.StringProperty(default = "help here")
+    noteText = bpy.props.StringProperty(default = "note here")
```
JacquesLucke: bpy.props isn't needed here. (Use star import at the top)

nodes/object/object_visibility_output.py
```diff
@@ -0,0 +1,160 @@
...
+        if self.useXray: lines.append("    object.show_x_ray = showXray")
+        
+        if not any((self.useView, self.useSelect, self.useRender, self.useName, self.useAxis, self.useXray)): 
+            lines = []
+            
+        #lines.append("object = object")
+        return lines
+
+
+class ShowHelp(bpy.types.Menu):
```
JacquesLucke: Until we have a better system for help menus we leave this as it is :)
only the bl_idname should be "an.show_object_visibilty_help"

nodes/object/object_visibility_output.py
og76: I shall use rrroarr ..

nodes/object/object_name_filter.py
```diff
@@ -0,0 +1,54 @@
+import bpy
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+class ObjectNameFilterNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ObjectNameFilterNode"
+    bl_label = "Object Name Filter"
+    
+    def checkedPropertiesChanged(self, context):
+        executionCodeChanged()
```
JacquesLucke: when the update function only contains *executionCodeChanged* you don't need this extra update function. Just call executionCodeChanged directly from the property

nodes/object/object_name_filter.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    useCaseSensitive = bpy.props.BoolProperty(default = False, description = "Use Case Sensitive", update = checkedPropertiesChanged)
```
JacquesLucke: still bpy.props

nodes/object/object_name_filter.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    filterType = bpy.props.EnumProperty(items = [("STARTS_WITH", "Starts With", "All Objects with names starting with"),
+                                                 ("ENDS_WITH", "Ends With", "All Objects with names ending with")],
```
JacquesLucke: can you maybe define the items above the class?
the variable containing the items should be called *filterTypeItems*

nodes/object/object_name_filter.py
```diff
@@ -0,0 +1,54 @@
...
+                                        update = checkedPropertiesChanged, default = "STARTS WITH")
```
JacquesLucke: the default needs the underscore as well

nodes/object/object_visibility_output.py
```diff
@@ -0,0 +1,160 @@
...
+    def create(self):
+        self.inputs.new("an_ObjectSocket", "Object", "object").defaultDrawType = "PROPERTY_ONLY"
+        self.inputs.new("an_BooleanSocket", "Hide", "hide")
+        self.inputs.new("an_BooleanSocket", "Hide Select", "hideSelect")
+        self.inputs.new("an_BooleanSocket", "Hide Render", "hideRender")
+        self.inputs.new("an_BooleanSocket", "Show Name", "showName")
+        self.inputs.new("an_BooleanSocket", "Show Axis", "showAxis")
+        self.inputs.new("an_BooleanSocket", "Show Xray", "showXray")
+        self.outputs.new("an_ObjectSocket", "Object", "object")
+        for socket in self.inputs: socket.value = False
```
JacquesLucke: should be *self.inputs[1:]* because the object socket has no value that you can set to False

nodes/object/object_visibility_output.py
```diff
@@ -0,0 +1,160 @@
...
+        
+        if self.useView: lines.append("    object.hide = hide")
+        if self.useSelect: lines.append("    object.hide_select = hideSelect")
+        if self.useRender: lines.append("    object.hide_render = hideRender")
+        
+        if self.useName: lines.append("    object.show_name = showName")
+        if self.useAxis: lines.append("    object.show_axis = showAxis")
+        if self.useXray: lines.append("    object.show_x_ray = showXray")
+        
+        if not any((self.useView, self.useSelect, self.useRender, self.useName, self.useAxis, self.useXray)):
```
JacquesLucke: this check shall come first in this function. So that you immediatly return an empty list when nothing is used

JacquesLucke:
I still think you should abandon all these useName, useAxis, .... properties and also only use the linked inputs.

PR #207, opened 2015-09-23
Refac og text and list

og76:
small things :


og76:
char prop

get index

nodes/list/get_list_element_index.py
```diff
@@ -0,0 +1,66 @@
...
+        self.invokeSocketTypeChooser(layout, "assignListDataType",
+            socketGroup = "LIST", text = "Change Type", icon = "TRIA_RIGHT")
+        
+    def getExecutionCode(self):
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return ""
+        
+        lines = []
+        lines.append("allIndices = [i for i, element in enumerate(list) if element == search]")
+        if isLinked["firstIndex"]: lines.append("firstIndex = allIndices[0] if len(allIndices) > 0 else None")  #this may be the most used, so we just avoid another get element
```
JacquesLucke: the firstIndex must be an integer. Assigning *None* to the variable can cause hard to detect bugs in other places because all other nodes rely on the fact that an integer socket always gets an integer value.
Many programming languages return -1 as index when the element was not found. i think we should do the same here.

nodes/list/get_list_element_index.py
```diff
@@ -0,0 +1,66 @@
...
+        if isLinked["occurrences"]: lines.append("occurrences = len(allIndices)")  #this is also an indication that exists
```
JacquesLucke: some code comments aren't needed ;)
Quite often they just add some noise to clean code :D

nodes/text/character_props_output.py
```diff
@@ -0,0 +1,92 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+class CharacterPropsOutputNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_CharacterPropsOutputNode"
+    bl_label = "Character Props Output"
```
JacquesLucke: please use *Properties* instead of *Props* in all the 3 cases

nodes/list/get_list_element_index.py
```diff
@@ -0,0 +1,66 @@
+import bpy
+from bpy.props import *
+from ... tree_info import keepNodeLinks
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+from ... sockets.info import getBaseDataTypeItems, toIdName, toListIdName, isBase, toBaseDataType
+
+class GetListElementIndexNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetListElementIndexNode"
+    bl_label = "Get Element Index"    #Search List Element ?
```
JacquesLucke: Lets use "Search List Element" :)
Also don't forget to update the bl_label, bl_idname, class name und file name.

nodes/text/character_props_output.py
```diff
@@ -0,0 +1,50 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+class CharacterPropertiesOutputNode(bpy.types.Node, AnimationNode):
```
JacquesLucke: also update the file name from props -> properties

nodes/text/character_props_output.py
```diff
@@ -0,0 +1,50 @@
...
+        if self.allowNegativeIndex: lines.append("    s, e = start, end")
+        else: lines.append("    s, e = max(0, start), max(0, end)")
+    
+        lines.append("    for char in textObject.body_format[s:e]:")
+        if isLinked["materialIndex"]: lines.append(" "*8 + "char.material_index = materialIndex")
+        if isLinked["bold"]: lines.append(" "*8 + "char.use_bold = bold")
+        if isLinked["italic"]: lines.append(" "*8 + "char.use_italic = italic")
+        if isLinked["underline"]: lines.append(" "*8 + "char.use_underline = underline")
+        if isLinked["smallCaps"]: lines.append(" "*8 + "char.use_small_caps = smallCaps")
+        lines.append("    pass")
```
JacquesLucke: this 'pass' isn't needed here. It is only needed when after an **:** nothing indented comes.

nodes/text/character_props_output.py
```diff
@@ -0,0 +1,50 @@
...
+    
+        if self.allowNegativeIndex: lines.append("    s, e = start, end")
+        else: lines.append("    s, e = max(0, start), max(0, end)")
+    
+        lines.append("    for char in textObject.body_format[s:e]:")
+        if isLinked["materialIndex"]: lines.append(" "*8 + "char.material_index = materialIndex")
+        if isLinked["bold"]: lines.append(" "*8 + "char.use_bold = bold")
+        if isLinked["italic"]: lines.append(" "*8 + "char.use_italic = italic")
+        if isLinked["underline"]: lines.append(" "*8 + "char.use_underline = underline")
+        if isLinked["smallCaps"]: lines.append(" "*8 + "char.use_small_caps = smallCaps")
```
JacquesLucke: Can you maybe rewrite this node to use the even newer system I use for output nodes now. Look at the normal text output for example

PR #216, opened 2015-10-02
Fcurve keyframes

og76:
we need to know keyframe pos and value
I chose 2 float lists instead of 1 list of pairs so that
-we access the value directly not with yet another get el
-we don't add other sockets just for this (and not using generic)
-sometimes we just need the frame (ex start and end for the range...)

### DIFF
--- a/nodes/fcurve/fcurve_keyframes.py
+++ b/nodes/fcurve/fcurve_keyframes.py
@@ -1,0 +1,21 @@
+import bpy
+from ... base_types.node import AnimationNode
+
+class FCurveKeyframesNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_FCurveKeyframesNode"
+    bl_label = "FCurve Keyframes"
+
+    def create(self):
+        self.inputs.new("an_FCurveSocket", "FCurve", "fCurve")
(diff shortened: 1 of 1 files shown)


JacquesLucke:
Can you give me an example when this node is useful?

Also I'm pretty sure that it will raise an exception when you connect the output sockets to something while there is no fcurve in the input

Beside that, I think you can use a list comprehension here :)

og76:
examples: 
- we can set keyframes but we cannot read them. 
- can evaluate the fcurve, but cannot say the range (start/end)
- can evaluate but cannot say when a keyframe occurs

- with this u can set precise values at precise frames as keyframes instead of values related to time/frame as nodes. This way u see and move your values in the dopesheet
- with this can set triggers for any change at certain frames

however, for me, the most important part is to see them in 3d, dopesheet etc

oh, I was also doin this "motion path to spline..."


+ also, I was doing the opposite: moving keyframes or values for the instances (I gave some examples in an old issue here), but I changed my mind.
that I would leave to expressions, scripts etc cause u have to really know what you're doing. First cause it's data loss prone (no undo here) and second, there are limitations etc..

PR #232, opened 2015-10-06
Create guided_track.py

og76:
alternative to direction to rotation, that gives extra control over rotation around track /direction axis
(and simplifies a lot some complex setups for just that extra rotation)

**for poly dupli instances**, as described here by JoseConseco:
http://blenderartists.org/forum/showthread.php?350296-Addon-Animation-Nodes&p=2913920&viewfull=1#post2913920
and "solved" after some tedious discussions here:
http://blenderartists.org/forum/showthread.php?350296-Addon-Animation-Nodes&p=2914994&viewfull=1#post2914994

other uses can be to: imitate the track modifier "target Z" option or control roll if you move along a spine etc.

### DIFF
--- a/nodes/vector/guided_track.py
+++ b/nodes/vector/guided_track.py
@@ -49,7 +49,7 @@ class GuidedTrack(bpy.types.Node, AnimationNode):
         if self.trackAxis != self.guideAxis and self.trackAxis != ("-" + self.guideAxis):
             # the escapes here are needed only for the matrix, to not fall to scale 0
             lines.append("z = direction.normalized() if direction != mathutils.Vector((0, 0, 0)) else mathutils.Vector((0, 0, 1))")          #direction
-            lines.append("y = (z.cross(guide)).normalized() if guide != mathutils.Vector((0, 0, 0)) else mathutils.Vector((0, 1, 0))")
+            lines.append("y = (z.cross(guide)).normalized() if guide != mathutils.Vector((0, 0, 0)) and guide.normalized() != z else mathutils.Vector((0, 1, 0))")
             lines.append("x = y.cross(z)")  #guide as perpendicular to dir
             
             tAx = self.trackAxis
--- a/nodes/vector/guided_track.py
+++ b/nodes/vector/guided_track.py
@@ -24,8 +24,7 @@ class GuidedTrack(bpy.types.Node, AnimationNode):
     def create(self):
         self.inputs.new("an_VectorSocket", "Direction", "direction")
         self.inputs.new("an_VectorSocket", "Guide", "guide")
-        self.outputs.new("an_VectorSocket", "Rotation Vector", "rotationVector")
-        self.outputs.new("an_MatrixSocket", "Rotation Matrix", "rotationMatrix")
+        self.outputs.new("an_EulerSocket", "Rotation Euler", "euler")
         
         self.width += 20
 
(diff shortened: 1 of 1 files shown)


JoseConseco:
Works great! Please make it to master branch 🙏.
Btw is adaptive poly  would be great too.

og76:
Now it works great ...

og76:
it will be affected by #239 as euler is the output, and not a vector.

if we get an euler dedicated socket, I might quit the matrix output and clean some escapes..

JacquesLucke:
Is it ok for you when I take the node and try to clean it up a bit?

og76:
yes, but first let me get a bit in the eulers and rot menu

JacquesLucke:
Ok tell me then

og76:
now that I removed matrix it's all cleaner ...
please take a look at it.

JacquesLucke:
Will check it tomorrow.
What do you think about the idea to combine this node with the other direction to rotation node?

og76:
how? with an option for the extra guide?

JacquesLucke:
Yes

og76:
hm. 
and the code would be this one? cause it covers for it all ..
and it would practically replace the direction to rotation?

however, we have o check this first against the other one, to see if all ok.

JacquesLucke:
Yes. Can you find out if you can replicate the same behavior of the previous Direction to Rotation node with this node?

og76:
Any idea on a speed comparison? 
Don't think should be much diff, but if it's gonna replace the old ..

Normally the old behavior happens if the guide is 000, where my guide will take xyz world.
I still have to check again, just to make sure xyz stuff works same for 0 guide. 
(kinda had my nose too much into it and got a bit sick of xy yx yz.. and not sure if I see it right)

+It does with a little +:
I added those exceptions so that for 0 dir it will take xyz+/-  world. So even if u have track 000 the xyz+- are still gonna matter.. I wonder if that really matters.. ??

JacquesLucke:
I just tried to change the direction to rotation node. Here is what I came up with: http://www.pasteall.org/61598/python

I'm also not sure if all these xyz changing is correct. I don't even know how I can check if it is right or not..

What does *if direction != 0* internally do? I wonder if it is the same as *if direction.length_squared != 0*..
I didn't found anything about that in the api.

og76:
1.
it was  != Vector((0,0,0)), but found somewhere that if vec != 0   works as   if length != 0, witch is same
maybe is more readable and direct to say != mathutils.Vector((0,0,0))
(I have to check where I got that ...)

2.
the getAxesChange should be ok. 
I have first made that with dir Z, guide X, thus the notation in the first part.

Then, to see what xyz become for each case (mx,my,mz), got checking with **an Empty with axes**, manually rotated it around to see for each case. 
I guess that is the way to check for it, so for each case to see if they align as expected.
Another way is like Jose blend above, where it compares with the constraint track-to. I did that, but it does not coverall the cases ...

I think the dir 0 and guide 0 parts are ok too, but should check again if the the mxmymz are world xyz
Plus, **this ia a bit weird**, normally if 0 the cross just gives 0, where I substitute with this world XYZ. 
So **you see guide as 000, but actually it is 100, 010, 001 as the buttons are xyz**

og76:
here's the code that works right: #257 or the exec part in 
https://gist.github.com/og76/b5f876f642d035504bb7

1. the getAxes is the same!
I have tested on all cases and works as expected, and as the old one.
2. for the exceptions, the colinear, similar or paralel, I have considered here the minimum flipping cases
 - in this version the only cases where u return 000 is when: u go thru 0 with both
 - an extra flip goes only for 2 cases have 0 guide, thus world axis and dir is on the same axis, going thru 0. And have tested intensively in comparison to the old ver, that does the same or worse...

**note that with these 2 "ifs" for direction and guide, I am pushing this to extra working cases, rather than dealing with errors.  these so called exceptions are in fact giving 000**
- In the old version: If direction is 000 it does nothing, here I consider xyz+/- so the guide is still working against world dir
- the guide, that was always up 001, now I give it XYZ world, but also


the solutions provided will improve the case, but will never cover for sudden flipping or such if u go thru 0. And the to track quat does no better job there cause in the end, u cannot predict colinear or 0 vectors intention..

JacquesLucke:
discussion moves to #257 from here

PR #236, opened 2015-10-06
Nla action strip sockets

og76:
and the sockets for the nla strip nodes, list and item
(the color is a dark brown, more like fcurves, but browner...)

I made this separated while the nodes still have generic sockets. So the nodes can be tested without this modification.


JacquesLucke:
maybe we should use "NLA Strip" as datatype; with *Space*

PR #243, opened 2015-10-07
add missing nodes to menu - part 1

kmeirlaen:
First effort in adding the missing nodes to the menu.
will do another push later this afternoon (lunch time!)

### DIFF
--- a/ui/node_menu.py
+++ b/ui/node_menu.py
@@ -63,9 +63,12 @@ class NumberMenu(bpy.types.Menu):
         insertNode(layout, "an_RandomNumberNode", "Randomize")
         insertNode(layout, "an_FloatWiggleNode", "Wiggle")
         insertNode(layout, "an_MixDataNode", "Mix", {"dataType" : repr("Float")})
+        insertNode(layout, "an_MapRangeNode", "Map Range")
         layout.separator()
         insertNode(layout, "an_FloatClampNode", "Clamp")
         insertNode(layout, "an_ConvertAngleNode", "Convert Angle")
+        insertNode(layout, "an_FloatToIntegerNode", "Convert Float To Integer")
+        insertNode(layout, "an_RoundNumberNode", "Round")
         insertNode(layout, "an_FloatMathNode", "Math")
 
 class VectorMenu(bpy.types.Menu):
@@ -169,6 +172,7 @@ class ListMenu(bpy.types.Menu):
         insertNode(layout, "an_CreateListNode", "Create")
         insertNode(layout, "an_CombineListsNode", "Combine")
         insertNode(layout, "an_AppendListNode", "Append")
+        insertNode(layout, "an_SearchListElementNode", "Search")
         layout.separator()
         insertNode(layout, "an_GetListElementNode", "Get Element")
         insertNode(layout, "an_GetListLengthNode", "Get Length")
@@ -176,6 +180,7 @@ class ListMenu(bpy.types.Menu):
         insertNode(layout, "an_ShuffleListNode", "Shuffle")
         insertNode(layout, "an_ReverseListNode", "Reverse")
         insertNode(layout, "an_SliceListNode", "Slice")
+        insertNode(layout, "an_ListMathNode", "Math")
 
 class ObjectMenu(bpy.types.Menu):
     bl_idname = "an_object_menu"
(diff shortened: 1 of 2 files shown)


kmeirlaen:
I'm not really sure about the "an_CurveObjectOutputNode" location, but at least it is there now and can be moved if needed

ui/node_menu.py
```diff
@@ -339,7 +360,10 @@ def draw(self, context):
         layout = self.layout
         insertNode(layout, "an_SoundBakeNode", "Bake Sound")
         insertNode(layout, "an_SoundFromSequencesNode", "Sound from Sequences")
+        insertNode(layout, "an_GetAllSequencesNode", "Get All Sequences")
+        insertNode(layout, "an_SequencesFromChannelNode", "from Channel")
```
JacquesLucke: The name here shouldn't be "from Channel" in my opinion. The user maybe thinks that it means "Sound from Channel" which isn't the case. Maybe "Sequences from Channel" would be better. Or we make a new category for Sequence nodes

JacquesLucke:
One thing you can do here is to correct some of your and some of my 'style mistakes'
- Between the menu classes should be exactly one empty line.
- layout.separator() shouldn't have empty lines around it. The difference in line length is enouph to separate it from the others.

Thanks.

kmeirlaen:
made suggested changes.

JacquesLucke:
Ups sorry. My fault. 
The node shouldn't be called "Sound from Channel" but "Sequences from Channel".
Rest looks ok. Maybe I will have to change the order of the nodes later but thats ok

kmeirlaen:
no problem :) I'll have it in the next commit.
and yes, I'm trying to apply some logic to the order, but some menus are becomming really big now.
But again, at least the nodes are there and can be found, it will be easy to move them around afterwards.

JacquesLucke:
right. maybe we can also make more submenus. I know, some people don't like that but the menu isn't there for fast node access, but to find a node you forgot the name from

kmeirlaen:
well, the names in menus are sometimes (read: usually) different then the node name itself, making it a bit more confusing... :)
Also, I think it also would be useful to have a tab interface on the left to drag nodes into the tree. But I would just keep this in mind and focus on some other things now (unless you have nothing else to do) :D

JacquesLucke:
I removed all these tabs intentionally because they were VERY annoying. You could never found the right tab because there are so many and I need the Animation Nodes tab most of the time anyway.
I fixt some of these naming things yesterday but you are right. The problem is that some nodes can be used differently in different contexts

kmeirlaen:
hehe, yes, there are sooooo many nodes :)
i'm focussing on getting them in the menu now, we can tackle other issues later...

JacquesLucke:
ok :)

kmeirlaen:
OK, almost done, the majority should be in the menu now

Below are the nodes that I could not find a proper place for. Some should not even be there I guess.

an_BlendDataByNameNode
an_ConvertNode
an_ExpressionNode
an_MixDataListNode
an_PrototypeNode
an_ObjectMeshDataNode
an_PolygonInfoNode
an_VertexInfoNode
an_LoopGeneratorOutputNode
an_ReassignLoopParameterNode
an_CharacterPropertiesOutputNode
an_ConvertToStringNode

JacquesLucke:
Right some of these nodes don't belong in the menu.
Thanks for your effort

JacquesLucke:
next step are submenus. Nodes you need often shouldn't be in a deeper menu but many of the nodes are used very rarely

JacquesLucke:
ah and also the list math node.  It shouldn't be in the list category because it isn't generic. It just works with a float list. I think the "Number" category is better here

kmeirlaen:
submenus: it is a matter of taste (just like the tabs).
And also how you use AN. If you use it for animation, the animation nodes are important, but if you use AN for mesh creation, the mesh/bmesh/edge/poly/... nodes become important.
I finally got my head around AN though: it is a node base programming interface :)
I only miss a better way to do if/then/else
There is switch, but you first have to perform both routines and then map one of those 2 to an outcome. It would be more performant if only one needs to execute. anyway, i'm full of ideas, but don't want to overload you and will jump in where possible.

JacquesLucke:
sure. I mean these top-categories will stay the same. But eg the Object category is a bit overloaded. 
Some nodes there should be in a submenu and even it is just called "Utils". Just so that we can better focus on the important nodes

Yeah If-statements are a bit difficult so far. I have an idea how it can work in mind but there is a lot of other stuff to do currently.

Great that you want to help!

PR #246, opened 2015-10-07
added to/from matrix to both eul/quat

og76:
matrix conversion gives clean rotation matrix, ready for transform nodes or matrix combine
(+this will make guided track complete :D )

1
 we need quat / matrix as that opens quat to any transforms
2
 I know euler / matrix is already there, but somehow is never in this concise, rotation only form
- either we have rot matrix separated xyz or in the combine node. i used that a lot for just rotation
- for output we have matrix decompose, but again, I just like having the simple eul to mat/mat to eul when it comes to combining several matrices
(btw, any improvement in the matrix compose?)

3
I also wanted to add quat to vector and angle / vec and angle to quat, but that is weird.
how would you use that in a loop? as separate vec/float or tuple of both?
the axis angle rotation is yet another data type/soc and uses this tuple, while I was more tempted by a separate vec, float in/out as that is easier to deal with in nodes that a weird tuple...

### DIFF
--- a/nodes/rotation/convert_quaternion_and_euler.py
+++ b/nodes/rotation/convert_quaternion_and_euler.py
@@ -5,7 +5,11 @@ from ... base_types.node import AnimationNode
 
 conversionTypeItems = [
     ("QUATERNION_TO_EULER", "Quaternion to Euler", "", "NONE", 0),
-    ("EULER_TO_QUATERNION", "Euler To Quaternion", "", "NONE", 1)]
+    ("EULER_TO_QUATERNION", "Euler To Quaternion", "", "NONE", 1),
+    ("QUATERNION_TO_MATRIX", "Quaternion to Matrix", "", "NONE", 2),
+    ("MATRIX_TO_QUATERNION", "Matrix To Quaternion", "", "NONE", 3),
+    ("EULER_TO_MATRIX", "Euler To Matrix", "", "NONE", 4),
+    ("MATRIX_TO_EULER", "Matrix To Euler", "", "NONE", 5)]
 
 class ConvertQuaternionAndEulerNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ConvertQuaternionAndEulerNode"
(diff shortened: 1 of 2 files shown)


nodes/rotation/convert_quaternion_and_euler.py
```diff
@@ -5,7 +5,11 @@
...
 conversionTypeItems = [
     ("QUATERNION_TO_EULER", "Quaternion to Euler", "", "NONE", 0),
-    ("EULER_TO_QUATERNION", "Euler To Quaternion", "", "NONE", 1)]
+    ("EULER_TO_QUATERNION", "Euler To Quaternion", "", "NONE", 1),
+    ("QUATERNION_TO_MATRIX", "Quaternion to Matrix", "", "NONE", 2),
+    ("MATRIX_TO_QUATERNION", "Matrix To Quaternion", "", "NONE", 3),
+    ("EULER_TO_MATRIX", "Euler To Matrix", "", "NONE", 4),
+    ("MATRIX_TO_EULER", "Matrix To Euler", "", "NONE", 5)]
 
 class ConvertQuaternionAndEulerNode(bpy.types.Node, AnimationNode):
```
JacquesLucke: you will have to rename the node for that. eg to ConvertRotationsNode

nodes/rotation/convert_quaternion_and_euler.py
```diff
@@ -14,7 +18,11 @@ class ConvertQuaternionAndEulerNode(bpy.types.Node, AnimationNode):
     onlySearchTags = True
     searchTags = [
         ("Quaternion to Euler", {"conversionType" : repr("QUATERNION_TO_EULER")}),
-        ("Euler to Quaternion", {"conversionType" : repr("EULER_TO_QUATERNION")}) ]
+        ("Euler to Quaternion", {"conversionType" : repr("EULER_TO_QUATERNION")}),
+        ("Quaternion to Matrix", {"conversionType" : repr("QUATERNION_TO_MATRIX")}),
+        ("Matrix to Quaternion", {"conversionType" : repr("MATRIX_TO_QUATERNION")}),
+        ("Euler To Matrix", {"conversionType" : repr("EULER_TO_MATRIX")}),
```
JacquesLucke: small 't' please in 'To'

PR #252, opened 2015-10-08
added integer range variant

og76:
very useful for indices (yes, I have an obsession..) and others

The name could change, but it's not necessary
The location can be in generic, though it 's only bout int and float, so it can stay in numbers very well


JacquesLucke:
Looks very good. You are definitly learning a lot here :+1:

og76:
yes. that is the point ..

PR #257, opened 2015-10-10
Update direction_to_rotation.py

og76:
this reduces the 000 cases to minimum
see the comments in the other pull #232

### DIFF
--- a/nodes/rotation/direction_to_rotation.py
+++ b/nodes/rotation/direction_to_rotation.py
@@ -31,18 +31,19 @@ class DirectionToRotationNode(bpy.types.Node, AnimationNode):
         if self.trackAxis[-1:] == self.guideAxis[-1:]:
             yield "eulerRotation = mathutils.Euler((0, 0, 0), 'XYZ')"
             return
+        
+        yield "zero, worldX, worldY, worldZ = " + getWorldVectors()
 
-        yield "if direction != 0: z = direction.normalized()"
-        if "X" in self.trackAxis: yield "else: z = mathutils.Vector((1, 0, 0))"
-        if "Y" in self.trackAxis: yield "else: z = mathutils.Vector((0, 1, 0))"
-        if "Z" in self.trackAxis: yield "else: z = mathutils.Vector((0, 0, 1))"
+        yield "if direction != zero: z = direction.normalized()"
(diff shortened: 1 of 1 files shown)


nodes/rotation/direction_to_rotation.py
```diff
@@ -62,3 +63,10 @@ def getAxesChange(track, guide):
     elif track == "-Y": a = "(-y,-z, x)" if guide == "Z" else "( x,-z, y)"
     elif track == "-Z": a = "( x,-y,-z)" if guide == "X" else "( y, x,-z)"
     return a
+
+def getWorldVectors():
+    o = "mathutils.Vector((0, 0, 0))"
+    x = "mathutils.Vector((1, 0, 0))"
+    y = "mathutils.Vector((0, 1, 0))"
+    z = "mathutils.Vector((0, 0, 1))"
+    return o + ", " + x + ", "+ y + ", "+ z
```
JacquesLucke: should be ", ".join([o, x, y, z)]

JacquesLucke:
I'm not sure if it is good to somehow use invalid input data.
Maybe we should always return (0, 0, 0) when the inputs have invalid values.
Advantages:
1. User immediately sees when the values are invalid -> otherwise he maybe tries all the different combinations to get a meaningfull result
2. Better performance because in you code here the world vectors are always created, even if they are not needed.
3. Code becomes cleaner because we do only one error check at the top.

og76:
I was wondering the same thing, if it's not exaggerated.

1. DIR The original leaves all 000 for direction=000. But there is only one input ...
However, could look at leaving all dead if the direction is 000. That is probably the exaggeration.
**I can look if the can go out ..**

2. GUI Now, the original tries to cover for all the UP exceptions to ensure kinda continuity or so. That makes here for guide. I've compared all the orientation for that and we kinda do the same. 
The point is to  **never loose the direction** for colinear or so.

In these cases I made it so it just flips on the other side, or at most on a perpendicular intermediary position, instead of going to full 3x0.( That would lose the dir). Somehow, these should be there.


The exceptions are not for no input or just 000, but for the (very rare) cases where you'd accidentally pass thru a0/paralel position or so.
(Imho that is avoidable by shifting a bit your data, but when it comes to procedural, hm)

I'll look if some can go out, but probably it would be on the dir =0

PR #261, opened 2015-10-13
Create float_format_as_string.py

og76:
this is the equivalent of (with these default values..):
```
'{:+08.3f}'.format(Float)
```
(as per https://docs.python.org/2/library/string.html#format-string-syntax)

formats the float with 
* +/- sign or _/-,
* precision, number of digits, also filled with 0 if round
* leading zeros or spaces to fill the len

format2

this escapes about 10 nodes needed to trim, join and condition such 

it is not a replacement for fill, as that fills with any char and has broader use
but for floats, that is not enough

### DIFF
--- a/nodes/text/float_format_as_string.py
+++ b/nodes/text/float_format_as_string.py
@@ -1,0 +1,40 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+signModeItems = [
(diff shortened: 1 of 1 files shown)


og76:
the name (+ of the file) is not great ..

og76:
any chance here?

JacquesLucke:
well I'm not sure if it will go in as it is now... I have lots of time today and tomorrow (at least looks like so atm :D)
I think I will make a node similiar to this one within these days :)

og76:
great

PR #262, opened 2015-10-13
adding useWorld as advanced option

og76:
default still true
this is important as I have used it to compare obj data dimensions with or without transforms

about half on the forum back I was arranging and resizing texts to fit a box or objects to fit each other
I used Bbox to get the original mesh size to compare with the scaled one.
Also useful for min max coords etc (min/max is vert 0-6, opposite corners..in ob space..)

### DIFF
--- a/nodes/object/object_bounding_box.py
+++ b/nodes/object/object_bounding_box.py
@@ -1,15 +1,14 @@
 import bpy
 from bpy.props import *
-from ... events import executionCodeChanged
 from mathutils import Vector
+from ... events import propertyChanged
 from ... base_types.node import AnimationNode
 
 class ObjectBoundingBoxNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ObjectBoundingBoxNode"
     bl_label = "Object Bounding Box"
     
-    useWorldSpace = BoolProperty(name = "Use World Space", default = True,
-        update = executionCodeChanged)
+    useWorldSpace = BoolProperty(name = "Use World Space", default = True, update = propertyChanged)
     
     def create(self):
         self.inputs.new("an_ObjectSocket", "Object", "object").defaultDrawType = "PROPERTY_ONLY"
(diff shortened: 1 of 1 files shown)


nodes/object/object_bounding_box.py
```diff
@@ -1,21 +1,31 @@
...
         self.outputs.new("an_VectorListSocket", "Vertices", "vertices")
         self.outputs.new("an_EdgeIndicesListSocket", "Edges", "edges")
         self.outputs.new("an_PolygonIndicesListSocket", "Polygons", "polygons")
-
+        
+    def drawAdvanced(self, layout):
+        layout.prop(self, "useWorldSpace")
+    
     def execute(self, object):
         if object is not None:
```
JacquesLucke: please change that to:
if object is None: return [], [], []
(i know that it was my 'mistake', not yours)

nodes/object/object_bounding_box.py
```diff
@@ -1,21 +1,31 @@
...
 from mathutils import Vector
 from ... base_types.node import AnimationNode
 
 class ObjectBoundingBoxNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ObjectBoundingBoxNode"
     bl_label = "Object Bounding Box"
-
+    
+    useWorldSpace = BoolProperty(name = "Use World Space", default = True,
+        update = executionCodeChanged)
```
JacquesLucke: we don't use custom execution code here. 
Using propertyChanged instead of executionCodeChanged is better.

nodes/object/object_bounding_box.py
```diff
@@ -1,21 +1,31 @@
 import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
```
JacquesLucke: just easier for my eyes -> sort the import statements based on their length :D

PR #264, opened 2015-10-13
make dreaw rectangles a bit transparent

og76:
these should have a transparency cause
1. you can select and manipulate things underneath without knowing, that can lead to unexpected events
2. you should see if any underneath, after all the nodes are bit transparent too
3. you may even compare graphs or values

goes for both interpolation and drawer (and all that will use rectangle..)


og76:
transparent graph

PR #284, opened 2015-10-23
more vector and rotation ops 1

og76:
separate eul and quat
quat to axis angle (and back)

vector dot, 
vector angle (and rotational difference - quat), 
vector angle 2d (with matrix for custom 2d plane)

I know I said to propose 1 by 1, but these are really small and they like being together :)
more to come...

### DIFF
--- a/nodes/rotation/convert_quaternion_and_axis_angle.py
+++ b/nodes/rotation/convert_quaternion_and_axis_angle.py
@@ -56,7 +56,7 @@ class ConvertQuaternionAndAxisAngleNodeNode(bpy.types.Node, AnimationNode):
             self.outputs.new("an_VectorSocket", "Axis", "axis")
             self.outputs.new("an_FloatSocket", "Angle", "angle")
         if self.conversionType == "AXIS_ANGLE_TO_QUATERNION":
-            self.inputs.new("an_VectorSocket", "Axis", "axis")
+            self.inputs.new("an_VectorSocket", "Axis", "axis").value = (0, 0, 1)
             self.inputs.new("an_FloatSocket", "Angle", "angle")
             self.outputs.new("an_QuaternionSocket", "Quaternion", "quaternion")
         self.inputs[0].defaultDrawType = "PREFER_PROPERTY"
--- a/nodes/rotation/convert_quaternion_and_axis_angle.py
+++ b/nodes/rotation/convert_quaternion_and_axis_angle.py
@@ -1,0 +1,62 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+conversionTypeItems = [
+    ("QUATERNION_TO_AXIS_ANGLE", "Quaternion to Axis Angle", "", "NONE", 0),
+    ("AXIS_ANGLE_TO_QUATERNION", "Axis Angle to Quaternion", "", "NONE", 1) ]
+
+class ConvertQuaternionAndAxisAngleNodeNode(bpy.types.Node, AnimationNode):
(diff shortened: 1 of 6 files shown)


nodes/rotation/separate_euler.py
```diff
@@ -0,0 +1,30 @@
...
+    
+    def getExecutionCode(self):
+        if self.useDegree:
+            toDegree = "180 / math.pi"
+            return "x, y, z = euler.x * {0}, euler.y * {0}, euler.z * {0}".format(toDegree)
+        else:
+            return "x, y, z = euler"
+    
+    def getUsedModules(self):
+        return ["mathutils", "math"]
```
JacquesLucke: you don't use mathutils here, do you?

nodes/vector/vector_angle.py
```diff
@@ -0,0 +1,22 @@
...
+    bl_idname = "an_VectorAngleNode"
+    bl_label = "Vector Angle"
+    
+    def create(self):
+        self.inputs.new("an_VectorSocket", "A", "a")
+        self.inputs.new("an_VectorSocket", "B", "b")
+        self.outputs.new("an_FloatSocket", "Angle", "angle")
+        self.outputs.new("an_QuaternionSocket", "Rotation Difference", "quat")
+        
+    def getExecutionCode(self):
```
JacquesLucke: lets use yield here instead of a list of lines. 
the "if not any..." is not needed here

nodes/vector/vector_angle.py
```diff
@@ -0,0 +1,22 @@
...
+
+class VectorAngleNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_VectorAngleNode"
+    bl_label = "Vector Angle"
+    
+    def create(self):
+        self.inputs.new("an_VectorSocket", "A", "a")
+        self.inputs.new("an_VectorSocket", "B", "b")
+        self.outputs.new("an_FloatSocket", "Angle", "angle")
+        self.outputs.new("an_QuaternionSocket", "Rotation Difference", "quat")
```
JacquesLucke: maybe the identifier should be "rotationDifference" :)

nodes/vector/vector_dot_product.py
```diff
@@ -0,0 +1,14 @@
+import bpy
+from ... base_types.node import AnimationNode
+
+class VectorDotProductNode(bpy.types.Node, AnimationNode):
```
JacquesLucke: Should we maybe make this operation part of the vector math node?
the api seems to be wrong about this operation. It sais the dot function returns a vector: http://www.blender.org/api/blender_python_api_2_76_1/mathutils.html#mathutils.Vector.dot

nodes/rotation/convert_quaternion_and_axis_angle.py
```diff
@@ -0,0 +1,62 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+conversionTypeItems = [
+    ("QUATERNION_TO_AXIS_ANGLE", "Quaternion to Axis Angle", "", "NONE", 0),
+    ("AXIS_ANGLE_TO_QUATERNION", "Axis Angle to Quaternion", "", "NONE", 1) ]
+
+class ConvertQuaternionAndAxisAngleNodeNode(bpy.types.Node, AnimationNode):
```
JacquesLucke: "NodeNode"

JacquesLucke:
Can you make the Separate Euler, Separate Quaternion and Vector Angle node in a separate pull request?
These are totally fine. I'm still thinking about the others :)

PR #298, opened 2015-11-02
Create convert_quaternion_and_axis_angle.py

og76:
this is a human way to make quaternions, otherwise kind of abstract

I thought it should actually go as alternative in combine/separate quaternion, but for the sake of not braking stuff it can go very well separated (for users it's the same..)

### DIFF
--- a/nodes/rotation/convert_quaternion_and_axis_angle.py
+++ b/nodes/rotation/convert_quaternion_and_axis_angle.py
@@ -1,0 +1,62 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+conversionTypeItems = [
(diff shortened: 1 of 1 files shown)


JacquesLucke:
Shouldn't this be a part of the convert rotations node?
I mean the basic structure is pretty much the same, isn't it?
Also this doesn't break any existing files as long as you don't change the numbers for the different types. (0 to 5)

og76:
yes, initially I did that.
but then I realized:
- is only between quat and axis angle, the rest being any to any
- it is actually a way to compose/decompose quat just like wxyz 
- it has this extra use degree that again is inconsistent

I would have made it part of combine/separate
However, I'm probably making too big a philosophy here and is probably more efficient for file management etc
if you think it should be there I think I have the other version..

og76:
you know what?
I'll make it part of convert rotations cause for users it's more convenient to just switch to any other type instead of getting another node

JacquesLucke:
I merged the other pull request so I will close this. Thank you :)

PR #308, opened 2015-11-07
Update float_to_string.py

og76:
fwiw, the same result, no need for sign acrobatics ..

or like this:
```python
    def getExecutionCode(self):
        yield 'sign = "+" if insertSign else ""'

        yield 'formatString = "{" + ":{}0{}.{}f".format(sign, max(minLength , 0), max(decimals, 0)) + "}"'
        yield 'text = formatString.format(number)'
```
(just use it or not)


og76:
Any chance here? :)

this is just a minor cleanup, 
cause you don't need to do the math for sign, the format does it already
and - is not needed as a 3rd variant for sign

JacquesLucke:
yep. thanks.

PR #311, opened 2015-11-08
random and wiggle rotations Patch 1

og76:
random eul / quat + wiggle  eul / quat
(just like the vectors)
for quat I keep
- w = 1 in random, cause otherwise it's very hard to control and scale does not matter
this way it works only for rotation, just like the eul variant
- w = amplitude.w same reasons. still, here I give control over w so that you can tweak it other than 1, but not the randomness

### DIFF
--- a/nodes/rotation/random_euler.py
+++ b/nodes/rotation/random_euler.py
@@ -13,7 +13,7 @@ class RandomEulerNode(bpy.types.Node, AnimationNode):
 
     def create(self):
         self.inputs.new("an_IntegerSocket", "Seed", "seed")
-        self.inputs.new("an_FloatSocket", "Scale", "scale").value = pi/6
+        self.inputs.new("an_FloatSocket", "Scale", "scale").value = pi / 6
         self.outputs.new("an_EulerSocket", "Euler", "randomEuler")
 
     def draw(self, layout):
--- a/nodes/rotation/random_euler.py
+++ b/nodes/rotation/random_euler.py
@@ -1,6 +1,6 @@
 import bpy
 import random
-from math import pi
+from math import radians
 from bpy.props import *
 from ... events import propertyChanged
 from ... base_types.node import AnimationNode
@@ -13,7 +13,7 @@ class RandomEulerNode(bpy.types.Node, AnimationNode):
 
     def create(self):
         self.inputs.new("an_IntegerSocket", "Seed", "seed")
-        self.inputs.new("an_FloatSocket", "Scale", "scale").value = pi / 6
+        self.inputs.new("an_FloatSocket", "Scale", "scale").value = radians(30)
         self.outputs.new("an_EulerSocket", "Euler", "randomEuler")
 
     def draw(self, layout):
--- a/nodes/rotation/random_euler.py
+++ b/nodes/rotation/random_euler.py
@@ -1,5 +1,6 @@
 import bpy
 import random
+from math import pi
 from bpy.props import *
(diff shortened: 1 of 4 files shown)


nodes/rotation/wiggle_euler.py
```diff
@@ -0,0 +1,33 @@
...
+class EulerWiggleNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_EulerWiggleNode"
+    bl_label = "Euler Wiggle"
+
+    nodeSeed = IntProperty(update = propertyChanged)
+
+    def create(self):
+        self.inputs.new("an_FloatSocket", "Seed", "seed")
+        self.inputs.new("an_FloatSocket", "Evolution", "evolution")
+        self.inputs.new("an_EulerSocket", "Amplitude", "amplitude").value = [1.570796, 1.570796, 1.570796]
```
JacquesLucke: would be better to say math.pi / 2 instead of this 'magic number'

nodes/rotation/random_euler.py
```diff
@@ -0,0 +1,31 @@
...
+
+class RandomEulerNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_RandomEulerNode"
+    bl_label = "Random Euler"
+
+    nodeSeed = IntProperty(name = "Node Seed", update = propertyChanged, max = 1000, min = 0)
+
+    def create(self):
+        self.inputs.new("an_IntegerSocket", "Seed", "seed")
+        self.inputs.new("an_FloatSocket", "Scale", "scale").value = 0.8
```
JacquesLucke: why 0.8?

og76:
yes, also changed the values to smaller ones, so that you can see what is happening 
(rotations should have smaller val, i.e. for 90* you get to flip from one side to another and the movement seams chaos)
eul to 30* (pi/6) quat to 0.3

nodes/rotation/wiggle_euler.py
```diff
@@ -0,0 +1,34 @@
...
+class EulerWiggleNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_EulerWiggleNode"
+    bl_label = "Euler Wiggle"
+
+    nodeSeed = IntProperty(update = propertyChanged)
+
+    def create(self):
+        self.inputs.new("an_FloatSocket", "Seed", "seed")
+        self.inputs.new("an_FloatSocket", "Evolution", "evolution")
+        self.inputs.new("an_EulerSocket", "Amplitude", "amplitude").value = [pi / 6, pi / 6, pi / 6]
```
JacquesLucke: sorry, just get the idea. Maybe its even better to use https://docs.python.org/3/library/math.html#math.radians here..

nodes/rotation/random_euler.py
```diff
@@ -0,0 +1,32 @@
...
+
+class RandomEulerNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_RandomEulerNode"
+    bl_label = "Random Euler"
+
+    nodeSeed = IntProperty(name = "Node Seed", update = propertyChanged, max = 1000, min = 0)
+
+    def create(self):
+        self.inputs.new("an_IntegerSocket", "Seed", "seed")
+        self.inputs.new("an_FloatSocket", "Scale", "scale").value = pi / 6
```
JacquesLucke: same here

JacquesLucke:
Thanks

Issue #321, opened 2015-11-13
[Feature Request] Debug Length


LeonCheung:
When debugging lists, it would be cool to show its length as well. (Sort of a length info feature? Or maybe a switch or a separate node?)

I found sometimes it is a bit boring to add many many Get List Length nodes first, just to know what's the lengths are...

JacquesLucke:
I work on that right now. 
It will be in the Debug Drawer node

JacquesLucke:
![debug drawer](https://cloud.githubusercontent.com/assets/5881326/11147586/41a46e32-8a17-11e5-9e4f-02439eacff5c.PNG)

LeonCheung:
Cool! How did you know that I have been waiting for the show indices button as well? Thanks a lot for the fast work!

JacquesLucke:
haha I wanted them as well. But they didn't fit into the 'design' before. No everything should work like charme. Tell me if something is broken still.

LeonCheung:
So far so good. :)
Btw, since there are several debug nodes commonly used, maybe we could list more debug node on the pie?

JacquesLucke:
I prefer it like it is now because having multiple 'debug directions' in the pie would be a bit overload for such a simple operation. And putting two buttons at the bottom isn't so nice as well because it takes longer to insert them because you have to move the mouse more carefully.
Also in most cases the simple algorithm that chooses which debug node the user might want should be good enouph in most cases.

LeonCheung:
Yep.

PR #322, opened 2015-11-13
Some Debug Drawer Improvements #321

JacquesLucke:
- Text Box has a small padding for better readibility
- show warning when not all lines fit in the text box
- show list indices
- text box font size changes depending on the dpi
- show list length for list sockets
- can decide to show list elements in a row or line by line
- the debug drawer is the default debug node for list sockets


JacquesLucke:
#321

Issue #294, opened 2015-11-01
Euler Math?


LeonCheung:
![euler math](https://cloud.githubusercontent.com/assets/494788/10869050/cddc1ccc-80dc-11e5-8b73-7aecab22a414.jpg)

When doing math for euler, it seems we have to use quite a bit of converters.
Maybe there is way to make it looking a bit more compact? An Euler Math node or mode, perhaps?

JacquesLucke:
yes. Euler Math seems to be the way to go here..

og76:
Already done. But besides +/- there's not much to it. I will explain later.
But yes, is needed.
The Quat math is another thing ... way more interesting

Dum, 1 nov. 2015, 15:18, Jacques Lucke <notifications@github.com> a scris:

JacquesLucke:
ok. Can you please test if the node works in all cases or if some operations are missing?
I didn't reimplement operations that only make sense with vectors. If you need these you should convert to vector.

JacquesLucke:
![euler math](https://cloud.githubusercontent.com/assets/5881326/12401377/4051737a-be26-11e5-8d18-20b3533cd366.PNG)

og76:
1.
Yes,  there is an issue with *multiply and divide*

For these we need to convert back and forth radians and degrees, so that the visual feedback of the euler socket corresponds to the expected operation:
- if you mul 90 with 2 you expect 180, but now radians will mul pi/2 with some 0.00xxx 
- to figure that out, the user needs to know how this works
- to solve that you have to add extra nodes. Always

So, to solve this, I added 
- a useDegrees prop that by default is True
- convert to degrees, multiply , convert to rad
- an advanced button (with and operations with degree type so that only appears for these two)
that should stay always on. But if someone in this world is using radians input as a preference in blender, that should be off to get back to the normal op

in the pic you can see the difference with the proposal above (and useDegree stuff) and old below
![euler math](https://cloud.githubusercontent.com/assets/11709974/12406788/08554216-be59-11e5-9918-3d0ba0930663.PNG)

-

2.
I don't see why you use vector / euler conversion 2 times each, it does not really help besides + and - and adds unnecessary float ops to the node.

So *I propose a direct mathutils.Euler variant*, that otherwise in code style/ lines efficiency is pretty much the same. I think it matters as this is a node prone to be used in loops and many times.

Please correct me here if I'm wrong.

-

3.
Sorry delayed this node that much, this issue would have been clear in the beginning
[...]

PR #361, opened 2016-01-19
Update euler_math.py for * and / and vec/euler

og76:
update to work as expected in * and / , as the euler socket input is visually misleading here
(thus adding useDegree prop, advanced button and operations with degree to show or not the button )

returning directly eulers, so that there are no 2 operations for vector to euler for each case
(vector is not helping much here and has nothing to do with the above)

more in issue #294


JacquesLucke:
Looks good! Thanks.

I will test it later today.

PR #362, opened 2016-01-19
eul math for master same as 1.5

og76:
@keepNodeLinks is the only difference

to complete #294


JacquesLucke:
Didn't test it but looks good as well, thanks.

PR #364, opened 2016-01-19
Create quaternion_list_combine.py

og76:
this works very much like the matrix combine node, for quat, 
and as a complement to quat math
(actually using the same  q*q *q in inverse order)


og76:
and that's it for #294

Issue #365, opened 2016-01-19
add shear matrix node


og76:
a matrix that does just that

For objects you'll see rather a rotation and scale on respective directions.
On verts (or instances) you'll get the expected sliding / shear

![shear matrix](https://cloud.githubusercontent.com/assets/11709974/12422807/5fcecd5a-bed1-11e5-8d21-b2176fd55d19.PNG)

I also add an option to scale on the 3rd dimension.
This makes it a bit like "stretch to". And makes use of the respective 3rd input coord...

I find it interesting that you don't need to have a loop for it ...
(of course you can do this with 2-3 matrices or so, but this is a very simple function that juat builds the respective matrix)

PR #366, opened 2016-01-19
Create shear_matrix.py

og76:
as per #365

### DIFF
--- a/nodes/matrix/shear_matrix.py
+++ b/nodes/matrix/shear_matrix.py
@@ -1,0 +1,42 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+planeItems = [("XY", "XY", ""), ("XZ", "XZ", ""), ("YZ", "YZ", "")]
+other = {"XY": "z", "XZ": "y", "YZ": "x"}
+coord = {"XY": "(0, 0, 1)", "XZ": "(0, 1, 0)", "YZ": "(1, 0, 0)"}
+
+class ShearMatrixNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ShearMatrixNode"
+    bl_label = "Shear Matrix"
+
+    plane = EnumProperty(items = planeItems, update = executionCodeChanged)
+    useThirdAsScale = BoolProperty(name = "Use Third as Scale", default = True, 
+                    description = "Use Third Coordinate as Scale, if not, only use the 2 shear coordinates", 
+                    update = executionCodeChanged)
+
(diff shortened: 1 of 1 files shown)


nodes/matrix/shear_matrix.py
```diff
@@ -0,0 +1,42 @@
...
+                    text = "Use {} as Scale".format(other[self.plane].upper()) )
+
+    def getExecutionCode(self):
+
+        if self.useThirdAsScale:
+            yield ("sca = mathutils.Matrix.Scale(shear.{}, 4, {})"
+                .format(other[self.plane], coord[self.plane]) )
+            yield ("mat = mathutils.Matrix.Shear('{}', 4, (shear.{}, shear.{}))"
+                .format(self.plane, self.plane[0].lower(), self.plane[1].lower()) )
+            yield "matrix = sca * mat"
```
JacquesLucke: let's use _scale and _matrix as temporary variables here :)

nodes/matrix/shear_matrix.py
```diff
@@ -0,0 +1,42 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+planeItems = [("XY", "XY", ""), ("XZ", "XZ", ""), ("YZ", "YZ", "")]
+other = {"XY": "z", "XZ": "y", "YZ": "x"}
```
JacquesLucke: maybe rename "other" in "perpendicularAxis" or so..

nodes/matrix/shear_matrix.py
```diff
@@ -0,0 +1,42 @@
...
+coord = {"XY": "(0, 0, 1)", "XZ": "(0, 1, 0)", "YZ": "(1, 0, 0)"}
```
JacquesLucke: and "planeNormal" instead or "coord" maybe.

nodes/matrix/shear_matrix.py
```diff
@@ -0,0 +1,42 @@
...
+        self.inputs.new("an_VectorSocket", "Shear", "shear").value = [1, 1, 1]
+        self.outputs.new("an_MatrixSocket", "Matrix", "matrix")
+
+    def draw(self, layout):
+        layout.prop(self, "plane", expand = True)
+        layout.prop(self, "useThirdAsScale", 
+                    text = "Use {} as Scale".format(other[self.plane].upper()) )
+
+    def getExecutionCode(self):
+
```
JacquesLucke: you use "self.plane" quite often in this function. Maybe it would be good to make the variable local:
"plane = self.plane"

Issue #372, opened 2016-01-21
Suggestion: Particle trajectory sampler node


YorsLosilla:
Last weeks I´m doing some experiments to recreate cool motion design examples done with other software like C4D and  Houdini. And I have found some functionality that could make AN better.

In this case the idea it´s to obtain samples from the location of a particle trough time which will help a lot to make effects like growing hair. Something similar to the Btrace plugin but with the control and flexibility that AN brings.

The idea behind this node is something similar to the curve samples nodes but applied to particles.

og76:
same as #248 
you don't have access to that data in script

the only way is to : 
- use render > path
- cache > bake
- in the modifier (particle sistems) you'll hace a convert button
- respective converted object is a mesh so you have to again convert to curves

and you get an obj with cute splines that are the paths of the partc
they can further be used as any spline in AN

YorsLosilla:
I guess that this answers my next suggestion that was a node that can retrieve from the points of each strand from a particle hair system.

Here ares some links that I can´t understand but maybe they could be usefull:

http://code.blender.org/2011/01/a-look-at-point-cache/
http://blender.stackexchange.com/questions/15577/how-to-extract-convert-data-from-blender-cache-files-bphys-into-a-human-readabl
http://blenderartists.org/forum/showthread.php?340983-How-do-you-set-hair-dynamics-and-particle-cache-with-Python
http://stackoverflow.com/questions/10581133/load-baked-particle-system-cache-in-blender

YorsLosilla:
Would it be possible  to access somehow to particle trail data?

og76:
just  proposed hair data (points etc)

they result in lists per particle, so NOT adding to particle list info
this breaks the existing node by adding other sockets, so **adding only to v1.5** !!

Not about cache, about normal hair data, as several points per particle
(the splines in the pic are generated per hairs)

![hair partc](https://cloud.githubusercontent.com/assets/11709974/12574707/fc545b6c-c40e-11e5-8b94-9c8e560f1727.PNG)

YorsLosilla:
I just can say one thing: awesome. A new Pandora's box has been opened.


-----Mensaje original-----
De: "og76" <notifications@github.com>
Enviado: ‎26/‎01/‎2016 8:28
Para: "JacquesLucke/animation_nodes" <animation_nodes@noreply.github.com>
CC: "YorsLosilla" <thexchanger@gmail.com>
[...]

PR #374, opened 2016-01-26
Update particle_info.py for hair data

og76:
adding hair data (points etc)

they result in lists per particle, so NOT adding to particle list info
this breaks the existing node by adding other sockets, so adding only to v1.5

not quite the answer and Not about cache, but got the idea on #372

### DIFF
--- a/nodes/particles/particle_info.py
+++ b/nodes/particles/particle_info.py
@@ -22,6 +22,11 @@ class ParticleInfoNode(bpy.types.Node, AnimationNode):
         self.outputs.new("an_VectorSocket", "Previous Rotation", "previousRotation").hide = True
         self.outputs.new("an_VectorSocket", "Previous Velocity", "previousVelocity").hide = True
         self.outputs.new("an_VectorSocket", "Previous Angular Velocity", "previousAngularVelocity").hide = True
+        
+        self.outputs.new("an_FloatListSocket", "HairTime", "hairTime").hide = True
+        self.outputs.new("an_FloatListSocket", "HairWeight", "hairWeight").hide = True
+        self.outputs.new("an_VectorListSocket", "HairPoints", "hairPoints").hide = True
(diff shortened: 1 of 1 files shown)


og76:
see pic in the issue

nodes/particles/particle_info.py
```diff
@@ -22,6 +22,11 @@ def create(self):
         self.outputs.new("an_VectorSocket", "Previous Rotation", "previousRotation").hide = True
         self.outputs.new("an_VectorSocket", "Previous Velocity", "previousVelocity").hide = True
         self.outputs.new("an_VectorSocket", "Previous Angular Velocity", "previousAngularVelocity").hide = True
+        
+        self.outputs.new("an_FloatListSocket", "HairTime", "hairTime").hide = True
+        self.outputs.new("an_FloatListSocket", "HairWeight", "hairWeight").hide = True
+        self.outputs.new("an_VectorListSocket", "HairPoints", "hairPoints").hide = True
+        self.outputs.new("an_VectorListSocket", "HairPointsLocal", "hairPointsLocal").hide = True
```
JacquesLucke: you forgot some spaces here
"Hair{Space}Time", etc

nodes/particles/particle_info.py
```diff
@@ -44,6 +49,17 @@ def getExecutionCode(self):
         if isLinked["previousRotation"]: lines.append("    previousRotation = mathutils.Vector(particle.prev_rotation.to_euler())")
         if isLinked["previousVelocity"]: lines.append("    previousVelocity = particle.prev_velocity")
         if isLinked["previousAngularVelocity"]: lines.append("    previousAngularVelocity = particle.prev_angular_velocity")
+        
+        hair = ["hairTime", "hairWeight", "hairPoints", "hairPointsLocal"]
+        if any ([isLinked[item] for item in hair]):
```
JacquesLucke: I don't use spaces between function names and its parameters "any{nospace}(...)"

PR #409, opened 2016-03-04
Material Indices Socket for 'Mesh data on object' node

hapit:
Post to this pull request: http://blenderartists.org/forum/showthread.php?350296-Addon-Animation-Nodes&p=3018035&viewfull=1#post3018035

Socket is hidden by default to not confuse some users.
The material indices are set on each polygon from the given integer list.
If the list is shorter than the polygon count, the list is looped.


JacquesLucke:
one problem I still see is compability with older files which already contain this node and don't have the second input.
I will correct this after the merge.

thanks.

btw: maybe you want to write the manual for this node? :D

JacquesLucke:
I modified your code a bit.. to make the whole thing faster and backward compatible.

PR #411, opened 2016-03-05
Added missing items to menu

LeonCheung:
Things to say:
1. Should the Text Sequence Ouput node be also put in Sound & Sequence
category?
2. Shape Key Input item still missing here.
3. "Filter XX List by Name" nodes should be put into the List category,
but I don't know how....


LeonCheung:
Maybe you can ignore the first two commits, just check the menu changes? I'm not sure why I commited three here...

JacquesLucke:
Hmm I'm not sure if we should mix sequence and sound stuff. They are different things.. (in fact I'm not even sure why I put them in the same category in the first place)

You mean that we have the "Text Sequence Output" node twice in the menu? Yes I think that is ok in this case.

LeonCheung:
I can foresee that there are even more data handlers coming, say Nla things etc, in that case, things will have to be put into their own submenus,imho. Sequence related nodes is one case for now.

I don't mean to have duplicates in menu, but it is also okay if it is bettwr to. I'm not sure either to be honest.

JacquesLucke:
So can you separate these categories?
I think we can duplicate the menu entry but maybe we also have to duplicate the Text Object Output node then too?...

JacquesLucke:
Thanks

Issue #430, opened 2016-03-18
add new set of geometry tools


og76:
I have tackled this for a while now and should start adding some

For this, I think we need a **dedicated folder/menu/category** called geometry
It contains ops that are not quite vector, not quite mesh stuff and so on, so I see it as a separate thing. For example the vector nodes are somewhat more basic than these 

Also, the recent distance to edge/plane are already part of it in a slightly different way

JacquesLucke:
I think we should make the new category now, but making more nodes for these category can be done later, because adding nodes doesn't break existing files etc.

og76:
the only break can be in the recent distance to line/plane that is redundant to my project point on line/plane . others indeed are no break


that also gives the location and a factor/parameter that is needed in some cases (just like the nearest bvh also gives not just distance)

also using another mathutils geometry for the edge
also, I wanna use edge as name elsewhere, when I actually use edge (socket ..) not any line ..

coming soon with these two and couple more related, like matrix to plane point/nor and inverse, or maybe, others, in time

og76:
added geometry folder (+ init)

adding:
- project point on line (includes _distance to line_ and _factor,_ as parameter 0-1 start line >> end line)
- project point on plane (includes _signed distance_ to plane)
- convert plane types: does point/normal to matrix and matrix to point/normal, 

as sometimes it's easier to refer to object matrix instead of point/normal, while point/normal is gonna be used everywhere to define a plane (cheaper). I avoid having this as option inside node, as it will repeat in many nodes

should remove : 

- distance to plane/line as those are redundant (and not more efficient)

examples:

![project point on line](https://cloud.githubusercontent.com/assets/11709974/13924919/f4203992-ef8e-11e5-8cc4-ca0d21cb9470.PNG)
![project point on plane](https://cloud.githubusercontent.com/assets/11709974/13924923/fac152e0-ef8e-11e5-80e8-3ec2f09ddcfb.PNG)

og76:
this also needs dedicated menu (and some more stuff in time)
#428

og76:
I will add more to this. 

But I just realized that I forgot to remove the distance to line/plane nodes (redundant)
I did change the vector menu to point to the Project... nodes and the proj nodes do have search tags Distance to ..

Even so, I should have removed the redundant :(.  But now the version is out ... now what?

JacquesLucke:
I'm not at the pc now, but I think I removed these redundant nodes

og76:
[...]

PR #431, opened 2016-03-18
adding geometry folder for respective cat of nodes

og76:
#430


JacquesLucke:
ok, this is a good idea. Can you move the files around and make a pull request?

og76:
this is it #430

nodes/geometry/project_point_on_plane.py
```diff
@@ -0,0 +1,32 @@
...
+        if isLinked["projection"]:
+            yield "int = mathutils.geometry.intersect_line_plane(point, point + plane_no, plane_co, plane_no, False)"
+            yield "projection = mathutils.Vector((0, 0, 0)) if int is None else int"
+        if isLinked["distance"]: yield "distance = mathutils.geometry.distance_point_to_plane(point, plane_co, plane_no)"
+ 
+    def getUsedModules(self):
+        return ["mathutils"]
+
+def getPlane():
+    return "planePoint, planeNormal if planeNormal.length_squared != 0 else mathutils.Vector((0, 0, 1))"
```
JacquesLucke: have you tested this?
Looks in the case that the length is 0, there will only be one value for two variables

og76: yep, the if only refers to normal, I'd have to put it in () to refer to both
pointplaneworks

JacquesLucke:
Looks good so far except for the one little thing I mentioned.

Also maybe the user should be able to search for "Distance Point to Line" or so? what do you think?
Projection is not what the normal user thinks of when it comes to measuring distances.

og76:
_I've worked with these for a while now, so they are ok._

They are projections, as in geometry. The distance is not the main deal, imho. Like the kdtree giving also the distance, but not the main deal

At most, **closest point to line/plane** is more readable. Distance is not enough ...hm. it gives you the point .. in the line it also gives 0-1 factor for inside line ..

JacquesLucke:
Yes I agree that the distance is not the main deal of these nodes but it is their main use for most users (at least I think so). So I still think users should find these nodes with multiple search tags. Similiar to the Time Info node which you can also find by typing "frame" in the search.
(we don't need the node twice in the menu)

og76:
should I also rename Distance to plane to **Signed Distance to plane** ? 
Cause it is not obvious, and in Line it's not

JacquesLucke:
Hmmmm okay.. yes, maybe it's better to have 'signed' in the name.

PR #432, opened 2016-03-20
correction for tuples case (integer type accepting indices types)

og76:


JacquesLucke:
good catch but I won't merge this because an Integer List should *always* be a list and not a tuple..
so node should raise an error.

og76:
Why?
this is about poly/edge tuples that are allowed as int lists. 


Use case is just this (currently not working)

set element

the input simply allows tuples .. nothing for the rest.
In the set node, the output is always an Integer list (and socket) and for the edge/poly you need to further create poly/edge indices anyway.

JacquesLucke:
ah okay, I see.
two things:
1. we should check wether the origin socket is really edge/polygon 
2. we should not use a list comprehension here to convert the tuple into a list. instead just use the list operator. for that we need to rename the identifier from "list" to something else...

og76:
should be ok now. 
and yeah, better to check the soc instead of inside code..

JacquesLucke:
Have you actually tested this?
I'm fine with the rename to _list but I think there are other problems with the code now.
I'll comment in the code

nodes/list/set_list_element.py
```diff
@@ -43,22 +43,29 @@ def drawAdvanced(self, layout):
 
     def getExecutionCode(self):
         yield "self.errorMessage = ''"
+        if self.isTuple: yield "_list = list(_list)"
```
JacquesLucke: do you miss () here?
self.isTuple will always be true.
Another option would be to use the "@property" decorator on the isTuple function

nodes/list/set_list_element.py
```diff
@@ -43,22 +43,29 @@ def drawAdvanced(self, layout):
...
+                yield "if 0 <= index <= len(_list) - 1: _list[index] = element"
         yield "else: self.errorMessage = 'Index out of range'"
 
     def edit(self):
         dataType = self.getWantedDataType()
         self.assignType(dataType)
 
+    def isTuple(self):
+        listInput = self.inputs["List"].dataOrigin
+        if listInput is not None and isLimitedList(listInput.bl_idname): return True
```
JacquesLucke: this line throws an exception when listInput is None, right?
If you would do the same thing in C code it would actually work, but Python always evaluates everything in the conditional expression.
listInput.bl_idname   throws the error here

Issue #429, opened 2016-03-18
make general change pivot node instead of extra options in the transform polygon


og76:
this would be a convenience, for the "hard to understand" combine matrix list operations needed to do such. 
point being, we need that elsewhere and should be bit more flexible

that may render the extra options in polygon transform useless (or confusing)

I'll post an attempt

JacquesLucke:
Yes, please make a simple prototype, or tell me the inputs/outputs this node should have.
We can have such a node but I think we should keep the pivotType property in the anyway. If you calculate the matrix somewhere else you just have to set the pivot type to World.

og76:
added several input types, default is matrix
(not adding different pivot types, as they would be substitutes of vector math or rot etc)

they are needed for certain cases as list, as you'll make a complex op in a loop and having change pivot for all is not easy (transform  * mat.invert, do stuff, transform * mat)

sure, now that I look at it, may wonder why is that not as an extra in the transform nodes. guess it is possible with an option, so that we still have the basic speedy transform.

![pivot](https://cloud.githubusercontent.com/assets/11709974/13926434/fc9fa926-ef94-11e5-8f3d-65616d167156.PNG)

og76:
as per #428 comments

JacquesLucke:
I'm still trying to understand how to use this node in context.. Can you provide some simple use cases for the different modes?
also, what exactly is a "Pivot Matrix"?

og76:
Pivot Point Matrix, 
(but actually it also changes the orientation/scale .. so it's **more like parent** than just pivot. )

This replaces the 
- combine matrices node case, where you'd have matrix invert, do stuff, matrix
- transform vector by matrix inverted, do stuff, transform by matrix
- if that is the only operation, it also has list variants, to skip a loop just for that (or for a transform Vector List )

For example, the flip polygons around edge case: The node is equivalent to the frame below and also can be used as rotation variant)
![pivot1](https://cloud.githubusercontent.com/assets/11709974/14005097/380c8658-f168-11e5-9a09-daf428e3741a.PNG)


This is where it all started, really (flip polys). And currently, the transform poly does not cover it

2.
**Should I also include Polygon and spline** into the accepted inputs? It kinda pairs with transform nodes. I did not, so that I leave the polygon as is

og76:
I remade the pivot, actually with some variants to see.
but first, have to check dir to rot algo as it is wrong in a case ( #445 )

og76:
#467 
[...]

PR #433, opened 2016-03-21
Create change_pivot.py #429

og76:


og76:
for #429

nodes/matrix/change_pivot.py
```diff
@@ -0,0 +1,63 @@
+import bpy
+from bpy.props import *
+from ... tree_info import keepNodeState
+from ... events import executionCodeChanged
+from ... base_types.node import AnimationNode
+
+inputTypeItems = [
```
JacquesLucke: would be nice to have all 5 values here for each item. Most importantly the last number which will be used as identifier by Blender. In my experience this is more 'stable' when we extend nodes later.

nodes/matrix/change_pivot.py
```diff
@@ -0,0 +1,63 @@
...
+
+    @keepNodeState
+    def generateSockets(self):
+        self.inputs.clear()
+        self.outputs.clear()
+        
+        type = self.inputType
+        condition = "List" if "LIST" in type else ""
+        name = names[type]
+        Name = name.capitalize() + " " + condition
```
JacquesLucke: maybe rename 'Name' to 'displayName' or so

JacquesLucke:
To be hones, I don't understand how the getExecutionCode function works atm but I'll trust you here that it works. I'll play with it later a bit and maybe clean it up a bit so that I can understand it as well :D

nodes/matrix/change_pivot.py
og76: it only make diff between the item/list case. like:

 - matrix = **stuff**
 - matrixList = [ **stuff** for x in matrixList]

so I format the  _" = "_ vs _"List = "_  and at the end _""_(nothing)  vs  _" for x in matrixList]"_

JacquesLucke:
I'm still not sure if "Pivot Matrix" is the best name here but I don't know any better so it is ok.
Maybe we should give the node a more specific name like "Change Matrix Pivot" or "Change Transformation Pivot" because just "Change Pivot" is too unspecific.

og76:
Should i also include it as one of the polygon cases? In the transform poly? That would break the node (version)

Names?
 maybe it should go like convert rotations, having different Different names/search for each case:
 "Change Matrix Pivot",  "Change Rotation Pivot",  "Change Matrix List Pivot" etc

 as it kinda goes along with transform nodes? Problem is that I included more than matrix transform ..

Not sure, maybe I should also have some search tags to include "parent matrix" or "reference" something.

JacquesLucke:
I'm testing the node right now and I don't understand everything.
I mean changing the pivot for a matrix makes sense but no for the rest... Also "Pivot for Vector" doesn't change the vector at all

JacquesLucke:
are the change rotation/vector pivot modes really needed because the seam to be a bit confusion to me

JacquesLucke:
Oh and it is very uneffective to calculate the inverted pivot matrix for each matrix in a list. It's better to calculate it beforehand

PR #435, opened 2016-03-25
Create change_text_case.py

og76:
just a small trick :) but useful

text case


JacquesLucke:
Yes looks useful. 
One problem here: the socket identifiers should be inText and outText
AN notices when the identifier of an input and an output socket are the same and reuses the same variable name in the execution code. 

change text case

(this happens only when you create the nodes in a certain order)

PR #441, opened 2016-03-25
change dir/rot  and rot/dir to use algo

og76:
also:
 - rot to dir has axis to specify 
 - dir to rot outputs eul/quat/matrix if needed

the matrix is also directly 4x4 and normalized for matrix output
normalization is necessary for matrix out, or the scale will be affected too
not for eul, quat ...

(prev ver was not using this way, but 3x3 to 4x4 etc)

### DIFF
--- a/nodes/rotation/direction_to_rotation.py
+++ b/nodes/rotation/direction_to_rotation.py
@@ -2,6 +2,7 @@ import bpy
 from bpy.props import *
 from mathutils import Vector
 from ... events import executionCodeChanged
+from ... algorithms.rotation import generateDirectionToRotationCode
 from ... base_types.node import AnimationNode
 
 trackAxisItems = [(axis, axis, "") for axis in ("X", "Y", "Z", "-X", "-Y", "-Z")]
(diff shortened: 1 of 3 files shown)


nodes/rotation/direction_to_rotation.py
```diff
@@ -28,36 +31,18 @@ def draw(self, layout):
...
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return ""
+        
+        rot = "eulerRotation" if isLinked["eulerRotation"] else ""
+        quat = "quaternionRotation" if isLinked["quaternionRotation"] else ""
+        mat = "matrixRotation" if isLinked["matrixRotation"] else ""
+
+        return generateDirectionToRotationCode("direction", "guide", 
+                                                self.trackAxis, self.guideAxis, 
+                                                matrixOutputName = "{}".format(mat),
```
JacquesLucke: why not simply use: matrixOutputName = mat?

og76: cause I need to normalize it for matrix, output, else the scale is affected too
only for that

og76: for eul/quat normalized matrix is not a must, but for matrix, yes

JacquesLucke: yes but where is the diference between:
"{}".format(mat)
mat

og76: ah, ok
silly, got caught in the style ..

PR #449, opened 2016-03-27
change matrix pivot + menu

og76:


nodes/matrix/change_matrix_pivot.py
```diff
@@ -0,0 +1,77 @@
...
+    ("MATRIX", "Pivot Matrix", "Change Pivot by Matrix, like parent", "", 0),
+    ("VECTOR", "Pivot Location", "Change Pivot Location", "", 3),
+    ("LOC_ROT", "Center and Rotation", "Change Pivot by Center and Custom Rotation", "", 2),
+    ("AXES", "Center and Axes Direction", "Change Pivot by Center and Custom Rotation Axes", "", 1) ]
+
+class ChangeMatrixPivotNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ChangeMatrixPivotNode"
+    bl_label = "Change Matrix Pivot"
+    
+    searchTags = [  ("Change Transform Matrix Pivot", {"pivotType" : repr("MATRIX")}),
```
JacquesLucke: do we really need the search tags here? I think its easier to search for "change matrix pivot" only and then select the type from a dropdown list.

nodes/matrix/change_matrix_pivot.py
```diff
@@ -0,0 +1,77 @@
+import bpy
+from bpy.props import *
+from ... tree_info import keepNodeState
+from ... events import executionCodeChanged
+from ... algorithms.rotation import generateDirectionToRotationCode
+from ... base_types.node import AnimationNode
+
+pivotTypeItems = [
+    ("MATRIX", "Pivot Matrix", "Change Pivot by Matrix, like parent", "", 0),
```
JacquesLucke: I think the iconvalue has to be "NONE" instead of ""
I'm not sure if this here throws exceptions.

nodes/matrix/change_matrix_pivot.py
```diff
@@ -0,0 +1,77 @@
...
+        
+        if type == "MATRIX":
+            self.inputs.new("an_MatrixSocket", "Pivot Matrix (Parent)", "pivotMatrix")
+        if type == "VECTOR":
+            self.inputs.new("an_VectorSocket", "Pivot Location", "pivot")
+        if type == "LOC_ROT":
+            self.inputs.new("an_VectorSocket", "Pivot Center", "pivot")
+            self.inputs.new("an_EulerSocket", "Rotation", "rotation")
+        if type == "AXES":
+            self.inputs.new("an_VectorSocket", "X Start (Center)", "start")
```
JacquesLucke: there are multiple different ways we could organise these axes Inputs right?
like it could also be: center, x direction, z direction
I understand that this mode here is the easiest to use for polygon transformations

og76: we also have center + rotation, where you can plug dir to rot into rotation for full xyz options. (otherwise, that rotation it's not gonna be used .. really). And of course, the matrix from dir to rot..
I can give some options like XY but I thought it clutters the node.

nodes/matrix/change_matrix_pivot.py
og76: can do without

nodes/matrix/change_matrix_pivot.py
og76: It's not the only place like this. guess. .. 
What is the diff to, say, vector math? 
Is there a problem with that comma in the text?

JacquesLucke:
I'll test the node when I come home again and then merge :)

og76:
hm. 
Do you think I should give more options here?
The xxz one is indeed ideal for poly edge flip
Could have 
- dir, guide (xz), 
- center,  Xdir, Zdir, 
- center X start, end, Z start, end
They are all just extra vector math, witch can be a problem for users, so we would save them the trouble.

JacquesLucke:
Yes this can be useful. But I wonder if we can group them somehow... Because just a list with 6 types or so would make this node quite difficult to use.

og76:
is there a way to group inside enum menu? 
or you mean to have a second enum dropdown/option for these under the axes case?

JacquesLucke:
I mean to have a second dropdown or whatever... I'm not sure what's the best way to group them.
Or maybe we should just leave most of these extra cases out. Most of them are just one Vector Math node difference right?

og76:
it's just a convenience for users not into vector math, more artist friendly ..
I'll give it a try with an extra menu to see how it looks, then we can finish this

PR #467, opened 2016-04-02
add change matrix pivot

og76:


JacquesLucke:
Maybe I'll change some names and clean stuff up but overall it looks good! thx

og76:
sure
just keep the Transform Matrix sockets, as a hint that is dedicated to transform nodes.
A good searchTag would be Change Transform Pivot or so...

PR #474, opened 2016-04-04
List fill

og76:
add list fill node, just like text fill
+small change in text fill

is isdataModified needed for these input sockets?


JacquesLucke:
Unfortunally this node is not as easy as it seems as first. Some data copies are needed here is the elements are copyable.

Here is the problem you get using your method:
fill list issue

og76:
just fooling around and experimenting. 
but it does work like this

fill with copy

PR #475, opened 2016-04-04
add repeat list node

og76:


og76:
options being
repeat list1

altough length instead of amount is more of a convenience, math is always there

JacquesLucke:
I really appreciate your work but the problem here is the same as in the other node.

og76:
is that an issue of `.dataIsModified = True` ?
I don't think is needed here, and on the other, especially on the element ..

JacquesLucke:
no that is not the issue. The problem is the following.
let's say we have a vector object v = Vector((0, 0, 0))
now you create a list: outList = [v] * 100
what you get is a list of vector objects as expected BUT all vectors in this list are the same object in memory. so if you change one of them you change all. (I tried to show that in the image for the other node)

The only way to do this correctly for vectors is to use the "Vector.copy()" function on each new element.

og76:
I guess for lists (here) _a [:] for every instance_ or _expand_ would do then

JacquesLucke:
No, doing [:] will make a copy of the list object but not of the individual objects inside.

To deal with these problems each socket type has a getCopyExpression function. But as I said in the other node. I'd prefer to make this myself.... because maybe I have to change something in the overall structure in this area.

og76:
got it. 
you could work on top of them, you know, to keep the incredibly slick design ... :D

og76:
just like #474 fooling around, experimenting to makin it work

I guess atm this is the kind of approach to copy elements. 
(I know I could have used getCopyExpression in execution/code_gen, but here I also use it for other than socket var name, in ping pong type, so I made it as a def here, with name (string) to be specified)

repeat with copy


only this one needed more small corrections anyway

og76:
well, now I have the same simplified def that works for both element/list cases (and both nodes)
```
    def getCopyString(self, fromSocket, variableString):
        if not fromSocket.isCopyable: return variableString
        return fromSocket.getCopyExpression().replace("value", variableString)
```
I guess this should go into sockets/info, 
(just not to pollute code_generator that is dedicated to final code)

the main diff from getCopyExpression or getCopyLine in code_generator is that
**this one allows any variable name**. It is important for cases where I do not copy the input, but another var, later in the code.

(the example here is ping-pong case, where I need to copy inList, but also a reversed inList
then need to copy in the loop each time)

JacquesLucke:
thanks for your tests and improvement ideas! I actually got a good idea when I read you write up here how to make that stuff working better.
I think that will change the sockets/info.py file quite a lot, but  it'll be worth it.
Will work on it as soon as I have time...

JacquesLucke:
I'll reimplement this node now.
In my opinion "Loop" and "Ping Pong" modes should be enough for now... You can always shuffle the list afterwards, the result might be slightly different but it's much faster.

Issue #477, opened 2016-04-05
Speedup Execution Code Creation


JacquesLucke:

JacquesLucke:
I will need to test more different files to make sure that I didn't break anything, because I changed quite a lot in the code creation area.

PR #478, opened 2016-04-05
Code creation Speedup

JacquesLucke:
This speeds up the execution code creation. The larger the node tree the more this will become apparent.
100 nodes -> roughly 2x faster
800 nodes -> up to 3.5x faster


JacquesLucke:
#477

PR #487, opened 2016-04-08
Boolean with dyna labels

og76:
and advanced assign type chooser

### DIFF
--- a/nodes/boolean/compare.py
+++ b/nodes/boolean/compare.py
@@ -41,7 +41,7 @@ class CompareNode(bpy.types.Node, AnimationNode):
         return label
 
     def drawAdvanced(self, layout):
-        self.invokeSocketTypeChooser(layout, "assignAdvancedDataType",
+        self.invokeSocketTypeChooser(layout, "assingType",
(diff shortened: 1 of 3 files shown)


nodes/boolean/compare.py
```diff
@@ -50,6 +68,9 @@ def getWantedDataType(self):
         if inputB is not None: return inputB.dataType
         return self.inputs[0].dataType
 
+    def assignAdvancedDataType(self, dataType):
```
JacquesLucke: why did you implement this function? Instead of directly invoking the assignType function.

og76: oh, guess reflex, copy/paste ..

nodes/boolean/compare.py
```diff
@@ -8,9 +8,13 @@
 compare_types = ["A = B", "A != B", "A < B", "A <= B", "A > B", "A >= B", "A is B"]
 compare_types_items = [(t, t, "") for t in compare_types]
 
+compareLabels = {item[0] : item[0] for item in compare_types_items}
```
JacquesLucke: this is used nowhere

PR #490, opened 2016-04-08
atan2, hypo and more label

og76:
seams not to break prev

### DIFF
--- a/nodes/number/float_math.py
+++ b/nodes/number/float_math.py
@@ -16,6 +16,7 @@ operationItems = [
     ("ARCSINE", "Arcsine", "asin A", "", 7),
     ("ARCCOSINE", "Arccosine", "acos A", "", 8),
     ("ARCTANGENT", "Arctangent", "atan A", "", 9),
+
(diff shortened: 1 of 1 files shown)


og76:
should we also allow X,Y instead of AB for sockets?

nodes/number/float_math.py
```diff
@@ -16,6 +16,7 @@
     ("ARCSINE", "Arcsine", "asin A", "", 7),
     ("ARCCOSINE", "Arccosine", "acos A", "", 8),
     ("ARCTANGENT", "Arctangent", "atan A", "", 9),
+
```
JacquesLucke: was this new line intended?

JacquesLucke:
I'm thinking... Yeah looks like, A and B make no sense for the atan2 operation..
So yes, we can use X, Y here.

nodes/number/float_math.py
og76: actually, I had others after 3 and 18 to separate more visually (the list is soo long)
I can remove it if u think is useless, 
or add the others (to split in math/trigo/ etc)

PR #498, opened 2016-04-09
Bool between and any all

og76:


og76:
now as 2 new nodes, like this:
bool between any all

questions are:

1. should they **be separate nodes** ? or _ between as part of compare_ and _any/all in logic_ ?
 (they kinda belong there, but I'll have to see if the socket generation will break prev )
2. names. as separate nodes. 
 is between? or compare between? 
 boolean list logic? compare boolean list?  .. this one sucks
3. Should I remove the negate in the between? It's handy but we can invert later ..

JacquesLucke:
No sure about the Is Between node yet... And yes, the negate option should be removed. 

I'll only implement the bool list node for now

og76:
I found between to be quite handy instead of 3 nodes
so takin negation out

+it actually is faster to say `0 < a < 1` than `0 < a and a < 1` cause it stops at 0<a if not true.
(not that it will really make a difference)

PR #505, opened 2016-04-12
adding capitalize words (ok?)

og76:
is this ok?

### DIFF
--- a/nodes/text/change_text_case.py
+++ b/nodes/text/change_text_case.py
@@ -5,8 +5,9 @@ from ... base_types.node import AnimationNode
 
 caseTypeItems= [("UPPER", "To Upper Case", ""),
                 ("LOWER", "To Lower Case", ""),
+                ("TITLE", "To Title Case", ""),
                 ("CAPITALIZE", "Capitalize Phrase", ""),
-                ("CAPITALIZE_WORDS", "Capitalize Words", "") ]
+                ("CAPWORDS", "Capitalize Words", "") ]
 
 caseTypeCode = { item[0] : item[0].lower() for item in caseTypeItems }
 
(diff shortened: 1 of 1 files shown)


nodes/text/change_text_case.py
```diff
@@ -28,4 +29,6 @@ def draw(self, layout):
         layout.prop(self, "caseType", text = "")
 
     def getExecutionCode(self):
+        if self.caseType == "CAPITALIZE_WORDS":
+            return "outText = ' '.join( [t.capitalize() for t in inText.split()])"
```
JacquesLucke: Two things:
- you don't need the quare brackets here for the list comprehension. It works without them.
- let's just use this function: https://docs.python.org/3.5/library/stdtypes.html#str.title

nodes/text/change_text_case.py
```diff
@@ -5,7 +5,8 @@
 
 caseTypeItems= [("UPPER", "To Upper Case", ""),
                 ("LOWER", "To Lower Case", ""),
-                ("CAPITALIZE", "Capitalize Phrase", "") ]
+                ("CAPITALIZE", "Capitalize Phrase", ""),
+                ("CAPITALIZE_WORDS", "Capitalize Words", "") ]
```
JacquesLucke: oh and since you're working on it right now, can you also set the enum item values here? like "'NONE', 0)" ...

nodes/text/change_text_case.py
JacquesLucke: Maybe we can have two options here... check this: http://stackoverflow.com/a/1549644/4755171
(also the comments below this post)

og76:
Thanks for the heads up,  `capwords()` it is

Does it make sense to also have the Title?
is kinda same, but lame. ?

now is like
title

JacquesLucke:
I think we can have both :) Capwords and Title

og76:
Then the modification is already in with both, (just like the pic)

LeonCheung:
Just tried the latest version, it appears more powerful. Thanks guys!

tt

og76:
yeah, but be aware that
 Capitalize Words works by spaces (like split by words, leon's becoming Leon's)
 Title case works by groups of letters (not good on leon's becoming Leon'S)

PR #510, opened 2016-04-12
correction for capwords, doh

og76:


og76:
sorry, did not realize this

Or I could just drop the string module and use the same:
`outText = ' '.join( t.capitalize for t in inText.split() )`

JacquesLucke:
no it's ok now :)
thx for the fix

PR #528, opened 2016-05-11
intersect line line

og76:


nodes/geometry/intersect_line_line.py
```diff
@@ -0,0 +1,40 @@
+import bpy
+from ... base_types.node import AnimationNode
+
+class IntersectLineLineNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_IntersectLineLineNode"
+    bl_label = "Intersect Line Line"
+    bl_width_default = 160
+    searchTags = ["Closest Points on 2 Lines"]
+        
+    def create(self):
```
JacquesLucke: why do you use such uncommon default values? Do they have a special meaning?
Maybe the standart vectors (0, 0, 0) as start and (1, 0, 0) / (0, 1, 0) as end would be better.

for better readability I'd prefer the identifiers lineStart1, lineEnd1, ... what do you think?

nodes/geometry/intersect_line_line.py
```diff
@@ -0,0 +1,40 @@
...
+        self.newInput("Vector", "Line 1 Start", "line1Start").value = [-2, 0, 0]
+        self.newInput("Vector", "Line 1 End", "line1End").value = [2, 0, 0]
+        self.newInput("Vector", "Line 2 Start", "line2Start").value = [1, -1, 0]
+        self.newInput("Vector", "Line 2 End", "line2End").value = [1, 1, 0]
+        
+        self.newOutput("Vector", "Intersection (Closest) 1", "int1")
+        self.newOutput("Vector", "Intersection (Closest) 2", "int2")
+        self.newOutput("Boolean", "Is Valid", "isValid")
```
JacquesLucke: I prefer 'Intersect' instead of 'Is Valid'..

And "Closest Line 1" + "Closest Line 2" for the vector outputs. Also 'int#' is not a good name... When I hear 'int', I always think about integer. Just name them 'closest1', ..

nodes/geometry/intersect_line_line.py
```diff
@@ -0,0 +1,40 @@
...
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return ""
+    
+        intersection1  = isLinked["int1"]
+        intersection2  = isLinked["int2"]
+        isValid = isLinked["isValid"]
+        
+        yield "int = mathutils.geometry.intersect_line_line(line1Start, line1End, line2Start, line2End)"
+        
+        yield "if int is None:"
```
JacquesLucke: I have not tested it, when does the intersect_line_line function return None?

nodes/geometry/intersect_line_line.py
```diff
@@ -0,0 +1,40 @@
...
+        
+    def getExecutionCode(self):
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return ""
+    
+        intersection1  = isLinked["int1"]
+        intersection2  = isLinked["int2"]
+        isValid = isLinked["isValid"]
+        
+        yield "int = mathutils.geometry.intersect_line_line(line1Start, line1End, line2Start, line2End)"
```
JacquesLucke: 'int' -> 'closestPoints'

JacquesLucke:
the general idea of the suggested changes is that this node will be more a node to find the close points instead of only the intersection.. Not sure if it should be renamed..

og76:
It is true that it actually get the closest points, but I'm thinking to keep the name in line with the other intersect... . This way they are more organized in a way

I would also keep is Valid as all the other nodes use it. 
(At most I could use can Compute, but not is intersecting, as actual intersection is only part of the cases)

None goes in many situations, lines parallel or colinear, start = end (x2) etc
(too many to create exceptions or warnings)

what do you think?

PR #530, opened 2016-05-11
Intersect line sphere

og76:


nodes/geometry/intersect_line_sphere.py
```diff
@@ -0,0 +1,51 @@
...
+        isValid2 = isLinked["isValid2"]
+        
+        yield "if lineStart == lineEnd:"
+        if intersection1 : yield "    intersection1 = center"
+        if intersection2 : yield "    intersection2 = center"
+        if isValid1: yield "    isValid1 = False"
+        if isValid2: yield "    isValid2 = False"
+        
+        yield "else:"
+        yield "    int1, int2 = mathutils.geometry.intersect_line_sphere(lineStart, lineEnd, center, radius, self.clip)"
```
JacquesLucke: 'int#' is not a very good name again

nodes/geometry/intersect_line_sphere.py
```diff
@@ -0,0 +1,51 @@
...
+from bpy.props import *
+from ... events import propertyChanged
+from ... base_types.node import AnimationNode
+
+class IntersectLineSphereNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_IntersectLineSphereNode"
+    bl_label = "Intersect Line Sphere"
+    bl_width_default = 160
+    
+    clip = BoolProperty(default = False, update = propertyChanged,
```
JacquesLucke: should be called "Clip" in the zu instead of lower cased "clip"

PR #531, opened 2016-05-11
Intersect polyline plane

og76:
This is a very special one
I could make the edges as an option to actually have vectors or vectors+edges, but somehow I prefer it this way

### DIFF
--- a/nodes/geometry/intersect_polyline_plane.py
+++ b/nodes/geometry/intersect_polyline_plane.py
@@ -22,17 +22,17 @@ class IntersectPolylinePlaneNode(bpy.types.Node, AnimationNode):
         default = True, update = executionCodeChanged)
         
     def create(self):
-        self.newInput("Vector", "Positions", "positions")
+        self.newInput("Vector List", "Positions", "positions")
         self.newInput("Edge Indices List", "Edge Indices", "edges")
         self.inputs[1].useIsUsedProperty = True
         self.inputs[1].isUsed = False
         self.newInput("Vector", "Plane Point", "planePoint")
-        self.newInput("Vector", "Plane Normal", "planeNormal").value = (0, 0, 1)
+        self.newInput("Vector", "Plane Normal", "planeNormal", value = (0, 0, 1))
 
         self.newOutput("Vector List", "Intersections List", "intersections")
-        self.newOutput("Integer List", "Intersected Edge Index", "cutEdges").hide = True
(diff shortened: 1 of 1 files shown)


nodes/geometry/intersect_polyline_plane.py
```diff
@@ -0,0 +1,130 @@
...
+    
+    edgesType = EnumProperty(name = "Plane Type", default = "POINTS",
+        items = edgesTypeItems, update = edgesTypeChanged)
+    message = StringProperty(name = "Message", default = "Expecting Points")
+    cyclic = BoolProperty(name = "Cyclic points", description = "Consider last point to first point also", 
+        default = True, update = executionCodeChanged)
+        
+    def create(self):
+        self.newInput("Vector", "Positions", "positions")
+        self.newInput("Edge Indices List", "Edge Indices", "edges")
```
JacquesLucke: the convention I use everywhere is:
socket = self.newInput(...)
socket.setting1 = ...
socket.setting2 = ...

JacquesLucke:
this node looks more complicated than the others, I will have to check it in more detail a bit later.

og76:
for examples see the issue #430 

as formula, it does:
- define edges as vecs in order or from edges (for different use, as poly or edges of an obj)
- (pre)check if the edges are cut (very important, see below)
- inters only if the case

the dot  check is important cause:
- it reduces the inters ops to those needed
- it provides position relative to plane normal  - / 0 / + (helps filter if u want to keep only above/below)
- for coplanar, I return the first point as the intx, so if the poly is in the plane it returns original points (the inters func itself does not)

I'll probably make it into an algo later, but for now I'll let it inside the node, with yeld etc

nodes/geometry/intersect_polyline_plane.py
```diff
@@ -0,0 +1,130 @@
...
+    bl_label = "Intersect Polyline Plane"
+    bl_width_default = 160
+    
+    def edgesTypeChanged(self, context):
+        executionCodeChanged()
+    
+    edgesType = EnumProperty(name = "Plane Type", default = "POINTS",
+        items = edgesTypeItems, update = edgesTypeChanged)
+    message = StringProperty(name = "Message", default = "Expecting Points")
+    cyclic = BoolProperty(name = "Cyclic points", description = "Consider last point to first point also",
```
JacquesLucke: capitalize to "Cyclic Points"

og76:
ok?

PR #532, opened 2016-05-11
Intersect plane plane

og76:


nodes/geometry/intersect_plane_plane.py
```diff
@@ -0,0 +1,50 @@
...
+from ... base_types.node import AnimationNode
+
+class IntersectPlanePlaneNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_IntersectPlanePlaneNode"
+    bl_label = "Intersect Plane Plane"
+    bl_width_default = 160
+
+    def create(self):
+        self.width = 160
+        self.newInput("Vector", "Plane 1 Point", "p1_co")
```
JacquesLucke: underscores in identifiers is very uncommon.

JacquesLucke:
will also test this when I'm at home. But looks mostly ok.
maybe you don't need to define the 'zero' string, just write mathutils.Vector(..)
(not even sure if the mathutils is still needed; I put 'Vector', 'Matrix', ... in  global namespace shortly, but I'm not sure in which version atm^^)

PR #533, opened 2016-05-11
Intersect plane sphere

og76:


nodes/geometry/intersect_sphere_plane.py
```diff
@@ -0,0 +1,39 @@
+import bpy
+from ... base_types.node import AnimationNode
+
+class IntersectSpherePlaneNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_IntersectSpherePlaneNode"
+    bl_label = "Intersect Sphere Plane"
+    bl_width_default = 160
+    
+    def create(self):
+        self.newInput("Vector", "Sphere Center", "c")
```
JacquesLucke: identifier should be sphereCenter
(same with radius)

PR #534, opened 2016-05-11
Intersect sphere sphere

og76:


nodes/geometry/intersect_sphere_sphere.py
```diff
@@ -0,0 +1,54 @@
+import bpy
+from ... base_types.node import AnimationNode
+
+class IntersectSphereSphereNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_IntersectSphereSphereNode"
+    bl_label = "Intersect Sphere Sphere"
+
+    def create(self):
+        self.width = 160
+        self.inputs.new("an_VectorSocket", "Sphere 1 Center", "c1")
```
JacquesLucke: I'd prefer longer identifiers here too

PR #535, opened 2016-05-11
barycentric transform

og76:
like an adaptive for tris
transform a vector (or list) by source 3 point to target 3 point


nodes/geometry/barycentric_transform.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @keepNodeState
+    def generateSockets(self):
+        self.inputs.clear()
+        self.outputs.clear()
+        
+        if self.operationType == "POINT":
+            self.newInput("Vector", "Location", "point")
+            self.newInput("Vector List", "Source Triangle Points", "t1")
+            self.newInput("Vector List", "Target Triangle Points", "t2")
```
JacquesLucke: more detailed identifiers. eg 'sourceTrianglePoints'

PR #536, opened 2016-05-11
point list normal

og76:
just calculates normal for an ordered list of points, giving the poly normal

### DIFF
--- a/nodes/geometry/point_list_normal.py
+++ b/nodes/geometry/point_list_normal.py
@@ -1,0 +1,31 @@
+import bpy
+from bpy.props import *
+from ... utils.layout import writeText
+from ... base_types.node import AnimationNode
+
+class PointListNormalNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_PointListNormalNode"
+    bl_label = "Point List Normal"
+    bl_width_default = 160
+    
+    searchTags = ["Points Normal", "Calculate Normal"]    
+    errorMessage = StringProperty()
+        
+    def create(self):
+        self.width = 160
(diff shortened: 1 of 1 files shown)


nodes/geometry/point_list_normal.py
```diff
@@ -0,0 +1,31 @@
...
+from bpy.props import *
+from ... utils.layout import writeText
+from ... base_types.node import AnimationNode
+
+class PointListNormalNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_PointListNormalNode"
+    bl_label = "Point List Normal"
+    bl_width_default = 160
+    
+    searchTags = ["Points Normal", "Calculate Normal"]
```
JacquesLucke: not sure if we need so many search tags for such a small node..

nodes/geometry/point_list_normal.py
```diff
@@ -0,0 +1,31 @@
...
+        self.newOutput("Vector", "Normal", "normal")
+        self.newOutput("Boolean", "Is Valid", "isValid")
+        
+    def draw(self, layout):
+        if self.errorMessage != "":
+            writeText(layout, self.errorMessage, icon = "ERROR", width = 20)
+        
+    def getExecutionCode(self):
+        yield "if len(points) > 2: normal, self.errorMessage = mathutils.geometry.normal(points), ''"
+        yield "else: normal, self.errorMessage = mathutils.Vector((0, 0, 0)), 'Expected min 3 different vectors'"
```
JacquesLucke: please split this up into multiple lines

PR #537, opened 2016-05-11
tesselate triangulate

og76:
this is a very interesting one to work with baricentric (tris) transform. It can help deform mesh cage or lattice or similar, and later for uv's and such

It has all the variants from simple vector or polys to  list of polygons or indices, and bmesh

there is a tricky part of it, that I don't like much, that there are 3 formulas to calc this (besides bmesh that works great)
 I would have included only the first as it is far faster, but somehow is not always accurate in terms of resulting normal (order of indices)

I may have another formula for triangulation later, but we'll see

### DIFF
--- a/nodes/geometry/tesselate_polygons.py
+++ b/nodes/geometry/tesselate_polygons.py
@@ -1,0 +1,250 @@
+import bpy
+from bpy.props import *
+from ... utils.layout import writeText
+from ... tree_info import keepNodeState
+from ... events import executionCodeChanged, propertyChanged
+from ... base_types.node import AnimationNode
+from bmesh.ops import triangulate
+from mathutils.geometry import tessellate_polygon
+from bpy_extras.mesh_utils import ngon_tessellate
+
+
+polyTypeItems = [
+    ("VECTORS", "Vectors in order", "", "", 0),
+    ("INDICES", "Simple Poly Index", "", "", 1),
+    ("INDICES_LIST", "Poly Indices List", "", "", 2),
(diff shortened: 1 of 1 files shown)


JacquesLucke:
will check this a bit later

JacquesLucke:
Is it ok to not have this in the next release? I don't find the time to check the correctness of that much 'indirect code'..

og76:
I'd prefer to have it along with barycentric (and any other that works with tris)
 with it I can do quite some tricks, like mesh/cage deform etc

as for the code, I guess the developer / write execution code can always make it clear what is there in the end 
 plus testing for actual working cases.
 I can make some blends if it's easyer ..

JacquesLucke:
Yes would be nice if you could create some very simple test files for the different scenarios.

og76:
[tesselator.blend1](https://drive.google.com/file/d/0BzzAoBwXO8dsX0J1cUE5NWFROGc/view?usp=sharing)

a quick one for all cases

JacquesLucke:
thx, will check tomorow and merge if everything works :)
Is there a specific reason why you did not call the node an_TriangulatePolygonsNode?

Looks like a lot of options, will see if we really need all of them, but let me check first.

JacquesLucke:
Until then you can start writing documentation for all the other nodes you implemented hehe

JacquesLucke:
Will merge this now. For you mainly :D
I don't think that so many people will use it.. This node is VERY likely to change in the next release.

PR #540, opened 2016-05-13
Bmesh mesh data

og76:
just like object mesh data

vert vgroups returns [] as vgroups are in object

to have vgroups "carried by bmesh" we can add a vert custom layer in bmesh from object

### DIFF
--- a/data_structures/mesh.py
+++ b/data_structures/mesh.py
@@ -92,7 +92,7 @@ class Polygon:
         return Polygon(
                  vertexLocations,
                  bmeshFace.normal,
-                 bmeshFace.calc_center_median(),
+                 bmeshFace.calc_center_median_weighted() if centerWeighted else bmeshFace.calc_center_median(),
                  bmeshFace.calc_area(),
                  bmeshFace.material_index)
  
--- a/data_structures/mesh.py
+++ b/data_structures/mesh.py
@@ -54,6 +54,13 @@ class Vertex:
     __slots__ = ("location", "normal", "groupWeights")
 
     @staticmethod
+    def fromBMeshVert(bmeshVert):
+        return Vertex(
+                 bmeshVert.co,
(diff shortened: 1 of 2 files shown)


data_structures/mesh.py
```diff
@@ -80,6 +87,16 @@ class Polygon:
     __slots__ = ("vertexLocations", "normal", "center", "area", "materialIndex")
 
     @staticmethod
+    def fromBMeshFace(bmeshFace, centerWeighted = False):
```
JacquesLucke: why is there a centered weight parameter?

nodes/mesh/bmesh_mesh_data.py
```diff
@@ -0,0 +1,66 @@
...
+        self.newOutput("Vertex List", "Vertices", "vertices")
+        self.newOutput("Polygon List", "Polygons", "polygons")
+
+    def getExecutionCode(self):
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return
+
+#        yield "meshName = ''"
+#        yield "if getattr(object, 'type', '') == 'MESH':"
+#        yield "    mesh = self.getMesh(object, useModifiers, scene)"
```
JacquesLucke: you should remove code that is not used.

data_structures/mesh.py
og76: ah, I had another variant cause you can calc center as median or median weighted with an advanced option

I'll test that again to make sure we need it

nodes/mesh/bmesh_mesh_data.py
og76: sorry, I'll del all commented

og76:
may remove the weigthed center variant if affects performance in any way
not sure how useful it really is, and the blender / mesh poly center is the other

Issue #543, opened 2016-05-14
Suggestion: Subprogram Template Improvement


LeonCheung:
If users want to create subprogram from Template menu (e.g, loop), they have to first, click on the Templates, then Empty Subprogram, then click on the dropdown menu, then click on Loop, then click OK, then click again to confirm position. 6 clicks in Total...

Possible solutions for a better workflow:

A. Use three buttons instead of a dropdown menu.
B. Make them as three sliblings to Templates. (not recommended for sure)
C. Move "Empty Subprogram" out of Templates menu, make it a separate 3-item menu.

Results:

A. 4 clicks needed.
B. 2 clicks needed.
C. 3 clicks needed.

I know it appears not kinda high priority issue. FYI.

og76:
agreed. 

- should it be 3 buttons (add ..) ?
- or a separate menu next to template (the complete subprograms menu, as in shiftA)?
I used to do that in my branches ..

og76:
var 1
![templates sub](https://cloud.githubusercontent.com/assets/11709974/15272782/e1ab954e-1a8c-11e6-95bd-6a25be64e2e3.PNG)

var 2
![templates sub2](https://cloud.githubusercontent.com/assets/11709974/15272784/ea894e7c-1a8c-11e6-9366-514a9acd7eb9.PNG)

var 3
![templates sub3](https://cloud.githubusercontent.com/assets/11709974/15272785/ef4f2daa-1a8c-11e6-88f5-047591f3ca41.PNG)

(I also include the struct part into the add / subprograms menu anyway, point is if we need it here too..)

I kinda prefer 2 or 3

og76:
I'm modifying the main menu anytway to add missing nodes etc, so it's a good time to do this also 
So?

JacquesLucke:
yes we can have the subprograms menu at the bottom next to the Templates menu.
I wonder, why did you insert the Struct nodes there? I don't think that they belong there.

og76:
where else? yet another dedicated menu? they are very .. generic ..
lists? 
I also redo the menu so.. ?

btw, I guess the subprogram menu should be just the one subprograms _menu py and in the node menu I should import it? 
It's basically duplicated, same class ..

JacquesLucke:
[...]

PR #544, opened 2016-05-15
Menu 1_6

og76:
menu and template subs


og76:
ok, no import, as per #543

PR #545, opened 2016-05-16
boolean compare between

og76:
any chance here?


JacquesLucke:
hmm not absolutely sure... I think it would be better to use the expression node for this kind of operations.

expression between

PR #546, opened 2016-05-18
rest of geom in menu

og76:
just that

### DIFF
--- a/ui/node_menu.py
+++ b/ui/node_menu.py
@@ -523,6 +523,7 @@ class GeometryMenu(bpy.types.Menu):
         layout = self.layout
         insertNode(layout, "an_ConvertPlaneTypeNode", "Point/Normal to Matrix", {"conversionType" : repr("POINT_NORMAL_TO_MATRIX")})
         insertNode(layout, "an_ConvertPlaneTypeNode", "Matrix to Point/Normal", {"conversionType" : repr("MATRIX_TO_POINT_NORMAL")})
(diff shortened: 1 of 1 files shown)


ui/node_menu.py
```diff
@@ -531,8 +532,13 @@ def draw(self, context):
         insertNode(layout, "an_IntersectLinePlaneNode", "Intersect Line Plane")
         insertNode(layout, "an_IntersectLineSphereNode", "Intersect Line Sphere")
         insertNode(layout, "an_IntersectPlanePlaneNode", "Intersect Plane Plane")
+        insertNode(layout, "an_IntersectSpherePlaneNode", "Intersect Plane Sphere")
+        insertNode(layout, "an_IntersectSphereSphereNode", "Intersect Sphere Sphere")
         layout.separator()
         insertNode(layout, "an_IntersectPolylinePlaneNode", "Intersect Polyline Plane")
+        layout.separator()
+        insertNode(layout, "an_BarycentricTransformNode", "Barycentric Transform")
+        insertNode(layout, "an_BarycentricTransformNode", "Barycentric Transform List", {"operationType" : repr("LIST")})
```
JacquesLucke: I don't think that we need both here. the first one should be enouph.

og76: I was imitating vector transform / transform list
for me that comes handy sometimes, and here as default is not list
however, if u want it out, out it is

PR #555, opened 2016-05-28
separate bmesh triangulate / tesselate

og76:


nodes/geometry/bmesh_tesselate.py
```diff
@@ -0,0 +1,43 @@
...
+from ... events import propertyChanged
+from ... base_types.node import AnimationNode
+
+class BMeshTessellateNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_BMeshTessellateNode"
+    bl_label = "Triangulate BMesh"
+    bl_width_default = 160
+    searchLabels = ["Tesselate BMesh"]
+
+    quad = IntProperty(name = "Quad Method", update = propertyChanged, max = 3, min = 0)
```
JacquesLucke: I hate it that these methods are undocumented, but I also don't see a better way atm..

nodes/geometry/bmesh_tesselate.py
```diff
@@ -0,0 +1,43 @@
...
+    ngon = IntProperty(name = "Ngon Method", update = propertyChanged, max = 1, min = 0)
+
+    def create(self):
+        self.newInput("BMesh", "BMesh", "bm")
```
JacquesLucke: is the original bmesh object changed? If yes then 'dataIsModified' is needed here

nodes/geometry/bmesh_tesselate.py
```diff
@@ -0,0 +1,43 @@
...
+        self.newOutput("BMesh", "BMesh", "bm")
+        self.newOutput("Integer List", "Matching Ngon Indices", "ngonIndices", hide = True)
```
JacquesLucke: what is this used for? Do you have an example?
Is there maybe a better (more descriptiv) name?

nodes/geometry/bmesh_tesselate.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def getExecutionCode(self):
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return
+        
+        yield "faces = bm.faces[:]"
+        
+        # do not invert isLinked order, this needs original faces
+        if isLinked["ngonIndices"]: 
+            yield "ngonIndices = list( range(len(faces)) )"
```
JacquesLucke: just a little thing: these exra spaces aren't necessary in my opinion. Sometimes it is helpful to make code clearer this way, but I think this case here is not special enouph to break the (unwritten) rules.

nodes/geometry/bmesh_tesselate.py
og76: they kinda switch the tris, but it's not clear to me how
I was wondering if I should keep them for advanced, I mean how often are they gonna be used?

nodes/geometry/bmesh_tesselate.py
og76: not as far as I tested, 
I'll test again for more cases, loops etc

nodes/geometry/bmesh_tesselate.py
JacquesLucke: yes, they should go into advanced

nodes/geometry/bmesh_tesselate.py
JacquesLucke: If the input and output are different bmesh objects than the socket identifier has to be different as well, but I think the original data is modified. Will test as well

JacquesLucke: bmesh copy
 dataIsModified has to be set to True.

nodes/geometry/bmesh_tesselate.py
og76: Source ngon ..? Original ngon ..? Initial ngon indices?

it shows original poly that tri belongs to
like
source ngon index

can use this if you need to ref the original face, if you only use tria in a hidden way (in the end use full mesh) etc

JacquesLucke: ah I see, thx

nodes/geometry/bmesh_tesselate.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def drawAdvanced(self, layout):
+        layout.prop(self, "quad")
+        layout.prop(self, "ngon")
+
+    def getExecutionCode(self):
+        isLinked = self.getLinkedOutputsDict()
+        if not any(isLinked.values()): return
+        
+        yield "faces = bm.faces[:]"
```
JacquesLucke: last thing: is it important to make this copy?

og76:
guess it works without copy, 
(never sure, I'm not that familiar with bmesh ways .., just get inspiration from here and there)
I don't remember, maybe I hoped it will keep intact after the triangulation, but not ..

JacquesLucke:
ok.
yet another last thing: I think we should either use the term "tesselate" or "triangulate", not both. And I think "triangulate" is better because more people already know what it means.

og76:
u mean no search tags?
and should I change the py file name also?

JacquesLucke:
yes and yes.
for both files

JacquesLucke:
thanks for your efforts. I think that was the last thing for v1.6

PR #559, opened 2016-06-02
add bmesh limited dissolve

og76:


nodes/geometry/bmesh_limited_dissolve.py
```diff
@@ -0,0 +1,27 @@
+import bpy
+from bpy.props import *
+from ... events import propertyChanged
```
JacquesLucke: this Import  is not needed

JacquesLucke:
I'd prefer a normal execution function instead of getExecutionCode.

JacquesLucke:
getExecutionCode should only be used if it has real benefits. otherwise it just adds complexity

JacquesLucke:
what happens when the angle is negative/very large?

PR #585, opened 2016-08-14
Get Active Camera

s-leger:
Hi, 
Missing this one : GetActiveCameraNode.
For things like orienting patches or frustrum reactions. 
So here it is.

### DIFF
--- a/nodes/object/utility_nodes/get_active_camera.py
+++ b/nodes/object/utility_nodes/get_active_camera.py
@@ -1,0 +1,16 @@
+import bpy
+from .... events import isRendering
+from .... base_types.node import AnimationNode
+
+class GetActiveCameraNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetActiveCameraNode"
+    bl_label = "Get Active Camera"
+    searchTags = ["Get Active Camera"]
+    bl_width_default = 200
+
+    def create(self):
+       self.newOutput("Object", "Active Camera", "activeCamera")
+
+    def execute(self):
+        context = bpy.context
+        return context.scene.camera
--- a/nodes/object/utility_nodes/get_active_camera.py
+++ b/nodes/object/utility_nodes/get_active_camera.py
@@ -6,7 +6,7 @@ class GetActiveCameraNode(bpy.types.Node, AnimationNode):
     bl_label = "Get Active Camera"
     
     def create(self):
-       self.newInput("Scene", "scene", "scene", value = bpy.context.scene)
+       self.newInput("Scene", "scene", "scene", hide = True)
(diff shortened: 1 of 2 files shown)


nodes/object/utility_nodes/get_active_camera.py
```diff
@@ -0,0 +1,16 @@
+import bpy
+from .... events import isRendering
+from .... base_types.node import AnimationNode
+
+class GetActiveCameraNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetActiveCameraNode"
+    bl_label = "Get Active Camera"
+    searchTags = ["Get Active Camera"]
```
JacquesLucke: The search tag is not needed when it is the same as the bl_label

nodes/object/utility_nodes/get_active_camera.py
```diff
@@ -0,0 +1,16 @@
+import bpy
+from .... events import isRendering
```
JacquesLucke: this import is not used, can be removed

nodes/object/utility_nodes/get_active_camera.py
```diff
@@ -0,0 +1,16 @@
...
+from .... base_types.node import AnimationNode
+
+class GetActiveCameraNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetActiveCameraNode"
+    bl_label = "Get Active Camera"
+    searchTags = ["Get Active Camera"]
+    bl_width_default = 200
```
JacquesLucke: not sure if it is necessary to make the node wider than the default

JacquesLucke:
We can have a node like this but it needs some modifications.
We cannot/should not access the bpy.context variable in the execute method. This can lead to errors during rendering because rendering should be context free.
Furthermore an active camera is always part of a scene. So we should have a scene input. This also remove the need for the context access. The scene input socket can be hidden because it will not be used very often.
And don't forget to insert a check if the input scene is None.
or just use ``getattr(scene, "camera", None)``; it will return None when the scene is None as well.

Thanks for the patch anyway! This these little correction I will merge it into master.

JacquesLucke:
One last thing about the scene socket. Please do it like here: https://github.com/JacquesLucke/animation_nodes/blob/master/nodes/animation/time_info.py#L10

The Scene socket has a builtin mechanism to choose the active scene (well not the active, but every node tree has a "main scene" which you can change in the Animation Node Tree panel in the left side bar. Also you can hide the Scene socket by default.

JacquesLucke:
Thank you very much!

JacquesLucke:
I've made another minor typo fix: https://github.com/JacquesLucke/animation_nodes/commit/616e4b19115ea228076b97aedbeae6df5ed118dd

PR #586, opened 2016-08-14
Camera infrustrum

s-leger:
Returns point position in image uvz coords 
Use in conjonction with map ramp allow
eg set up material opacity from position of a point on screen.

 ```
        map float on a ramp in out range
        input:
        
          /
         /
        /
        
        output:
         _
        / \
        
```

### CODE REVIEW

**Comment by JacquesLucke** [comment]
Did you see the Map Range node?
map range
It looks a bit similiar. I didn't check the code in detail yet but it is possible to use this instead?

File: nodes/geometry/point_inside_camera_frustrum.py [hunk 1]
```diff
@@ -0,0 +1,42 @@
...
+
+class PointInCameraFrustrum(bpy.types.Node, AnimationNode):
+    bl_idname = "an_PointInCameraFrustrum"
+    bl_label = "Point in Camera Frustrum"
+    
+    def create(self):
+        self.newInput("Scene", "Scene", "scene", hide = True)
+        self.newInput("Object", "Camera", "camera")
+        self.newInput("Vector", "3d Point", "point", value=(0, 0, 0))
+        self.newInput("Float", "Threshold % (positive = interior)", "threshold", value = 0.0)
```
JacquesLucke: Hmm, the socket name should really just be the name. What exactly does this threshold do?

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,42 @@
...
+    def getUsedModules(self):
+        return ["bpy_extras"]
+    
+    def isInFrustrum(self, scene, camera, point, threshold):
+        if camera is None:
+            return 0, 0, 0, False, 0
+        if camera.type != 'CAMERA':
+            return 0, 0, 0, False, 0
+        co = bpy_extras.object_utils.world_to_camera_view(scene, camera, point)
+        cs, ce = camera.data.clip_start, camera.data.clip_end
```
JacquesLucke: I'd prefer ``clipStart`` and ``clipEnd`` as variable names here

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,42 @@
...
+        th = threshold / 100
```
JacquesLucke: Maybe the user of the node should divide the number beforehand. Do you expect the threshold to be in [0, 100] here?

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,42 @@
...
+        if th > 0.5:
+            th = 0.5
+        limit = 0.5 - th
+        u = co.x
+        v = co.y
+        z = co.z
+        visible = (0.0 + th < u < 1.0 - th and 0.0 + th < v < 1.0 - th and cs < z <  ce)
```
JacquesLucke: with the above changes this line get's longer, so I sugest to split it into three lines

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,42 @@
...
+    def getExecutionCode(self):
+        yield "u, v, z, visible = self.isInFrustrum(scene, camera, point, threshold)"
+            
+    def getUsedModules(self):
+        return ["bpy_extras"]
+    
+    def isInFrustrum(self, scene, camera, point, threshold):
+        if camera is None:
+            return 0, 0, 0, False, 0
+        if camera.type != 'CAMERA':
```
JacquesLucke: both if conditions can be combined into ``if getattr(camera, "type", "") != "CAMERA":``
This way we only need to write the default values once

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,42 @@
...
+            return 0, 0, 0, False, 0
+        co = bpy_extras.object_utils.world_to_camera_view(scene, camera, point)
+        cs, ce = camera.data.clip_start, camera.data.clip_end
+        th = threshold / 100
+        if th > 0.5:
+            th = 0.5
+        limit = 0.5 - th
+        u = co.x
+        v = co.y
+        z = co.z
```
JacquesLucke: ``u, v, z = co.xyz``

s-leger:
Well, 
The issue is clamp acting on both sides, so with unclamped ranges, the center part of the ramp is not clamped where it should.
Maybe a complex node setup may overcome this issue. 
I'ts all about more nodes vs complex setup.

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,42 @@
+import bpy
+from bpy.props import *
+from ... base_types.node import AnimationNode
+import bpy_extras
+
+class PointInCameraFrustrum(bpy.types.Node, AnimationNode):
+    bl_idname = "an_PointInCameraFrustrum"
```
JacquesLucke: Missing "Node" at the end of the name. (both class and bl_idname)

s-leger:
Threshold is a percent of image allow to force infrustrum value to return true when point is near border.
Positive are inside the frustrum, negative values are outside of the frustrum.
So the range should be in [-inf, 50]

JacquesLucke:
ok, I see. I'll need some more time to think about it. However we can finish the other node first. 
Please create a separate pull request for the Map Ramp In-Out node.

JacquesLucke:
Ah okay, I understand. Wouldn't it be better to just use (-inf, 0.5]?

s-leger:
Percent may be easyer to setup by user, don't you think ?
Ok to separate pull requests.

JacquesLucke:
We should also think about who will use this node. I don't think it will be used by many artists, more by technical persons. Hmm thinking about how to explain that... but I think we should go with (-inf, 0.5]... Something like -10% doesn't really makes sense. -0.1 is better here IMO... Also if the threshold is the result of another calculation it is easier
There are a few stylish issues with the code still as well. Will comment on them now. After that I'll merge.

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
+import bpy
+from bpy.props import *
+from ... base_types.node import AnimationNode
+import bpy_extras
```
JacquesLucke: move to line 2, just looks a bit nicer :)

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
...
+
+class PointInCameraFrustrumNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_PointInCameraFrustrumNode"
+    bl_label = "Point in Camera Frustrum"
+    
+    def create(self):
+        self.newInput("Scene", "Scene", "scene", hide = True)
+        self.newInput("Object", "Camera", "camera")
+        self.newInput("Vector", "Point", "point", value=(0, 0, 0))
```
JacquesLucke: spaces around the equal sign

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
...
+        self.newInput("Float", "Threshold", "threshold", value = 0.0)
+        
+        self.newOutput("Float", "Image u", "u")
+        self.newOutput("Float", "Image v", "v")
+        self.newOutput("Float", "Z depth", "z")
+        self.newOutput("Boolean", "Visible", "visible")
+        
+    def getExecutionCode(self):
```
JacquesLucke: is there any particular reason you use the ``getExecutionCode`` method here instead of just ``execute``?

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
...
+        yield "u, v, z, visible = self.isInFrustrum(scene, camera, point, threshold)"
+            
+    def getUsedModules(self):
```
JacquesLucke: not needed, because we don't use ``bpy_extras`` in the execution code.

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
...
+    
+    def isInFrustrum(self, scene, camera, point, threshold):
+        if getattr(camera, "type", "") != 'CAMERA':
+            return 0, 0, 0, False
+        co = bpy_extras.object_utils.world_to_camera_view(scene, camera, point)
+        clipStart, clipEnd = camera.data.clip_start, camera.data.clip_end
+        th = threshold / 100
+        if th > 0.5:
+            th = 0.5
+        limit = 0.5 - th
```
JacquesLucke: the limit variable is not used

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
...
+        return ["bpy_extras"]
+    
+    def isInFrustrum(self, scene, camera, point, threshold):
+        if getattr(camera, "type", "") != 'CAMERA':
+            return 0, 0, 0, False
+        co = bpy_extras.object_utils.world_to_camera_view(scene, camera, point)
+        clipStart, clipEnd = camera.data.clip_start, camera.data.clip_end
+        th = threshold / 100
+        if th > 0.5:
+            th = 0.5
```
JacquesLucke: join all three lines into: ``threshold = min(0.5, threshold)``

nodes/geometry/point_inside_camera_frustrum.py
```diff
@@ -0,0 +1,40 @@
...
+        limit = 0.5 - th
+        u, y, z = co.xyz
+        visible = ( 0.0 + th < u < 1.0 - th and 
+                    0.0 + th < v < 1.0 - th and
```
JacquesLucke: ``th`` variable is not needed. We just use ``threshold`` here.

JacquesLucke:
(just in case you have not seen it because you committed so fast, there were a few comments you did not implement yet)
I hope I don't scare you off with so many code comments -.-

s-leger:
Comments are always welcome, since i'm not so comfortable with python, and coding style should remain consistant.

JacquesLucke:
there are still comments in the code. If you want I can just merge it and correct it myself. Normally I would not do this.

JacquesLucke:
perfect, looks very clean now :)

PR #589, opened 2016-08-15
IsNone boolean check

s-leger:
Hi,
Is None to check if data path (or any value) is available.


JacquesLucke:
Maybe we can make this part of the  Compare node?

JacquesLucke:
I could update it to also allow only one input socket.

s-leger:
Great Idea, 
Less nodes, but more powerfull ones.

JacquesLucke:
#590

PR #591, opened 2016-08-15
Boolean compare A Is None

s-leger:
First attempt for A is None implementation.


JacquesLucke:
This looks quite good already.
Two notes:

1. In the ``getExecutionCode`` function all return statements (except the last one) should be aligned.
2. The ``getWantedDataType`` function (I know the name is not perfect....) can be simplified with a loop that iterates over all input sockets.

Issue #588, opened 2016-08-14
Baking attribute output


s-leger:
Hi, 
Trying to bake object's material opacity with an object attribute output, but with no luck until now.
Any idea, advice on how to do such things, or why this can't be done and what should we do to get it working ?

JacquesLucke:
can you make a screenshot of the property you want to bake please?

s-leger:
![Image bake material](https://raw.githubusercontent.com/s-leger/images/master/material.jpg)

s-leger:
Found something:
C.object.keyframe_insert("data.materials[\"Material\"].alpha", index = -1)
ValueError: bpy_struct.keyframe_insert() path spans ID blocks

bpy.data.materials["Material"].keyframe_insert(data_path="alpha", index = -1)
True

C.object.data.materials[0].keyframe_insert(data_path="alpha", index=-1)
True

So code generation must insert keyframe on object path less property part


    def getBakeCode(self):

        yield "if object is not None:"
        yield "    try: object."++".keyframe_insert(path, index = arrayIndex)"
        yield "    except: pass"

s-leger:
Try to get object path for baking this way seem to work in ObjectDataPathOutputNode

    def getPropertyPath(self, object, path):
        if "." in path:
            path_prop, path_attr = path.rsplit(".", 1)
            prop = object.path_resolve(path_prop)
        else:
            prop = object
            path_attr = path
        return prop, path_attr
        
    def getBakeCode(self):
        yield "if object is not None:"
        yield "    pathToProperty, propertyName = self.getPropertyPath(object, path)"
        yield "    try: pathToProperty.keyframe_insert(propertyName, arrayIndex)"
        yield "    except Exception as e:"
        yield "        print(\"Bake error : %s %s %s\" % (object.name, (\"{}.keyframe_insert({},{})\".format(pathToProperty, repr(propertyName), arrayIndex)), e))"
        yield "        pass"

s-leger:
The same for ObjectAttributeOutputNode, but dont know how to handle indexed attributes in array

    def getBakeCode(self):
        if isCodeValid(self.attribute):
            yield "if object is not None:"
            yield "    pathToProperty, propertyName = self.getPropertyPath(object, {})".format(repr(self.attribute))
            yield "    try: pathToProperty.keyframe_insert(propertyName, -1)"
        #   yield "    try: object.keyframe_insert({})".format(repr(self.attribute))
            yield "    except: pass"

PR #593, opened 2016-08-18
attempt to fix baking issue

s-leger:
Fix for issue #588
keyframe_insert may return ValueError: bpy_struct.keyframe_insert() path spans ID blocks

### DIFF
--- a/nodes/object/object_data_path_output.py
+++ b/nodes/object/object_data_path_output.py
@@ -35,19 +35,19 @@ class ObjectDataPathOutputNode(bpy.types.Node, AnimationNode):
     
     def getPropertyPath(self, object, path):
         if "." in path:
-            propPathStr, propName = path.rsplit(".", 1)
-            propPath = object.path_resolve(propPathStr)
+            propPath, propName = path.rsplit(".", 1)
+            dataPath = object.path_resolve(propPath)
         else:
-            propPath = object
+            dataPath = object
             propName = path
-        return propPath, propName
+        return dataPath, propName
     
     def getBakeCode(self):
         yield "if object is not None:"
         yield "    try: object.keyframe_insert(path, index = arrayIndex)"
         yield "    except:"
-        yield "        propPath, propName = self.getPropertyPath(object, path)"
-        yield "        try: propPath.keyframe_insert(propName, index = arrayIndex)"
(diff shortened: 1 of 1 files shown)


nodes/object/object_data_path_output.py
```diff
@@ -32,12 +32,24 @@ def execute(self, object, path, arrayIndex, value):
...
             self.errorMessage = "Error"
         return object
-
+    
+    def getPropertyPath(self, object, path):
+        if "." in path:
+            propPathStr, propName = path.rsplit(".", 1)
+            propPath = object.path_resolve(propPathStr)
+        else:
+            propPath = object
```
JacquesLucke: the name "propPath" is very misleading as it contains an object

s-leger:
Well, maybe dataPath could be better, as with material the path resolve in  bpy.data.materials ?

nodes/object/object_data_path_output.py
```diff
@@ -32,12 +32,24 @@ def execute(self, object, path, arrayIndex, value):
         except:
             self.errorMessage = "Error"
         return object
-
+    
+    def getPropertyPath(self, object, path):
+        if "." in path:
+            propPath, propName = path.rsplit(".", 1)
+            dataPath = object.path_resolve(propPath)
```
JacquesLucke: can this throw an exception if the path is not valid?

nodes/object/object_data_path_output.py
```diff
@@ -36,7 +36,10 @@ def execute(self, object, path, arrayIndex, value):
     def getPropertyPath(self, object, path):
         if "." in path:
             propPath, propName = path.rsplit(".", 1)
-            dataPath = object.path_resolve(propPath)
+            try:
```
s-leger: Prevent ValueError from rising if path is not valid

JacquesLucke:
Looks good. Can you maybe give 2-3 examples of data paths that work now but did not work before?

s-leger:
At least anything related to standard materials and textures, like data.materials[0].alpha and so on.
Not found any other one until now.

JacquesLucke:
ok, so at least this works now :)
All things that worked before shouldn't be broken by this change. good job :)

PR #621, opened 2016-10-16
Add option to not copy addon files to Blender's addon directory

jasperges:
For me personally it's a bit annoying that the files will be copied over when building. So I thought it might be nice to add on option to disable this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,8 @@ def main():
         compileCythonFiles()
         if "-export" in initialArgs:
             export()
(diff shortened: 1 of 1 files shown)


JacquesLucke:
ok I see! will implement your change when I'm back at the pc :)

JacquesLucke:
I've implemented it a little bit differently. Also I'll implement that the setup.py file automatically creates the config.py file now.
thanks for your help.

PR #642, opened 2016-12-02
Particle Animation Nodes

lvxejay:
Added More useful Particle Nodes and the ability to grab particle settings and transfer them across particle systems. Still not animatable due to Blender limitations


acil1996:
Hey. Where can I download your assembly ???

PR #645, opened 2016-12-08
Patch 1

ricardoroehe:
A fix to keep the style (bold, italic, underline and small caps) the same from the input text.
Seems to work with **some** special characters (space, tabulation and line break).


JacquesLucke:
Thanks I will check it in a few days when I have a bit more time!

JacquesLucke:
I fixed it here: 7771316f2c9559808858eba931a5944cf1fe8fb2
Thanks for figuring out what the problem was!

PR #724, opened 2017-07-06
Avoid misunderstanding

LeonCheung:
Since currently not all node types are listed, we should avoid users thinking the other way. Or better idea?


JacquesLucke:
Hmm I see the problem, however "Key Node" is not really meaningful either.

Not sure if a fix is really needed here..

PR #766, opened 2017-08-17
Cythonized And Vectorized Geometry Intersection And Projection Nodes

OmarSquircleArt:


animation_nodes/nodes/geometry/c_utils.pyx
```diff
@@ -0,0 +1,269 @@
...
+        else:
+            valid.data[i] = False
+            firstParameter.data[i] = 0
+            secondParameter.data[i] = 0
+            nearestPoint1.data[i] = Vector3(0,0,0)
+            nearestPoint2.data[i] = Vector3(0,0,0)
+    return nearestPoint1, nearestPoint2, firstParameter, secondParameter, valid
+
+def IntersectLineSphere(Py_ssize_t amount,
+                        VirtualVector3DList LineStart,
```
JacquesLucke: please don't start variable names with a capital letter. I see this in multiple places here, should be corrected.

animation_nodes/nodes/geometry/intersect_line_line.py
```diff
@@ -1,36 +1,75 @@
...
+        self.newInput(VectorizedSocket("Vector", "useSecondLineStartList",
+            ("Second Line Start", "secondLineStart", dict(value = (1, -1, 0))),
+            ("Second Lines Start", "secondLinesStart"),
+            codeProperties = dict(default = (1, -1, 0))))
+        self.newInput(VectorizedSocket("Vector", "useSecondLineEndList",
+            ("Second Line End", "secondLineEnd", dict(value = (-1, 1, 0))),
+            ("Second Lines End", "secondLinesEnd"),
+            codeProperties = dict(default = (-1, 1, 0))))
+
+        self.newOutput(VectorizedSocket("Vector", ["useFirstLineStartList", "useFirstLineEndList", "useSecondLineStartList", "useSecondLineEndList"],
```
JacquesLucke: would be better to define something like `props = ["useFirstLineStartList", ...]` first and then just pass this to every socket to reduce duplication and line length.

animation_nodes/nodes/geometry/c_utils.pyx
OmarSquircleArt: Fixed This in https://github.com/JacquesLucke/animation_nodes/pull/766/commits/c2b8bd60db89cd826b57d41c3065ec9d5bb1231f

animation_nodes/nodes/geometry/c_utils.pyx
```diff
@@ -0,0 +1,494 @@
...
+    cdef DoubleList secondParameterList = DoubleList(length = amount)
+    cdef BooleanList validList = BooleanList(length = amount)
+    cdef Vector3 firstNearestPoint, secondNearestPoint
+    cdef double firstParameter, secondParameter
+    cdef bint valid
+    cdef Py_ssize_t i
+    for i in range(amount):
+        intersectLineLine(firstLineStartList.get(i), firstLineEndList.get(i),
+                        secondLineStartList.get(i), secondLineEndList.get(i),
+                        &firstNearestPoint, &secondNearestPoint,
```
JacquesLucke: Instead of `&firstNearestPoint` you could just pass `firstNearestPointList.data + i`.
Same for the other cases parameters.

JacquesLucke:
for better function structuring please use those names:
`intersect_LineLine`
`intersect_LineLine_Single`
`intersect_LineLine_List`
in that order, for all intersection cases.

I know that this is not the normal style convention but in my opinion it groups the functions that belong together in a better way.

For each group of 3 functions create a comment at the top:
```
# Line-Line Intersection
##########################################
```

animation_nodes/nodes/geometry/c_utils.pyx
```diff
@@ -0,0 +1,494 @@
...
+
+cdef intersectPlanePlane(Vector3 *firstPlanePoint, Vector3 *firstPlaneNormal,
+                        Vector3 *secondPlanePoint, Vector3 *secondPlaneNormal,
+                        Vector3 *outLineDirection, Vector3 *outLinePoint, bint *outValid):
+    cdef Vector3 direction, unitNormal1, unitNormal2, factor1, factor2, point
+    cdef double h1, h2, dot, c1, c2, invDotSquared
+    normalizeVec3(&unitNormal1, firstPlaneNormal)
+    normalizeVec3(&unitNormal2, secondPlaneNormal)
+    crossVec3(&direction, firstPlaneNormal, secondPlaneNormal)
+    if lengthVec3(&direction) > 1e-6:
```
JacquesLucke: please invert the branch cases.
```
if lengthVec3(&direction) < 1e-6:
    ...
    return

...
```

This is called a guard condition: https://refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html
It helps to keep the main execution code at first level of indentation, making it easier to read.

JacquesLucke: Same in all other low level functions as well.

animation_nodes/nodes/geometry/c_utils.pyx
JacquesLucke: This also eliminates many variables in other functions which only store the values just to assign them to the list.

animation_nodes/nodes/geometry/intersect_line_line.py
```diff
@@ -1,36 +1,81 @@
...
+            ("Second Line Start", "secondLineStart", dict(value = (1, -1, 0))),
+            ("Second Line Starts", "secondLineStarts"),
+            codeProperties = dict(default = (1, -1, 0))))
+        self.newInput(VectorizedSocket("Vector", "useSecondLineEndList",
+            ("Second Line End", "secondLineEnd", dict(value = (-1, 1, 0))),
+            ("Second Line Ends", "secondLineEnds"),
+            codeProperties = dict(default = (-1, 1, 0))))
+
+        props = ["useFirstLineStartList", "useFirstLineEndList",
+        "useSecondLineStartList", "useSecondLineEndList"]
```
JacquesLucke: indent this line like so:
```
props = ["...
         "...
```

animation_nodes/nodes/geometry/intersect_line_line.py
```diff
@@ -1,36 +1,81 @@
...
+    def getExecutionFunctionName(self):
+        useList = any((self.useFirstLineStartList, self.useFirstLineEndList,
+        self.useSecondLineStartList, self.useSecondLineEndList))
+        if useList:
+            return "execute_List"
+        else:
+            return "execute_Single"
+
+    def execute_List(self, firstLineStarts, firstLineEnds, secondLineStarts, secondLineEnds):
+        firstLineStarts = VirtualVector3DList.fromListOrElement(firstLineStarts,
```
JacquesLucke: Don't break the line here.
It is likely that I will rename the `fromListOrElement` function soon to give it a shorter name.
Also, almost always when you do a line break your should indent the lower line to make it easier to see that it belongs to the line before.

PR #776, opened 2017-08-24
Geometry Unit Tests And Minor Fixes.

OmarSquircleArt:


animation_nodes/nodes/geometry/c_utils.pyx
```diff
@@ -20,7 +21,7 @@ cdef intersect_LineLine(Vector3 *firstLineStart, Vector3 *firstLineEnd,
     subVec3(&direction1, firstLineEnd, firstLineStart)
     subVec3(&direction2, secondLineEnd, secondLineStart)
     crossVec3(&normal, &direction1, &direction2)
-    if lengthVec3(&normal) < 1e-6:
+    if lengthVec3(&normal) < 1e-6 or lengthVec3(&direction1) == 0 or lengthVec3(&direction2) == 0:
```
JacquesLucke: is this extra check really necessary? The length of the normal is 0 when one of the directions is 0, right?
Also, if you just want to compare to 0 it would be faster to use the lengthsquared function.
Another good idea would be to create another `isZeroVec3` in the vector math file that just checks the individual components.

animation_nodes/nodes/geometry/test_intersection.py
```diff
@@ -0,0 +1,107 @@
+from mathutils import Vector
+from unittest import TestCase
+from . c_utils import (
+                       intersect_LineLine_Single,
```
JacquesLucke: please indent these lines only once (I know I don't do this myself everytime, I'm not sure yet what the best way is, to import many things)

animation_nodes/nodes/geometry/c_utils.pyx
OmarSquircleArt: Right, will remove them.

PR #922, opened 2018-07-11
New Algorithm For The Splines From Edges Node.

OmarSquircleArt:
This new algorithm create as much continuous splines from edges as possible, resulting in more workable output for processing and rendering.

### DIFF
--- a/animation_nodes/nodes/spline/splines_from_edges.py
+++ b/animation_nodes/nodes/spline/splines_from_edges.py
@@ -47,7 +47,8 @@ class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
             layout.prop(self, "radiusType", text = "")
 
     def execute(self, vertices, edgeIndices, radii):
-        if len(edgeIndices) == 0 or edgeIndices.getMaxIndex() >= len(vertices):
+        if len(edgeIndices) == 0: return []
+        if edgeIndices.getMaxIndex() >= len(vertices):
             self.raiseErrorMessage("Invalid Edge Indices.")
 
         radii = VirtualDoubleList.create(radii, 0.1)
--- a/animation_nodes/nodes/spline/splines_from_edges.py
+++ b/animation_nodes/nodes/spline/splines_from_edges.py
@@ -17,6 +17,7 @@ radiusTypeItems = [
 class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_SplinesFromEdgesNode"
     bl_label = "Splines from Edges"
+    errorHandlingType = "EXCEPTION"
     bl_width_default = 160
 
     algorithmType = EnumProperty(name = "Algorithm Type", default = "EDGE",
@@ -46,6 +47,9 @@ class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
             layout.prop(self, "radiusType", text = "")
 
     def execute(self, vertices, edgeIndices, radii):
+        if len(edgeIndices) == 0 or edgeIndices.getMaxIndex() >= len(vertices):
+            self.raiseErrorMessage("Invalid Edge Indices.")
+
         radii = VirtualDoubleList.create(radii, 0.1)
         if self.algorithmType == "EDGE":
             return splinesFromEdges(vertices, edgeIndices, radii, self.radiusType)
--- a/animation_nodes/nodes/spline/splines_from_edges.py
+++ b/animation_nodes/nodes/spline/splines_from_edges.py
@@ -50,4 +50,4 @@ class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
         if self.algorithmType == "EDGE":
             return splinesFromEdges(vertices, edgeIndices, radii, self.radiusType == "VERTEX")
         else:
-            return splinesFromBranches(vertices, edgeIndices)
+            return splinesFromBranches(vertices, edgeIndices, radii)
--- a/animation_nodes/nodes/spline/splines_from_edges.py
+++ b/animation_nodes/nodes/spline/splines_from_edges.py
@@ -5,7 +5,7 @@ from ... data_structures import VirtualDoubleList
 from ... base_types import AnimationNode, VectorizedSocket
 from . splines_from_edges_utils import splinesFromBranches, splinesFromEdges
 
-algorithmItems = [
+algorithmTypeItems = [
     ("EDGE", "Spline Per Edge", "", "NONE", 0),
     ("BRANCH", "Spline Per Branch", "", "NONE", 1)
 ]
@@ -21,7 +21,7 @@ class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
 
     algorithmType = EnumProperty(name = "Algorithm Type", default = "EDGE",
         description = "Choose the algorithm used for generating the splines.",
-        update = propertyChanged, items = algorithmItems)
+        update = propertyChanged, items = algorithmTypeItems)
 
     radiusType = EnumProperty(name = "Radius Type", default = "EDGE",
         description = "Only important if there is a list of radii.",
@@ -48,6 +48,6 @@ class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
     def execute(self, vertices, edgeIndices, radii):
         radii = VirtualDoubleList.create(radii, 0.1)
         if self.algorithmType == "EDGE":
-            return splinesFromEdges(vertices, edgeIndices, radii, self.radiusType == "VERTEX")
+            return splinesFromEdges(vertices, edgeIndices, radii, self.radiusType)
         else:
             return splinesFromBranches(vertices, edgeIndices, radii)
--- a/animation_nodes/nodes/spline/splines_from_edges.py
+++ b/animation_nodes/nodes/spline/splines_from_edges.py
@@ -1,10 +1,14 @@
 import bpy
 from bpy.props import *
-from . c_utils import splinesFromEdges
 from ... events import propertyChanged
-from ... data_structures import DoubleList
+from ... data_structures import VirtualDoubleList
 from ... base_types import AnimationNode, VectorizedSocket
+from . splines_from_edges_utils import splinesFromBranches, splinesFromEdges
 
+algorithmItems = [
+    ("EDGE", "Spline Per Edge", "", "NONE", 0),
+    ("BRANCH", "Spline Per Branch", "", "NONE", 1)
+]
 radiusTypeItems = [
     ("EDGE", "Radius per Edge", "", "NONE", 0),
     ("VERTEX", "Radius per Vertex", "", "NONE", 1)
@@ -14,6 +18,10 @@ class SplinesFromEdgesNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_SplinesFromEdgesNode"
     bl_label = "Splines from Edges"
     bl_width_default = 160
+
+    algorithmType = EnumProperty(name = "Algorithm Type", default = "EDGE",
+        description = "Choose the algorithm used for generating the splines.",
+        update = propertyChanged, items = algorithmItems)
 
     radiusType = EnumProperty(name = "Radius Type", default = "EDGE",
         description = "Only important if there is a list of radii.",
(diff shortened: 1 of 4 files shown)


animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,117 @@
...
+    PyMem_Free(nonBipolarVertices)
+    PyMem_Free(neighboursAmounts)
+    PyMem_Free(neighboursStarts)
+    PyMem_Free(filledSpaces)
+    PyMem_Free(neighbours)
+
+    return splines
+
+def splinesFromEdges(Vector3DList vertices, EdgeIndicesList edges, VirtualDoubleList radii,
+                     bint isVertexRadius):
```
JacquesLucke: would prefer a `str` input here for `isVertexRadius` because "how are the radii used?" is not really a yes-no question.
There could be more than two options in theory.

animation_nodes/nodes/spline/splines_from_edges.py
```diff
@@ -1,10 +1,14 @@
 import bpy
 from bpy.props import *
-from . c_utils import splinesFromEdges
 from ... events import propertyChanged
-from ... data_structures import DoubleList
+from ... data_structures import VirtualDoubleList
 from ... base_types import AnimationNode, VectorizedSocket
+from . splines_from_edges_utils import splinesFromBranches, splinesFromEdges
 
+algorithmItems = [
```
JacquesLucke: just by convention either this should be called `algorithmTypeItems` or the property in the node should just be called `algorithm`.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,117 @@
...
+def splinesFromBranches(Vector3DList vertices, EdgeIndicesList edges, VirtualDoubleList radii):
+    cdef int i, j
+    cdef int edgesAmount = len(edges)
+    cdef int verticesAmount = len(vertices)
+
+    # Compute how many neighbour each vertex have.
+    cdef int *neighboursAmounts = <int*>PyMem_Malloc(verticesAmount * sizeof(int))
+    memset(neighboursAmounts, 0, verticesAmount * sizeof(int))
+    for i in range(edgesAmount):
+        neighboursAmounts[edges.data[i].v1] += 1
```
JacquesLucke: this could raise an segmentation fault if there aren't enough vertices.
The node itself should check wether the inpu is valid (max vertex index of an edge is smaller than the number of vertices, like in the original code)

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,117 @@
...
+    VirtualDoubleList, FloatList
+)
+
+def splinesFromBranches(Vector3DList vertices, EdgeIndicesList edges, VirtualDoubleList radii):
+    cdef int i, j
+    cdef int edgesAmount = len(edges)
+    cdef int verticesAmount = len(vertices)
+
+    # Compute how many neighbour each vertex have.
+    cdef int *neighboursAmounts = <int*>PyMem_Malloc(verticesAmount * sizeof(int))
```
JacquesLucke: in general it would be nice to use the Lists AN provides instead of doing memory allocations yourself if appropriate. The performance impact should be negilible.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,117 @@
...
+    for i in range(nonBipolarVertsCount):
+        nonBipolarVertex = nonBipolarVertices[i]
+        for j in range(neighboursAmounts[nonBipolarVertex]):
+            nextVertex = neighbours[neighboursStarts[nonBipolarVertex] + j]
+            if (filledSpaces[nextVertex] == neighboursAmounts[nextVertex] or
+                                           neighboursAmounts[nextVertex] != 2):
+                splineVertices = Vector3DList.__new__(Vector3DList, capacity = verticesAmount)
+                splineRadii = FloatList.__new__(FloatList, capacity = verticesAmount)
+
+                splineVertices.append(vertices[nonBipolarVertex])
```
JacquesLucke: using the `append_LowLevel` function here might result in a noticable performance improvement

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,117 @@
...
+                splineRadii.append(radii.get(nonBipolarVertex))
+                splineVertices.append(vertices[nextVertex])
+                splineRadii.append(radii.get(nextVertex))
+                filledSpaces[nonBipolarVertex] -= 1
+                filledSpaces[nextVertex] -= 1
+
+                currentVertex = nonBipolarVertex
+                for _ in range(verticesAmount):
```
JacquesLucke: maybe a better `range(...)` can be found so that this does not feel like a very inefficient inner loop?
Also by just looking at the way you use the loop it might be more appropriate to use a while loop instead?

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
OmarSquircleArt: Should the node itself have error handling then?

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
OmarSquircleArt: I am wondering if it would be better to do:
```python
splineRadii = FloatList.__new__(FloatList, length = verticesAmount, capacity = verticesAmount)
splineRadii[0] = radii.get(nonBipolarVertex)
splineRadii[1] = radii.get(nextVertex)
count = 2
for _ in range():
    ...
    splineRadii[count] = radii.get(nextVertex)
    count += 1
```

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
JacquesLucke: just ment a better upper bound could be used in `range(...)`.

Anyway, looks like a for loop shouldn't be used here at all. Can you try to convert it into a while loop?

OmarSquircleArt: Sure, that would be my first while loop :smile:.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
JacquesLucke: yes, I think the node should do the check. "cutils" usually should not have to make expensive checks, they can expect the data to valid.
Nodes on the other hand can not expect that.

OmarSquircleArt: Alright.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
JacquesLucke: that would be even faster, but not sure if it is worth the extra variable

btw, when you specify the length of a list in the constructor, you don't need to specify the capacity (if it is the same)

OmarSquircleArt: We could reuse the variable `j`, then we won't have to create any extra variables. And it will be used in a narrow vertical scope. Should we do that?

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,114 @@
...
+                splineVertices.append(vertices[nonBipolarVertex])
+                splineRadii.append(radii.get(nonBipolarVertex))
+                splineVertices.append(vertices[nextVertex])
+                splineRadii.append(radii.get(nextVertex))
+                filledSpaces[nonBipolarVertex] -= 1
+                filledSpaces[nextVertex] -= 1
+
+                currentVertex = nonBipolarVertex
+                while neighboursAmounts[nextVertex] == 2:
+                    i = currentVertex
```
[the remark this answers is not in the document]
JacquesLucke: the thing is this: it is not used now but you never know if it will be used in the future. By reusing it you sign up for trouble.

Also if anyone wants to convert the code to c, you can't simply reuse the variable, because you would probably end up in an endless loop or so.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
OmarSquircleArt: I meant to use it instead of `count`, I was responding to your comment:
> but not sure if it is worth the extra variable.
But yah, I will no longer reuse variables, will have to edit some stuff in account with this. :D

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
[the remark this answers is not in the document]
OmarSquircleArt: I see, will make sure not to do that again.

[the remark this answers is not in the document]
OmarSquircleArt: How about this form?
```python
while neighboursAmounts[nextVertex] == 2:
    if neighbours[neighboursStarts[nextVertex]] == currentVertex:
        currentVertex = nextVertex
        nextVertex = neighbours[neighboursStarts[nextVertex] + 1]
    else:
        currentVertex = nextVertex
        nextVertex = neighbours[neighboursStarts[nextVertex]]
```
`currentVertex = nextVertex` will be written twice, but we will avoid creating any temporary variables or extra lines of code.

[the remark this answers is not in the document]
OmarSquircleArt: I think such inline expression will also work. Which one should we use?
```python
currentVertex, nextVertex = nextVertex, neighbours[neighboursStarts[nextVertex] + neighbours[neighboursStarts[nextVertex]] == currentVertex]
```

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,120 @@
...
+
+    # Compute the start index of each group of neighbours of each vertex.
+    cdef int *neighboursStarts = <int*>PyMem_Malloc(verticesAmount * sizeof(int))
+    cdef int start = 0
+    for i in range(verticesAmount):
+        neighboursStarts[i] = start
+        start += neighboursAmounts[i]
+
+    # Keep track of how many index is in each group of neighbours at each iteration.
+    cdef int *filledSpaces = <int*>PyMem_Malloc(verticesAmount * sizeof(int))
```
JacquesLucke: I still think you should use an `IntegerList` instead of allocating it yourself.
`IntegerList.fromValue(0, length = verticesAmount)`

OmarSquircleArt: OK, so, in general, we shouldn't use malloc if there is a list alternative.

JacquesLucke: Right, by using lists the code is safer, we can't forget to free and it is cleaner.

OmarSquircleArt: Everything now uses lists.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,111 @@
...
+
+def splinesFromBranches(Vector3DList vertices, EdgeIndicesList edges, VirtualDoubleList radii):
+    cdef int i, j
+    cdef int edgesAmount = edges.length
+    cdef int verticesAmount = vertices.length
+
+    # Compute how many neighbour each vertex have.
+    cdef IntegerList neighboursAmounts = IntegerList.fromValue(0, length = verticesAmount)
+    for i in range(edgesAmount):
+        neighboursAmounts[edges.data[i].v1] += 1
```
JacquesLucke: for performance reasons you should use `.data` when accessing list elements from cython code.

OmarSquircleArt: Was just going to ask you about this. Will do.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,113 @@
...
+        neighboursStarts.data[i] = start
+        start += neighboursAmounts.data[i]
+
+    # Keep track of how many index is in each group of neighbours at each iteration.
+    cdef IntegerList filledSpaces = IntegerList.fromValue(0, length = verticesAmount)
+
+    # Compute the indices of neighbouring vertices of each vertex.
+    cdef IntegerList neighbours = IntegerList(length = edgesAmount * 2)
+    for i in range(edgesAmount):
+        neighbours.data[neighboursStarts.data[edges.data[i].v1] +
```
JacquesLucke: I'd suggest extra `v1` and `v2` variables of type `unsigned int` to reduce line length and improve readability.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,113 @@
...
+    cdef FloatList splineRadii
+    cdef Vector3DList splineVertices
+    cdef int nonBipolarVertex, currentVertex, nextVertex
+    for i in range(nonBipolarVertsCount):
+        nonBipolarVertex = nonBipolarVertices.data[i]
+        for j in range(neighboursAmounts.data[nonBipolarVertex]):
+            nextVertex = neighbours.data[neighboursStarts.data[nonBipolarVertex] + j]
+            if (filledSpaces.data[nextVertex] == neighboursAmounts.data[nextVertex] or
+                                                neighboursAmounts.data[nextVertex] != 2):
+                splineVertices = Vector3DList.__new__(Vector3DList, length = verticesAmount)
```
JacquesLucke: these lists are wayy to large in many cases. Better use the `append_LowLevel` function instead of allocating huge arrays for every spline.

OmarSquircleArt: What should the initial capacity of those lists be?

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
JacquesLucke: I think in this case it is fine. You could just check if cython generates efficient code for it by checking the generated c code.
(You can create an annotated html version of the file with this command: `cython -a path/to/file.pyx`)

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
JacquesLucke: Just don't give them an initial capacity, or maybe `2`.

animation_nodes/nodes/spline/splines_from_edges_utils.pyx
```diff
@@ -0,0 +1,117 @@
...
+    for i in range(edgesAmount):
+        v1, v2 = edges.data[i].v1, edges.data[i].v2
+        neighbours.data[neighboursStarts.data[v1] + usedSlots.data[v1]] = v2
+        neighbours.data[neighboursStarts.data[v2] + usedSlots.data[v2]] = v1
+        usedSlots.data[v1] += 1
+        usedSlots.data[v2] += 1
+
+    # Generate Splines.
+    cdef list splines = []
+    cdef PolySpline spline
```
JacquesLucke: is this used?

OmarSquircleArt: No, my bad.

PR #935, opened 2018-08-17
Added Material Socket and Set Object Materials Node.

OmarSquircleArt:
This patch add support for material as an explicit type. The *Set Object Materials Node* set one or multiple materials to a certain object.


JacquesLucke:
Looks good to me. I think this can be ported to Blender 2.8 and then merged.

PR #937, opened 2018-08-18
Vector output, center along options and order change for grid option in distribute node.

OmarSquircleArt:
- Added vector list output for distribute matrices node.
- Added center along axis options for distribute matrices node grid option.
- Changed the order of points/matrices for the grid option in the distribute matrices node such that it grows on x, then y, then z (Conventional cartesian indexing).
- Fixed bug in Compose Matrix node where if all options were disabled and the output was connected to a matrix list, a fatal error occur.

### DIFF
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -42,6 +42,13 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
     meshMode = EnumProperty(name = "Mesh Mode", default = "VERTICES",
         items = meshModeItems, update = AnimationNode.refresh)
 
+    centerAlongX = BoolProperty(name = "Center Along X", default = True,
+        description = "Center the grid along the x axis.", update = AnimationNode.refresh)
+    centerAlongY = BoolProperty(name = "Center Along Y", default = True,
+        description = "Center the grid along the y axis.", update = AnimationNode.refresh)
+    centerAlongZ = BoolProperty(name = "Center Along Z", default = False,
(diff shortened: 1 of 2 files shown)


animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -43,6 +42,13 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
     meshMode = EnumProperty(name = "Mesh Mode", default = "VERTICES",
         items = meshModeItems, update = AnimationNode.refresh)
 
+    centerAlongX = BoolProperty(name = "Center Along X", default = True,
+        description = "Center the grid along the x axis.", update = AnimationNode.refresh)
```
JacquesLucke: Instead of `AnimationNodes.refresh` the `executionCodeChanged` event should be used. 
The current one will trigger a rebuild of the sockets of the node which does seem to be necessary.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -88,25 +95,40 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
     def drawAdvanced(self, layout):
         if self.mode == "CIRCLE":
             layout.prop(self, "exactCircleSegment")
 
-    def getExecutionFunctionName(self):
+    def getExecutionCode(self, required):
         if self.mode == "LINEAR":
-            return "execute_Linear"
+            if self.distanceMode == "STEP":
+                list = ["matrices = self.execute_Linear(amount, distance)"]
```
JacquesLucke: Don't use `list` as a variable name.
Also `yield` can be used to instead of explicitely creating a list.

PR #983, opened 2018-12-24
Fix preference access

amirpavlo:
Latest blender chagned bpy.context.user_preferences
to
bpy.context.preferences


OmarSquircleArt:
This was already fixed. Moreover, development happen in the blender2.8 branch, not in master.

PR #990, opened 2018-12-31
Fixed Separate Text Object

iroot:
Fixed selection and unlinking of objects for correct operation of the “Separate text object” node.

### DIFF
--- a/animation_nodes/nodes/text/separate_text_object.py
+++ b/animation_nodes/nodes/text/separate_text_object.py
@@ -29,7 +29,6 @@ class SeparateTextObjectNode(bpy.types.Node, AnimationNode):
     sourceObjectName: StringProperty(name = "Source Object")
     currentID: IntProperty(default = 0)
     objectCount: IntProperty(default = 0)
-    addToMainContainer: BoolProperty(name = "Add To Main Container", default = True)
     materialName: StringProperty(name = "Material", default = "")
 
     outputType: EnumProperty(name = "Output Type", default = "MESH",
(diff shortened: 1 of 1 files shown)


animation_nodes/nodes/text/separate_text_object.py
```diff
@@ -235,6 +234,7 @@ def removeObject(object):
 def addObjectsToMainContainer(objects):
     mainContainer = getMainObjectContainer(bpy.context.scene)
     for object in objects:
+        bpy.context.collection.objects.unlink(object)
```
JacquesLucke: This line seems wrong. Why should objects be unlinked from some collection in the `addObjectsToMainContainer` function?
Either the name should change or unlinking should be done in a separate function.
Also this line has the hidden assumption that the objects are within `bpy.context.collection`.

iroot: Good point, I will put it into a separate function and will call it higher.

iroot:
> Did you try the node? I don't think this is sufficient to fix the node. The main problem with this node is the `object.convert` function, its poll fails even in the correct context.

Yes, I tried this node.

It was necessary to select objects in `onlySelectList` via the `object.select_set(True)` method and set `object.hide_select = True` only after all the execution of all operations requiring the correct context.

OmarSquircleArt:
Ok I just tested it and it seems to be working. Shouldn't we just link to the appropriate collection in the `newCharacterObject` function. So it should be something like:

```python
if self.addToMainContainer:
    mainContainer = getMainObjectContainer(bpy.context.scene)
    mainContainer.objects.link(characterObject)
else:
    bpy.context.collection.objects.link(characterObject)
```
What do you think?

iroot:
Without strong refactoring, there are two options.
1. Pass the `self.addToMainContainer` field to `newCharacterObject`
2. Or a collection with which to link objects.

I think that the second option is better.

JacquesLucke:
What would you do if you would refactor the node completely?

I think the `addToMainContainer` option could be removed alltogether now that there are collections. Just always add the objects to the "main container collections". Users can still move it somewhere else if they want.
The main container mostly existed to keep the outliner a bit cleaner.

iroot:
By strong refactoring, I meant moving methods to a class. To reduce the number of arguments. But now it is no longer necessary.

Removed advanced settings, including the relationship lines (this option didn't work and apparently lost its meaning).

PR #996, opened 2019-01-09
A new sound system.

OmarSquircleArt:
The new system uses a just-in-time computing approach allowing real-time and interactive visualizations, effectively removing the need for sound baking. The code adds:

- Sound From Sequence node.
- Sound Spectrum node.
- Sound Intensity node.
- Sound, SoundSequence, and SoundData  classes.


animation_nodes/data_structures/sounds/sound_sequence.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, data, start, end, volume, fps):
+        self.data = data
+        self.start = start
+        self.end = end
+        self.volume = volume
+        self.fps = fps
+
+    @classmethod
+    def fromSequence(cls, sequence):
+        for scene in bpy.data.scenes:
```
JacquesLucke:
```suggestion
        scene = findSceneWithSequence(sequence)
```

animation_nodes/nodes/sound/sound_from_sequence.py
```diff
@@ -0,0 +1,19 @@
+import bpy
+from ... data_structures import Sound, SoundSequence
+from ... base_types import AnimationNode, VectorizedSocket
+
+class SoundFromSequenceNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_SoundFromSequenceNode"
+    bl_label = "Sound From Sequence"
+
+    multipleSequences: VectorizedSocket.newProperty()
```
JacquesLucke:
```suggestion
    useSequenceList: VectorizedSocket.newProperty()
```

animation_nodes/nodes/sound/sound_intensity.py
```diff
@@ -0,0 +1,61 @@
...
+class SoundIntensityNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_SoundIntensityNode"
+    bl_label = "Sound Intensity"
+    errorHandlingType = "EXCEPTION"
+
+    smoothingSamples: IntProperty(name = "Smoothing Samples", default = 5, min = 0)
+    reductionFunction: EnumProperty(name = "Reduction Function", default = "MAX",
+        items = reductionItems)
+
+    useFloatList: VectorizedSocket.newProperty()
```
JacquesLucke:
```suggestion
    useFrameList: VectorizedSocket.newProperty()
```

animation_nodes/nodes/sound/sound_intensity.py
```diff
@@ -0,0 +1,61 @@
...
+import numpy
+from bpy.props import *
+from ... data_structures import DoubleList
+from ... base_types import AnimationNode, VectorizedSocket
+
+reductionItems = [
+    ("MEAN", "Mean", "", "", 0),
+    ("MAX", "Max", "", "", 1)
+]
+reductionFunctions = {
```
JacquesLucke:
```suggestion
reductionFunctionsItems = {
```

animation_nodes/nodes/sound/sound_intensity.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def getExecutionFunctionName(self):
+        if self.useFloatList: return "executeMultiple"
+        else: return "executeSingle"
+
+    def executeSingle(self, sound, frame, attack, release, scene):
+        if len(sound.soundSequences) == 0: self.raiseErrorMessage("Empty sound!")
+
+        fps = scene.render.fps
+        return sound.computeTimeSmoothedIntensity(frame / fps, (frame + 1) / fps,
```
JacquesLucke: I'm still unsure if `frame / fps, (frame + 1) / fps` is really the right approach to calculate the windows, but I guess we can leave it as is for now..

animation_nodes/nodes/sound/sound_spectrum.py
```diff
@@ -0,0 +1,145 @@
...
+        self.newInput("Float", "Amplitude", "amplitude", value = 1)
+
+        if self.samplingMethod == "EXP":
+            self.newInput("Integer", "Count", "count", value = 20, minValue = 1)
+            self.newInput("Float", "Low", "low", value = 0, minValue = 0, maxValue = 1, hide = True)
+            self.newInput("Float", "High", "high", value = 1, minValue = 0, maxValue = 1, hide = True)
+            self.newInput("Float", "K", "k", value = 5, minValue = 0.00001)
+        elif self.samplingMethod == "CUSTOM":
+            self.newInput("Float List", "Pins", "pins")
+        elif self.samplingMethod == "SINGLE":
```
JacquesLucke: I think the `SINGLE` method can be moved completely to the Sound Intensity node (which should also have the high and low input then.

animation_nodes/nodes/sound/sound_spectrum.py
```diff
@@ -0,0 +1,145 @@
...
+
+class SoundSpectrumNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_SoundSpectrumNode"
+    bl_label = "Sound Spectrum"
+    errorHandlingType = "EXCEPTION"
+
+    reductionFunction: EnumProperty(name = "Reduction Function", default = "MAX",
+        items = reductionItems)
+    smoothingSamples: IntProperty(name = "Smoothing Samples", default = 5, min = 0)
+    beta: FloatProperty(name = "Kaiser Beta", default = 6, min = 0)
```
JacquesLucke:
```suggestion
    kaiserBeta: FloatProperty(name = "Kaiser Beta", default = 6, min = 0)
```

animation_nodes/sockets/sound.py
```diff
@@ -1,46 +1,21 @@
...
-        yield ("SPECTRUM_{}_{}".format(sequenceIndex, bakeIndex),
-               "#{} - {} - Spectrum".format(bakeIndex, sequence.name),
-               "Attack: {:.3f}  Release: {:.3f}".format(data.attack, data.release),
-               strToEnumItemID(data.identifier))
+    i = 0
+    for scene in bpy.data.scenes:
+        if scene.sequence_editor is not None:
+            for strip in scene.sequence_editor.sequences_all:
+                if strip.type == "SOUND":
+                    items.append((strip.name, strip.name, "", i))
```
JacquesLucke: This simple approach might be problematic when sequences are removed.
When a sequence is removed, many enum items change their values -> the selected item changes without the user noticing.

Can you try that?

animation_nodes/sockets/sound.py
```diff
@@ -50,48 +25,44 @@ class SoundSocket(bpy.types.NodeSocket, AnimationNodeSocket):
...
-            return None
-
-        if soundType == "AVERAGE": return AverageSound.fromSequences([sequence], int(bakeIndex))
-        if soundType == "SPECTRUM": return SpectrumSound.fromSequences([sequence], int(bakeIndex))
-        return None
+        if self.soundSequence == "NONE": return Sound([])
+        for scene in bpy.data.scenes:
+            if scene.sequence_editor is not None:
+                sequence = scene.sequence_editor.sequences_all.get(self.soundSequence)
+                if sequence is not None: break
```
JacquesLucke: just return instead of `break`

animation_nodes/nodes/sound/sound_intensity.py
OmarSquircleArt: It might make sense to ensure a certain spectral resolution in the Sound Spectrum node as you suggested. However, I don't see why that would make sense for the sound intensity node. The intensity is computed directly from the samples and have no consequence on low sample counts, in fact, it is more accurate.

I tried with multiple frame rates and the result was nice, so I agree, we should leave this for now.

animation_nodes/nodes/sound/sound_intensity.py
```diff
@@ -0,0 +1,63 @@
...
+    smoothingSamples: IntProperty(name = "Smoothing Samples", default = 5, min = 0,
+        description = ("The number of frames computed to smooth the output."
+        " High value corresponds to more accurate results but with higher execution time"))
+
+    useFrameList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput("Sound", "Sound", "sound")
+        self.newInput(VectorizedSocket("Float", "useFrameList",
+            ("Frame", "frame"), ("Frames", "frame")))
```
JacquesLucke:
```suggestion
            ("Frame", "frame"), ("Frames", "frames")))
```

animation_nodes/nodes/sound/sound_spectrum.py
OmarSquircleArt: I think the user should expect this option to be in the Sound Spectrum node. I associate the *"Sound Intensity"* with the overall sound intensity and not the amplitude of a certain frequency bin. One might think of it as follows: If one wants to sample overlapping frequency bins, one would use a loop with the `SINGLE` option to get the spectral data. Thus the option was used to compute spectral data and not sound intensity, so we ask, what node should be used here? Moreover, the Sound Intensity node is already simple with no enums, so why crowd it with an extra option that can be in another place that is—in my opinion—more appropriate?

So I think the `SINGLE` option should stay in the Sound Spectrum node.

JacquesLucke: I just wonder about the typical workflow. Usually when someone wants to animate a single value using sound, he would use the Sound Intensity node. However often more control is needed, so later in the process he replaces the Sound Intensity node with a Sound Spectrum node.

With this option the Sound Intensity node is redundant and should be removed.
Maybe the best solution is to add an additional search result for the Sound Spectrum node that initializes the Single option directly?

animation_nodes/sockets/sound.py
OmarSquircleArt: I see. I tried that and this is indeed what happens. How should we fix this?

JacquesLucke: You can use a similar approach to the one I used before: Use some kind of hash to compute the value. This is definitely not a perfect solution but worked relatively well in my experience.

animation_nodes/nodes/sound/sound_spectrum.py
OmarSquircleArt: This is not the only use for the single option. The single option can be used inside loops to generate frequency bins that are otherwise not possible to generate using the rest of the options, like overlapping frequency bins. So the user might be tempted to think that this is not possible because she is looking in the wrong place, that is, the Sound Spectrum node instead of the Sound Intensity. At least that's where I would look for this option.

That might be a solution, but I am not sure about the relation between the sound intensity and the frequency domain. Will research this and get back to you.

OmarSquircleArt: Ok, I think the Sound Intensity node can be removed and a search entry can be added.
So what should we do?

animation_nodes/sockets/sound.py
OmarSquircleArt: I think it works properly now. However, blender will return a warning about the missing enum ID. The warning is not fatal or anything, so we can just ignore it and return an empty sound, which is what I did.

animation_nodes/nodes/sound/sound_intensity.py
OmarSquircleArt: Since we might remove this node anyway, I will leave this for now.

animation_nodes/nodes/sound/sound_spectrum.py
JacquesLucke: Remove the Sound Intensity node and add the search entry.

OmarSquircleArt: Ok, done.

OmarSquircleArt: I also removed the amplitude parameter and set the default amplitude to 10. You were right, users can easily make use of it in that order of magnitude.

animation_nodes/data_structures/sounds/sound.py
```diff
@@ -0,0 +1,44 @@
...
+from functools import lru_cache
+from math import ceil, log, isclose
+
+class Sound:
+    def __init__(self, soundSequences):
+        self.soundSequences = soundSequences
+
+    def getSamplesInRange(self, start, end):
+        if end <= start: raise ValueError("Invaild range!")
+        maxSampleRate = max(sequence.data.sampleRate for sequence in self.soundSequences)
```
JacquesLucke: Somehow it feels wrong that you evaluate some sounds using a wrong sample rate, or do I not read that correctly?

animation_nodes/nodes/sound/sound_spectrum.py
```diff
@@ -0,0 +1,152 @@
...
+            attack, release, self.smoothingSamples, self.kaiserBeta)
+        maxFrequency = len(spectrum) - 1
+
+        scale = expm1(k) / (high - low)
+        pins = [low + expm1(k * i / count) / scale for i in range(count + 1)]
+
+        bars = DoubleList(count)
+        reductionFunction = reductionFunctions[self.reductionFunction]
+        for i in range(count):
+            x, y = int(pins[i] * maxFrequency), int(pins[i + 1] * maxFrequency)
```
JacquesLucke: I think the computation of the frequency bins could be moved to a separate function. This function should be a generator function that yields the start and end of every frequency bin.

I believe that this way, the `executeExponential`, `executeCustom` and `executeFull` methods could share more code. Currently there is quite a lot of code duplication going on if I see that correctly.

animation_nodes/data_structures/sounds/sound.py
OmarSquircleArt: You are right, this should be a constant sample rate.

animation_nodes/nodes/sound/sound_spectrum.py
OmarSquircleArt: Alright. Time to know how to use `yield`.

OmarSquircleArt: hmmm, looking at this again, each execution function have different inputs, so we can't join them into a single function. Do you mean that we should use a `getExecutionCode` function? Moreover, there are slight differences in the for loop used to compute the frequency bins, so we can use a unified generator.
Can you elaborate?

OmarSquircleArt: Ok, but three cases? The `executeFull` is very different from the other two. I think including it in the helper function would needlessly complicate it, don't you think? After all, it is just two lines of code.

JacquesLucke: hmm ok, leave it as is.

JacquesLucke:
We probably have to change the way the sample rate is managed at some point. It can't really be expected that all sounds have the same. However that can wait a little longer. Looks good to me know, feel free to merge it.

OmarSquircleArt:
@JacquesLucke We are not assuming that all sounds have the same sample rate. In the `getCachedSoundData`, we are resampling to a constant 44100 sample rate if we find the sample rate to be different. We do it like this to avoid resampling during execution, which can be expensive.

Will probably merge it after some final testing. Squash and merge, right?

JacquesLucke:
Do you have any numbers about how expensive resampling is compared to the fft?

Yes, squash and merge.
Squash means that all your commits will be combined into a single commit in the Blender 2.8 branch.
That usually makes the git history a bit cleaner. (except you squash too many commits ^^)

OmarSquircleArt:
I don't have an implementation to do tests right now, so I can't give you numbers. The resampling algorithm that I know and I think is used by scipy's `resample` function is the Whittaker Shannon interpolation, this interpolation uses FFT and Inverse FFT to do the resampling, so it is defintly more expensive that FFT.

PR #1000, opened 2019-01-14
Added Material Socket and Set Object Materials Node.

OmarSquircleArt:


animation_nodes/nodes/material/set_object_materials.py
```diff
@@ -0,0 +1,35 @@
+import bpy
+from ... base_types import AnimationNode, VectorizedSocket
+
+class SetObjectMaterialsNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_SetObjectMaterialsNode"
+    bl_label = "Set Object Materials"
+
+    useList: VectorizedSocket.newProperty()
```
JacquesLucke:
```suggestion
    useMaterialList: VectorizedSocket.newProperty()
```

animation_nodes/nodes/material/set_object_materials.py
```diff
@@ -0,0 +1,35 @@
...
+        self.newOutput("Object", "Object", "object")
+
+    def getExecutionFunctionName(self):
+        if self.useList:
+            return "executeList"
+        else:
+            return "executeSingle"
+
+    def executeSingle(self, object, material):
+        if object == None or not hasattr(object.data, "materials"): return object
```
JacquesLucke:
```suggestion
        if object is None or not hasattr(object.data, "materials"): return object
```

Usually checking if something is `None` is done using `is` instead of `==`.

OmarSquircleArt:
@JacquesLucke AN example for the Set Object Materials.
materials

Maybe we should rename it to be more consistent with the other output nodes.

animation_nodes/nodes/material/material_output.py
```diff
@@ -0,0 +1,48 @@
...
+            socket.isUsed = False
+        for socket in self.inputs[4:]:
+            socket.hide = True
+
+        self.newOutput("Material", "Material", "material")
+
+    def getExecutionCode(self, required):
+        yield "if material is not None:"
+        yield "    pass"
+        for i, (socketType, name, attr) in enumerate(attributes):
```
JacquesLucke:
```suggestion
        for i, (socketType, name, attr) in enumerate(attributes, 1):
```

PR #1002, opened 2019-01-15
Fixed compose matrix node and added center options and vectors output for distribute matrices node.

OmarSquircleArt:


animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -131,16 +154,17 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             yDis = size2 / max(yDiv - 1, 1)
             zDis = size3 / max(zDiv - 1, 1)
 
-        xOffset = xDis * (xDiv - 1) / 2
-        yOffset = yDis * (yDiv - 1) / 2
+        xOffset = xDis * (xDiv - 1) / 2 * self.centerAlongX
```
JacquesLucke:
```suggestion
        xOffset = xDis * (xDiv - 1) / 2 * int(self.centerAlongX)
```
just to be a little bit more explicit

JacquesLucke: After that change you can merge it. Can't see any reason not to merge it currently.

PR #1008, opened 2019-01-19
Add Viewport Input node.

OmarSquircleArt:
Added Viewport Input node as an alternative to the *Show In Viewport* option.


OmarSquircleArt:
One thing we need to take care of is to limit the selector to types that have a draw property. I can't seem to find a way to do it using the current options. I am assuming we have to add another option to the `ChooseSocketType` operator?

JacquesLucke:
Yes, you can add a new type to the socket type chooser.

OmarSquircleArt:
Alright, just added the new type.

PR #1012, opened 2019-01-21
Blender2.8 background mode

Wcubed:
Edits I made after trying to render an animation node scene in blender's background mode.

It looks like the only problems were the calls to functions of the `gpu` module, which is not allowed when there is no UI. So I added guards around them checking whether we are in background mode or not.

No idea if this is enough to have all the features work. The startup tests return positive though, and I've rendered out a basic sound visualizer as a test.

I've also updated the build requirements to mention the right blender and python version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 *.a
 *.o
 .vscode
+.idea
 build/
 conf.json
 animation_nodes/compilation_info.json
--- a/animation_nodes/graphics/drawing_2d.py
+++ b/animation_nodes/graphics/drawing_2d.py
@@ -1,10 +1,16 @@
+import bpy
 import gpu
 import blf
 from bgl import *
 from gpu_extras.batch import batch_for_shader
 
 dpi = 72
-shader = gpu.shader.from_builtin('2D_UNIFORM_COLOR')
+
+# Can't use OpenGl functions when running in background mode.
+if not bpy.app.background:
+    shader = gpu.shader.from_builtin('2D_UNIFORM_COLOR')
+else:
+    shader = None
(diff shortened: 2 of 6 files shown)


brothermechanic:
Thank you!
Patch working perfect!
Hope it will be added in master

pit:
I've tried to use this PR in animation_nodes with blender2.8 in background mode.
Got error when running:

`Traceback (most recent call last):
  File "/opt/blender-2.80/2.80/scripts/addons/animation_nodes/utils/handlers.py", line 61, in loadPost
    handler()
  File "/opt/blender-2.80/2.80/scripts/addons/animation_nodes/events.py", line 64, in fileLoaded
    tree_info.updateIfNecessary()
  File "/opt/blender-2.80/2.80/scripts/addons/animation_nodes/tree_info/__init__.py", line 33, in updateIfNecessary
    return getPreferences().developer.debug
  File "/opt/blender-2.80/2.80/scripts/addons/animation_nodes/preferences.py", line 191, in getPreferences
    return bpy.context.preferences.addons[addonName].preferences
    updateFile()
    update()
  File "/opt/blender-2.80/2.80/scripts/addons/animation_nodes/utils/timing.py", line 13, in wrapper
  File "/opt/blender-2.80/2.80/scripts/addons/animation_nodes/utils/timing.py", line 16, in wrapper
KeyError: 'bpy_prop_collection[key]: key "animation_nodes" not found'`

Is that build error? Or usage error?

OmarSquircleArt:
Thanks for the patch! A simpler fix has been implemented in 802e3ef. So closing this for now.

PR #1024, opened 2019-02-02
Added End Point option to Float Range and Elementwise option to Compare node.

OmarSquircleArt:


OmarSquircleArt:
Compare functions are now generated using the pypreprocessor.

OmarSquircleArt:
We will not be vectorizing the compare node for now.

PR #1025, opened 2019-02-06
Branch having some extra nodes

3DSinghVFX:
This branch has some new nodes which are in "newnodes" directory. These nodes are, Set vertex weight node, Set bevel vertex weight node, Set bevel edge weight node, Shader controller node, and Material Instances node.

**Set Vertex Weights Node:**
This node requires four inputs:
 Object --> Mesh Object
 Vertex Group Index --> 0, 1, 2 etc
 Weights --> float list input and length of this list should be equal to the number of vertex locations
 Mode --> REPLACE, ADD or SUBTRACT

_Limitations:_ 1) This node does not work when the input object has Use Modifier option is ON.
2) Changes weight does not update in the viewport but during rendering it does. However, if we add object attribute output having Subdivision Modifier with level = 0, then it works.
3) This node does not havecache option.
vertexweight
   
**Set Bevel Vertex/Edge Weight Node:**
This node requires four inputs:
 Object --> Mesh Object
 Weights --> float list input and length of this list should be equal to the number of vertex locations and for bevel edge weight, length should be equal to the number of edges (or edges center).

_Limitations:_ 1) This node does not work when the input object has Use Modifier option is ON.
2) Sometimes changes in bevel weight do not update in the viewport.
3) This node does not have cache option.
bevelweight

**Shader Controler Node:**
This node requires four inputs:
 Materials --> Material List
 Shader Node Name --> Like Diffuse BSDF, Mix Node etc
 Shader Property Name --> Color, Roughness, Factor etc
 Values --> Values of Shader Property

_Limitations:_ 1) This node does not have cache option.

**Material Instancer Node:**
This node requires four inputs:
 Instance Material --> Boolean, It is like the switch to instance the material (keep off for better performance). 
Remove Material --> Boolean, It is like the switch to remove instanced the material.
Base Material Name  --> Material which going to be instanced
Prefix for instance materials  --> A prefix text to recognize the instanced materials
Amount --> Number of copies of Base Materials

_Limitations:_ 1) Blender Sometimes crashed when I removed the instanced materials mainly when AN in Auto Execution mode
2) This node does not have cache option.

_For demo of shader controller node and material instancer node I used two nodes of Alan Odom "https://github.com/Clockmender/My-AN-Nodes"._
Material Instancer Node:-
peek 2019-02-06 22-41

Shader Controller Node:-
peek 2019-02-06 22-42


3DSingh:
I forgot to include the blend files for extra nodes. I have attached a zip file which has blend file for each node setup for blender 2.79 (latest build).
[ExtraNodeBlendFiles.zip](https://github.com/JacquesLucke/animation_nodes/files/2837530/ExtraNodeBlendFiles.zip)

OmarSquircleArt:
Putting the content of this patch aside for the moment, here are some pointer on how to contribute to Animation Nodes:

- You should follow the same coding style used in the rest of the projects, this include the spacing, indentation, and variable naming style. Check existing code.
- You should use Blender 2.8 python API. Check the release note to see what you should update. This include using type annotations for class properties.
- For error reporting, use UI Extensions instead of string properties. See any node that have the `errorHandlingType` property to know how to use this feature.
- Don't use comments and global variables excessively, sometimes they are just redundant.
- All nodes should have at least one output. Data setter nodes should have their input object as output. See any output node.
- Accessing `bpy_prop_collection` attributes should be done using the `foreach_get` and `foreach_set` because they are much faster. See `MeshProperties` class to know how use this.
- Please review that patch more closely. Some blocks are wrongly indented and you literally have a file that say `hi` ...

OmarSquircleArt:
Now to discuss the need for and implementation method for the nodes.

- **Set Vertex Weight**: If I remember correctly, Jacques didn't like the performance of this method, so we have to find a more performant method.
- **Shader Control**: This node already exist a *Cycles Material Output*. Moreover, this version is rather unpredictable because node trees have no defined form. What do you think?
- **Material Instancer**: Why do we need this node? If one wants value variations across objects, one can pass this information to the source material in a more efficient manner, instead of copying the same material and only changing a single value. What do you think?
- **Bevel Vertex/Edge Weight**: While there is no technical reason not to add this node. The question is if it is worth adding or not. Its use case seem rather limited and its implementation seems simple, so will an expression or script node be sufficient in those use cases?

I don't want to make you feel discouraged by those comments, I just want a more detailed reasoning to help in making better decisions. Thanks!

3DSinghVFX:
Hi Omar...
Based on your suggestions I have rewritten the codes of extra nodes in AN style and modified (& added)  some of the sockets of nodes to match AN nodes criteria. Here is the image of nodes:
screen

Here is my opinion for the need of extra nodes:
 **Set Vertex Weights**: Yes, I know that performance will degrade if the number of vertices are very large but applications of this node are many (to control particles, cloth simulation, most of the modifier have vertex group input to control the effects). I have been trying to write this in Cython gain some performance and the second possibility is to add bake system.

 **Set Vertex Colors**: The current version of AN has Set Vertex Color node but set the single color for all vertices. But Set Vertex Colors node allow to set vertex color per vertices, so we can change/control shaders on the different parts of the object. Another thing is, we can convert the vertex weight to vertex color with this node.

 **Vertex Color Input**: This node allows to get vertex color data i.e, we can get vertex color, modify vertex colors then set back with Set Vertex Colors node. Another thing is, we can also convert the vertex colors to vertex weight.

 **Set bevel vertex/edge weight**: I think these nodes are an easy way to control the bevel weight. Because to control the bevel weight, we have to first enable this option "use_customdata_edge_bevel" then we can set/change the bevel weight and if we are doing for all the vertices or edges then we need also the for loop. So these nodes are important.

 **Texture Input**: This node gives the texture _(works for blender internal and external textures, images or movie)_ data (colors and red, green, blue, alpha) based on input vectors location.  That means we can use texture colors for anything in AN e.g, we can set vertex weight or vertex color directly from textures or now the same material texture can also control the animation if we need.

 **Material Instancer**: I know that for large number of objects it is not best to copy materials but smoetimes we need different materials for different objects, particulary when we want to control more than one properties of a material.

 **Shader Controller**: You are right there is node _Cycles Material Output_ in AN. But this node does not allow input material from outside, I means material socket is property type. So, it is impossible to use this in a loop over material list. If it is possible to input material in loop in exsiting _Cycles Material Output_ then there is no need of Shader Controller node.

Thanks

OmarSquircleArt:
@3DSinghVFX Hi,

- **Vertex Weight:** The problem is with the python API access itself, so Cython won't be much help. Lets wait and see what Jacques think about this.
- **Vertex Color/Texture:** The color system we have now is very limited, so I think any node that deals with lists of colors should be skipped for now, at least until we rewrite the color system. We also need to investigate 2D vectors before making an image sampler.
- **Material Instancer/Shader Controller:** Can you provide a practical example where this would be mist useful?

3DSinghVFX:
Hi @OmarSquircleArt,
Here is an example of Material and ShaderController nodes. In this example, I'm changing metallic and roughness property with Wiggle Falloff, and at the same time, I'm also controlling the Base Color with Object Controller Falloff with orbiting object. Node Tree:
 
materialinstancer1
materialinstancer2

peek 2019-03-08 15-53

Thanks.

OmarSquircleArt:
@3DSinghVFX Thanks for the example! I feel like material instancing is still an inefficient way to do this. I think Cycles developers should fix this, not us. Lets wait for Jacques and see what he thinks.

animation_nodes/nodes/newnodes/gpencils_stroke_line_width.py
```diff
@@ -0,0 +1,58 @@
...
+        if len(strokes) == 0 or len(lineWidths) == 0 or len(strokes) != len(lineWidths): return strokes
+        outStrokes = []
+        for i, stroke in enumerate(strokes):
+            if stroke is not None:
+                outStroke = self.copyStroke(stroke)
+                outStroke.line_width = lineWidths[i]
+                outStrokes.append(outStroke)
+        return outStrokes
+
+    def copyStroke(self, stroke):
```
JacquesLucke: Just a quick note: you probably don't have to copy the stroke yourself every time. AN has builtin functionality that makes copies when necessary. Check how it is done for e.g. meshes or lists.

You need to define a copy expression in the socket, so that AN knows how to copy your type. Then, when a socket input is modified, you can use the `dataIsModified` property to tell AN that this input will be modified. It's not a perfect solution, but it has less overhead then your current approach.

3DSinghVFX: Thanks, @JacquesLucke for the suggestion. Initially, I added the copy option to the socket but it was not working probably I was missing the "dataIsModified" option. I'll try my best to add this in gpencil node codes. Sorry for the late reply.

JacquesLucke:

PR #1027, opened 2019-02-08
Added subprogram conditional invoking.

OmarEmaraDev:
Added a boolean input to the invoke node. This boolean is a condition that determine if a subprogram should be invoked or not. If not, default values are returned. The condition is hidden by default.


mathlusiverse:
An example blander file was created specifically to test this feature, please see
https://github.com/JacquesLucke/animation_nodes/issues/1026#issuecomment-462014484

The feature works correctly as planed.

PR #1031, opened 2019-02-13
Fixed TypeError when converting Plane type to Matrix

dragere:
Just a small fix when converting Point/Normal to Matrix, because * is an element-wise multiplication in 2.8.


OmarSquircleArt:
Thanks!

Issue #928, opened 2018-08-02
Vectorized sockets stops automatic conversion upon duplication.


OmarSquircleArt:
This bug has been happening since the refactor of the vectorization code, in this issue, I shall report the situations where it happens to possibly narrow its cause.

## Situation 1

A node with a vectorized input is connected to a list, if this node is duplicated and the input of the duplicate is connected to another socket---Single or list(possibly the same socket as the original node), the original node gets disconnected. If the type connected to the duplicate was a list and not a single, the original node input won't accept lists, but only singles. If the type connected to the duplicate was a single, the original node accepts both as predicted.

![duplicating_bug](https://user-images.githubusercontent.com/25300994/43582069-5ee6ef3a-965b-11e8-9913-840ce0656287.gif)

***
I will try to add other situations if encountered.

mattdw:
It's interesting that it disconnects the socket of the original copy – my guess would be that something isn't getting deep-copied on duplication and there's an unexpected shared object underneath? I have the code checked out and building now, so I might see if I can use this as a chance to explore the code a bit.

PR #1033, opened 2019-02-17
fix for #928 – clone data on node duplication

mattdw:
Hi,

I've fixed (I think) a data-sharing problem on node duplication, noted in issue https://github.com/JacquesLucke/animation_nodes/issues/928

I'm not sure if copying inputs/outputs in my clone() method is strictly necessary, as links aren't maintained, but it seemed the smallest change. Let me know if I need to rethink any of this!


OmarSquircleArt:
Thanks! I did some tests and it seems to work as expected.

Issue #855, opened 2018-01-22
Why does Rotation-Wiggle node not have 'list option' ?


3DSinghVFX:
Hi,
Vector-Wiggle node has list option but Number-Wiggle and Rotation-Wiggle nodes don't have list option, why?

![rot_wig_list](https://user-images.githubusercontent.com/30294746/35236727-4ba0c1c2-ffce-11e7-9a1b-5895c7c45717.png)

[rot_wig_list.blend.zip](https://github.com/JacquesLucke/animation_nodes/files/1653082/rot_wig_list.blend.zip)

Singh-KS:
OK. Yes, I'm using the loop for the Rotation-Wiggle list. 

Another query is:- In a group node, is there a way that I can make a "New Input" connection in between the previous exiting connections (see the attached image). Because for a large number of inputs (say ~100), it takes a lot of time to move a particular newly created connection from bottom to top. 
Is it possible to just drag them up or down in the property window (Ctrl+N)? 

Thank You :) 

![screenshot](https://user-images.githubusercontent.com/30294746/35345736-612d8c94-0156-11e8-957e-ca4cc1d4cc1e.png)

JacquesLucke:
unfortunately that is not possible, sry.
Currently there is no better way than to move the socket up manually..

Singh-KS:
OK. Thanks, Jacques :)

mattdw:
(Just noting that Quaternion Wiggle is also not listified yet.)

OmarEmaraDev:
Added in cef7e33.

PR #1034, opened 2019-02-18
Vectorize more nodes (partial fix for #855)

mattdw:
Partial fix for https://github.com/JacquesLucke/animation_nodes/issues/855 – only Number Wiggle, not Euler for now.

I very directly copied the vector wiggle code, so I'm not sure if it's 100% correct or efficient – I'd be keen to hear if there are any problems with it.

There seems like a lot of double calculation around seeds + evolution; I assume that's just to make sure everything has properly unique seeds?

If I get some time I'll come back and do the same thing for the Euler Wiggle node.

### DIFF
--- a/animation_nodes/algorithms/perlin_noise.pyx
+++ b/animation_nodes/algorithms/perlin_noise.pyx
@@ -1,7 +1,7 @@
 cimport cython
 from . random cimport randomDouble_UnitRange
 from .. utils.limits cimport INT_MAX
-from .. data_structures cimport Vector3DList, FloatList
+from .. data_structures cimport Vector3DList, FloatList, EulerList, Euler3
 
 # http://freespace.virgin.net/hugo.elias/models/m_perlin.htm
 
@@ -32,6 +32,19 @@ def wiggleFloatList(amount, double evolution, amplitude, int octaves, double per
     cdef Py_ssize_t i
     for i in range(amount):
         values[i] = perlinNoise1D(evolution + i * 354623, persistance, octaves) * amplitude
+    return result
+
+def wiggleEulerList(amount, double evolution, amplitude, int octaves, double persistance):
+    cdef EulerList result = EulerList(length = amount)
+    cdef Euler3 *values = <Euler3*>result.data
+    cdef float _amplitude[3]
+    cdef Py_ssize_t i
+    _amplitude[0], _amplitude[1], _amplitude[2] = amplitude[0], amplitude[1], amplitude[2]
+    for i in range(amount):
+        values[i].x = perlinNoise1D(evolution + i * 354623, persistance, octaves) * _amplitude[0]
+        values[i].y = perlinNoise1D(evolution + i + 0.33 * 354623, persistance, octaves) * _amplitude[1]
+        values[i].z = perlinNoise1D(evolution + i + 0.66 * 354623, persistance, octaves) * _amplitude[2]
+        values[i].order = 0
     return result
 
 cpdef double perlinNoise1D(double x, double persistance, int octaves):
--- a/animation_nodes/algorithms/perlin_noise.pyx
+++ b/animation_nodes/algorithms/perlin_noise.pyx
@@ -1,7 +1,7 @@
 cimport cython
 from . random cimport randomDouble_UnitRange
 from .. utils.limits cimport INT_MAX
-from .. data_structures cimport Vector3DList
+from .. data_structures cimport Vector3DList, FloatList
 
 # http://freespace.virgin.net/hugo.elias/models/m_perlin.htm
 
@@ -27,12 +27,12 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
     return result
 
 def wiggleFloatList(amount, double evolution, amplitude, int octaves, double persistance):
-    cdef list values = []
+    cdef FloatList result = FloatList(length = amount)
+    cdef float *values = <float*>result.data
     cdef Py_ssize_t i
     for i in range(amount):
-        res = perlinNoise1D(evolution + i * 354623, persistance, octaves) * amplitude
-        values.append(res)
-    return values
+        values[i] = perlinNoise1D(evolution + i * 354623, persistance, octaves) * amplitude
+    return result
 
 cpdef double perlinNoise1D(double x, double persistance, int octaves):
     cdef:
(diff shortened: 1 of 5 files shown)


mattdw:
Now with Euler Wiggle. Just Quaternion to do, I guess… looks like random Quaternions are just generated as 4 random 0..1 floats and then normalized?

mattdw:
Quaternion Wiggle support is in, but I'm really confident about it. I notice the non-list version always has w=1.0, which I couldn't replicate when normalizing was happening. Might need a pointer on the maths here (or if my approach is wrong – I also couldn't figure out how to call normalize() from cython, so I've hand-coded it.)

mattdw:
Just bumping this – I think it's a solid improvement, the only question is whether there's a nicer form for random quaternions (they appear to be rotating correctly still.)

OmarSquircleArt:
@mattdw I agree. Sorry for being late on this. We will try reviewing this as soon as possible.
Thanks for contributing!

animation_nodes/algorithms/perlin_noise.pyx
```diff
@@ -26,6 +27,54 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
         values[i] = perlinNoise1D(evolution + i * 354623, persistance, octaves) * _amplitude[i % 3]
     return result
 
+def wiggleFloatList(amount, double evolution, amplitude, int octaves, double persistance):
```
OmarEmaraDev: Float Lists in Animation Nodes are actually of type `DoubleList` not `FloatList`, so this function should be edited to reflect that.

animation_nodes/algorithms/perlin_noise.pyx
```diff
@@ -26,6 +27,54 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
...
+    cdef float *values = <float*>result.data
+    cdef Py_ssize_t i
+    for i in range(amount):
+        values[i] = perlinNoise1D(evolution + i * 354623, persistance, octaves) * amplitude
+    return result
+
+def wiggleEulerList(amount, double evolution, amplitude, int octaves, double persistance):
+    cdef EulerList result = EulerList(length = amount)
+    cdef Euler3 *values = <Euler3*>result.data
+    cdef float _amplitude[3]
```
OmarEmaraDev: Why is this needed? Just use the components directly.

animation_nodes/algorithms/perlin_noise.pyx
```diff
@@ -26,6 +27,54 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
...
+        values[i].x = perlinNoise1D(evolution + i * 354623, persistance, octaves) * _amplitude[0]
+        values[i].y = perlinNoise1D(evolution + i + 0.33 * 354623, persistance, octaves) * _amplitude[1]
+        values[i].z = perlinNoise1D(evolution + i + 0.66 * 354623, persistance, octaves) * _amplitude[2]
+        values[i].order = 0
+    return result
+
+def wiggleQuaternionList(amount, double evolution, amplitude, int octaves, double persistance):
+    cdef QuaternionList result = QuaternionList(length = amount)
+    cdef Quaternion *values = <Quaternion*>result.data
+    cdef float _amplitude[4]
```
OmarEmaraDev: Same as above.

animation_nodes/algorithms/perlin_noise.pyx
```diff
@@ -26,6 +27,54 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
...
+    cdef float _length
```
OmarEmaraDev: No need for `_`, just call it `length`.

OmarEmaraDev:
Also, please add your changes to the CHANGELOG file.

animation_nodes/algorithms/perlin_noise.pyx
mattdw: I kept my code as close to wiggleVectorList as possible; it does this, and I assumed it was a cython detail. It does appear to work fine without, so I'll fix up my changes.

OmarEmaraDev: I think the Vector Wiggle code needed this because of the `_amplitude[i % 3]` part, so it is not needed here.

mattdw:
@OmarEmaraDev thanks for those suggestions – I've implemented them all and tidied up the Quaternion Wiggle logic, which is now behaving more as expected. A note has been added to the changelog.

animation_nodes/algorithms/perlin_noise.pyx
```diff
@@ -26,6 +27,50 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
         values[i] = perlinNoise1D(evolution + i * 354623, persistance, octaves) * _amplitude[i % 3]
     return result
 
+def wiggleDoubleList(amount, double evolution, amplitude, int octaves, double persistance):
```
OmarEmaraDev: `amplitude` could be typed as `double` here, I think.

animation_nodes/algorithms/perlin_noise.pyx
```diff
@@ -26,6 +27,50 @@ def wiggleVectorList(amount, double evolution, amplitude, int octaves, double pe
...
+def wiggleEulerList(amount, double evolution, amplitude, int octaves, double persistance):
+    cdef EulerList result = EulerList(length = amount)
+    cdef Euler3 *values = <Euler3*>result.data
+    cdef Py_ssize_t i
+    for i in range(amount):
+        values[i].x = perlinNoise1D(evolution + i * 354623, persistance, octaves) * amplitude[0]
+        values[i].y = perlinNoise1D(evolution + i + 0.33 * 354623, persistance, octaves) * amplitude[1]
+        values[i].z = perlinNoise1D(evolution + i + 0.66 * 354623, persistance, octaves) * amplitude[2]
+        values[i].order = 0
+    return result
```
OmarEmaraDev: Sorry for the misunderstanding, but what I meant is that the array of three floats you had before wasn't needed in this case. Instead, just write them as variables outside of the loop:
```python
xAmplitude, yAmplitude, zAmplitude = amplitude[0], amplitude[1], amplitude[2]
```

The same goes for the Quaternion function.

PR #1041, opened 2019-03-07
Added endpoint option to float range node.

OmarSquircleArt:


animation_nodes/nodes/number/number_range.py
```diff
@@ -28,6 +29,8 @@ class NumberRangeNode(bpy.types.Node, AnimationNode):
     floatStepType: EnumProperty(name = "Float Step Type", default = "START_STEP",
         items = floatStepTypeItems, update = AnimationNode.refresh)
 
+    endPoint: BoolProperty(name = "End Point", default = True, update = executionCodeChanged)
```
JacquesLucke: Bool properties be named like things which can be Yes or No. In this case `useEndPoint` or `includeEndPoint` might be better.

animation_nodes/nodes/number/number_range.py
```diff
@@ -57,13 +62,20 @@ def getExecutionFunctionName(self):
...
 
     def execute_FloatRange_StartStep(self, amount, start, step):
         return range_DoubleList_StartStep(amount, start, step)
 
-    def execute_FloatRange_StartStop(self, amount, start, stop):
+    def execute_FloatRange_StartStop_EndPoint(self, amount, start, stop):
         return range_DoubleList_StartStop(amount, start, stop)
+
+    def execute_FloatRange_StartStop(self, amount, start, stop):
+        return range_DoubleList_StartStep(amount, start,
```
JacquesLucke: Better use a guard condition to check for `amount <= 0` here. That makes the expression below a bit simpler.

animation_nodes/nodes/number/number_range.py
OmarSquircleArt: Originally named it after numpy's option. But that makes more sense.

Issue #697, opened 2017-04-29
Implement object groups triggers with multiple ID properties per trigger.


OmarSquircleArt:
Currently, if one were to add triggers for multiple object, he would have to add them manually. Instead, I propose that the user can just group the required objects and add this group as a trigger.

Also, each trigger allow only 1 property to watch over. Instead, I suggest user be able to enter multiple IDs per trigger separated by commas.

OmarSquircleArt:
Implemented in a945cf7.

OmarSquircleArt:
Finally! Thanks!

OmarSquircleArt:
Glad I could help, should we close this then?

OmarSquircleArt:
Yah sure.

JacquesLucke:
That conversation was weird ^^

Clockmender:
Sometimes the only way you hear sense is if you talk to yourself... 🥇

Cheers!

OmarSquircleArt:
haha, never though someone would see this :smile: .

JacquesLucke:
I got 4 mails, it's hard not to see those :p

OmarSquircleArt:
Ah sorry, I though only those subscribed to the issue got those emails. :smile:

Clockmender:
^^ If only that were true..... You never know who's watching you :-)

PR #1045, opened 2019-03-09
Refactor and improve Execution Triggers.

OmarSquircleArt:
- The data path can now be a list of comma separated paths.
- Added a collection ID type where the the path is resolved for every object in the collection.
- The `triggerType` was partially removed for now since it is always constant.
- Execution Trigger UI and creation workflow have been redesigned. The trigger now defaults to object without a popup menu. Each trigger is now a UI box with a header that contains the minimize, the enable, and the remove buttons and a column that contains the ID type, the object selector, and the paths. The row-only structure was very crowded, small and uncomfortable to use. Here is how it looks:

New UI

The enable test is dynamic, it specifies which objects and which paths are monitored as you can see in the example above.

I was thinking of adding a way to let the user define the properties using a python expression. What do you think?

### DIFF
--- a/animation_nodes/base_types/tree_auto_execution.py
+++ b/animation_nodes/base_types/tree_auto_execution.py
@@ -20,6 +20,7 @@ class AutoExecutionTrigger_MonitorProperty(bpy.types.PropertyGroup):
     dataPath: StringProperty(name = "Data Path", default = "")
 
     lastState: StringProperty(default = "")
+    expanded: BoolProperty(default = True)
     enabled: BoolProperty(default = True)
     hasError: BoolProperty(default = False)
 
@@ -57,28 +58,35 @@ class AutoExecutionTrigger_MonitorProperty(bpy.types.PropertyGroup):
             return self.scene
 
     def draw(self, layout, index):
-        row = layout.row(align = True)
-        if self.hasError:
-            row.label(text = "", icon = "ERROR")
+        box = layout.box()
 
-        icon = "LAYER_ACTIVE" if self.enabled else "LAYER_USED"
-        row.prop(self, "enabled", icon = icon, text = "")
+        header = box.row()
+        icon = 'TRIA_DOWN' if self.expanded else 'TRIA_RIGHT'
+        header.prop(self, "expanded", icon = icon, text = "", emboss = False)
+        if self.hasError: header.label(text = "", icon = "ERROR")
 
-        row.active = self.enabled
+        enableText = "Enable "
+        object = self.getObject()
+        if object is not None: enableText += object.name + "." + self.dataPath
+        header.prop(self, "enabled", text = enableText, toggle = True)
 
-        props = row.operator("an.assign_active_object_to_auto_execution_trigger", icon = "EYEDROPPER", text = "")
-        props.index = index
+        header.operator("an.remove_auto_execution_trigger", icon = "X",
+            text = "", emboss = False).index = index
 
-        if self.idType == "OBJECT":
-            row.prop(self, "object", text = "")
-        elif self.idType == "SCENE":
-            row.prop(self, "scene", text = "")
+        if self.expanded:
+            col = box.column(align = True)
+            col.active = self.enabled
+            col.prop(self, "idType", text = "")
+            row = col.row(align = True)
+            row.operator("an.assign_active_object_to_auto_execution_trigger",
+                icon = "EYEDROPPER", text = "").index = index
 
-        row.prop(self, "dataPath", text = "")
+            if self.idType == "OBJECT":
+                row.prop(self, "object", text = "")
+            elif self.idType == "SCENE":
+                row.prop(self, "scene", text = "")
 
-        props = row.operator("an.remove_auto_execution_trigger", icon = "X", text = "")
-        props.triggerType = "MONITOR_PROPERTY"
-        props.index = index
+            col.prop(self, "dataPath", text = "")
 
 
 class CustomAutoExecutionTriggers(bpy.types.PropertyGroup):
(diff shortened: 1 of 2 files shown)


animation_nodes/base_types/tree_auto_execution.py
```diff
@@ -16,69 +14,96 @@ class AutoExecutionTrigger_MonitorProperty(bpy.types.PropertyGroup):
...
+                    properties.extend(obj.path_resolve(p.strip()) for p in paths)
+                return properties
+            else:
+                return [object.path_resolve(p.strip()) for p in paths]
         except:
             self.hasError = True
-            return None
+            return []
 
     def getObject(self):
```
JacquesLucke: Maybe change this function so that it returns a list of ID blocks.
Then call does not need to have a special case for collections.

animation_nodes/base_types/tree_auto_execution.py
```diff
@@ -16,69 +14,96 @@ class AutoExecutionTrigger_MonitorProperty(bpy.types.PropertyGroup):
...
+            text = "", emboss = False).index = index
 
-        if self.idType == "OBJECT":
-            row.prop(self, "object", text = "")
-        elif self.idType == "SCENE":
-            row.prop(self, "scene", text = "")
+        if self.expanded:
+            col = box.column(align = True)
+            col.active = self.enabled
+            col.prop(self, "idType", text = "")
```
JacquesLucke: Maybe it would be better if it is impossible to change the `idType`.
Then the API can become cleaner and it is easier to avoid references to objects/scenes/collections that are not actually used by the trigger.

OmarSquircleArt: What do you mean by "not actually used by the trigger"?

OmarSquircleArt: I wanted to provide the enum directly because it is much faster to use, compared to deleting and adding again.

animation_nodes/base_types/tree_auto_execution.py
```diff
@@ -16,69 +14,89 @@ class AutoExecutionTrigger_MonitorProperty(bpy.types.PropertyGroup):
...
+            for obj in objects:
+                properties.extend(obj.path_resolve(p.strip()) for p in self.dataPaths.split(","))
+            return properties
         except:
             self.hasError = True
-            return None
+            return []
 
-    def getObject(self):
+    def getObjects(self):
```
JacquesLucke: Should be renamed to `getIDBlocks` or something similar.

animation_nodes/base_types/tree_auto_execution.py
JacquesLucke: For example the trigger could have a reference to a collection AND to a scene, although only one of those is used. This could become problematic when the node tree is appended into a new file because Blender will try to append the linked stuff as well (afaik).

How often do you actually change the type? Not sure if it is worth having this option in the UI.

OmarSquircleArt: I see. So I should bring back the the pop up menu as before?

JacquesLucke: An alternative would be to reset the other fields when the type changes. Whatever you prefer.

animation_nodes/base_types/tree_auto_execution.py
```diff
@@ -50,19 +50,19 @@ def getPropertiesState(self):
 
     def getProperties(self):
         self.hasError = False
-        objects = self.getObjects()
-        if len(objects) == 0 or self.dataPaths.strip() is "": return []
+        IDBlocks = self.getIDBlocks()
+        if len(IDBlocks) == 0 or self.dataPaths.strip() is "": return []
```
JacquesLucke: hmm difficult, looks like a class name. Maybe `idBlocks` is better, or do you have another idea?

animation_nodes/base_types/tree_auto_execution.py
OmarSquircleArt: I chose the later. I think it is more user friendly.

animation_nodes/base_types/tree_auto_execution.py
```diff
@@ -2,83 +2,104 @@
...
+
+    def getProperties(self):
+        self.hasError = False
+        IDBlocks = self.getIDBlocks()
+        if len(IDBlocks) == 0 or self.dataPaths.strip() is "": return []
+
+        try:
+            properties = []
+            for IDBlock in IDBlocks:
+                properties.extend(IDBlock.path_resolve(p.strip()) for p in self.dataPaths.split(","))
```
JacquesLucke: Better make a utility getter that prepares a list of data paths.

animation_nodes/base_types/tree_auto_execution.py
OmarSquircleArt: Not sure, `idBlocks` seems ok, should I use it?

JacquesLucke: yes

animation_nodes/base_types/tree_auto_execution.py
OmarSquircleArt: You mean something like this:
```python
    def getDataPaths(self):
        return [p.strip() for p in self.dataPaths.split(",")]
```

JacquesLucke: yes

OmarSquircleArt:
To self, remember to close https://github.com/JacquesLucke/animation_nodes/issues/697 if this made it.

JacquesLucke:
Would be better if the data paths string would only be split once and not for every id block. Otherwise seems fine now.

OmarSquircleArt:
Fixed. Thanks!

PR #1046, opened 2019-03-11
Mesh from Spline Node: Output a simple Mesh without any extrusions / bevel / shape

OgierM:
image
image

You could obviously pull this off with an Evaluate Spline Node too, but I think it's rather convenient to have this option here aswell. Just trying out stuff / looks is a bit easier this way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 *.zip
 *.a
 *.o
+*.lnk
 .vscode
 build/
 conf.json
--- a/animation_nodes/nodes/spline/mesh_from_spline.py
+++ b/animation_nodes/nodes/spline/mesh_from_spline.py
@@ -3,8 +3,10 @@ from bpy.props import *
 from ... data_structures import Mesh
 from ... base_types import AnimationNode
 from . c_utils import getMatricesAlongSpline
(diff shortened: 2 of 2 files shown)


OmarSquircleArt:
Not sure I understand, can you give a quick example?

OgierM:
This is what I added:


My intention was making it easier to try out different looks when using the slider, because per default it doesn't go lower than 1 - so to see how that looks you would have to add a small setup with the "Evaluate Splines Node"

In it's self it doesn't allow for any particularly new features, but I think it might be able to speed you up, when trying out different looks. On the other hand - I get that this isn't what this node is designed for and that you probably should use a setup with the evaluate splines node instead.

OmarSquircleArt:
I don't think that it is worth it, and it is—as you said—out of the scope of this node. So I am not exactly in favor of this.

OgierM:
Ok 👍

OmarSquircleArt:
@ogierm That was just my opinion, and I could be wrong. Jacques decides if a feature should be added or not. So if you still think it is worth adding, reopen the pull request until Jacques see it an make a decision.

ogierm:
I'll reopen it if I see myself using it very often

Issue #1047, opened 2019-03-11
The Convert Plane Type Node seems to be broken for me


ogierm:
I tried to recreate [this](https://animation-nodes-manual.readthedocs.io/en/latest/user_guide/nodes/geometry/project_point_on_plane.html?highlight=project%20point%20on%20plane) setup from the documentation. This happens:

![image](https://user-images.githubusercontent.com/29630575/54156965-9e41f400-4447-11e9-8f37-1df755c85f44.png)

OmarSquircleArt:
Fixed in 08b995a.

PR #1048, opened 2019-03-12
Fixed issue #1047

dragere:
Basically the same problem as in #1031.


dragere:
Fixed in 08b995a23ae1912485277d33c57c61e8061b3851

OmarSquircleArt:
We fix that seconds apart :smile: . We started writing the changes to a CHANGLOG file as you can see in 08b995a. So make sure to include that in your future commits.

PR #1054, opened 2019-03-19
Blender2.8

Tlilong:


Clockmender:
Small question, why has the Sort List node been removed for B 2.8 - AN 2.1? We seem to only have a sort Number node now. Sort List was very useful for sorting object lists so they matched other list orders for example.

PR #1075, opened 2019-04-02
Added Sound falloff node.

OmarSquircleArt:


JacquesLucke:
Can you post a .gif that shows how it works?

OmarSquircleArt:
@JacquesLucke It pretty much works as before. Here is a test for the *Index Frequency Spectrum* type:

Index Frequency Spectrum

An example for *Falloff Frequency Spectrum* type:

Falloff Frequency Spectrum

And finally the *Average* type:

Average

The main differences from the node we had before is the fact that the interpolation now controls the scale/distribution of frequencies. By default, it is set to *Exponential*, which is usually what the user wants. The other differences are the extra options you can play with.

Node

animation_nodes/nodes/falloff/sound_falloff.pyx
```diff
@@ -0,0 +1,194 @@
...
+        return bins
+
+cdef class Average_Index_SoundFalloff(BaseFalloff):
+    cdef:
+        int fps, smoothingSamples
+        object sound, reductionFunction
+        float frame, scale, attack, release, amplitude, low, high, kaiserBeta
+
+    def __cinit__(self, object sound, float frame, float scale, float attack, float release,
+        float amplitude, float low, float high, int fps, int smoothingSamples, float kaiserBeta,
```
JacquesLucke: better indent long parameter lists twice

animation_nodes/nodes/falloff/sound_falloff.pyx
```diff
@@ -0,0 +1,194 @@
...
+    ("FALLOFF_FREQUENCY", "Falloff Frequency", "NONE", 1)
+]
+
+indexFrequencyExtensionItems = [
+    ("LOOP", "Loop", "", "NONE", 0),
+    ("MIRROR", "Mirror", "", "NONE", 1),
+    ("EXTEND", "Extend Last", "NONE", 2)
+]
+
+reductionFunctionItems = [
```
JacquesLucke: some of these options are shared with other sound nodes, aren't they? Might be better to import them from there.

PR #1078, opened 2019-04-04
Added missing nodes to node menu

dragere:
- Added missing nodes to node menu.
For those in the unsorted category i have either no idea where to put them or i only added the base of a generic node.

- Fixed Material Output and renamed it to Workbench Material Output node

- changed location to Animation Nodes:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fixed bad path resolution in the *Set Keyframe* node.
 - Fixed *Convert Plane Type* node.
 - Fixed *ReturnDefaultsOnExceptionCodeEffect* and *Get Struct* node.
+- Fixed the *fromFloatList* method of the *3DVectorList* structure.
 
 ### Changed
 
--- a/animation_nodes/data_structures/lists/__vector_list_functions.src
+++ b/animation_nodes/data_structures/lists/__vector_list_functions.src
@@ -2,7 +2,7 @@
 def fromFloatList(self, FloatList source):
     if source.length % 3 != 0:
         raise ValueError("length of float list has to be divisable by 3")
-    cdef Vector3DList newList = Vector3DList(length = source.length / 3)
+    cdef Vector3DList newList = Vector3DList(length = source.length // 3)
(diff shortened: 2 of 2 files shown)


animation_nodes/nodes/material/material_output.py
```diff
@@ -16,7 +16,7 @@
 
 class MaterialOutputNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_MaterialOutputNode"
-    bl_label = "Material Output"
+    bl_label = "Workbench Material Output"
```
OmarSquircleArt: Why should we rename it to Workbench? It doesn't seem to be restricted to it.

animation_nodes/nodes/material/material_output.py
```diff
@@ -38,7 +38,7 @@ def getExecutionCode(self, required):
             if self.inputs[i].isUsed: yield f"    self.{setFunction}(material, '{attr}', {attr})"
 
     def setAttribute_Color(self, material, attribute, color):
-        newColor = color[:3]
+        newColor = color[:4]
```
OmarSquircleArt: You can just removed this line and use the `color` variable directly instead of `newColor`.

animation_nodes/nodes/material/material_output.py
dragere: As far as I know, the standard viewport renderer is also Workbench, so the node only affects Workbench.

animation_nodes/nodes/material/material_output.py
dragere: Done in 6083254059e18a91aa0c39ad4e6eea9e5e53da36

dragere:
I tried to reorder them as good as i can in 2213befebc14c91ab1237e678e72953ec18a1369, so im open to critique.

OmarSquircleArt:
For the ordering:
- I think the L System node could be after the Text Length node, with a separator in between.
- The Tilt Spline node could be after Smooth Bezier Spline node.

I will leave the other decisions for Jacques to decide.

PR #1081, opened 2019-04-05
Frame selected

dragere:
Automaticly frame selected nodes, when adding a frame.
frame-selected

As I have created this branch based on my blender2.8 branch, it depends on #1078, but it can still be reviewed, I think.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Added *Include End Point* option to the *Float Range* node.
 - Added collection option to *Execution Triggers*.
+- Added missing nodes to the node menu
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


OmarSquircleArt:
Don't create a pull request that depends on another pull request. You should branch from the blender2.8 branch and add the operator again independent of the other pull request.

OmarSquircleArt:
It seems you created the other pull request from the blender2.8 branch. You should always branch when creating a feature like this. In this case, branch from upstream the blender2.8 branch.

dragere:
Yeah, I just learned that, I'm not very experienced with git yet.
If you approve this pr, I will merge it with my 2.8 branch so that it can be merged straight with #1078.

dragere:
Dependencies fixed in #1085

PR #1085, opened 2019-04-07
Frame selected

dragere:
Automaticly frame selected nodes, when adding a frame.
frame-selected

Same as #1081, but without dependencies.


OmarSquircleArt:
I already provided some reviews in #1081. So please refer to it and update the patch.

dragere:
This pull request is independent and the branch is from the upstream blender2.8 branch.
Is there something I missed?

OmarSquircleArt:
As it turns out, there is already an operator to do that. It has the shortcut of <kbd>Ctrl</kbd> + <kbd>J</kbd>. So closing this pull request.

Issue #1086, opened 2019-04-21
[Feature Request] Adding Collection Operations as replacement for Object Group Operations


radifar:
When I'm trying the Spawn Rigid Body with AN tutorial from Youtube (https://www.youtube.com/watch?v=G1ysjPeLKSQ) I realize that there is no Object Group Operations in AN 2.1 for Blender 2.8. I suspect that this is caused by the change in Blender from Group to Collection. Therefore the Objects can not be moved to RigidBodyWorld Collection. I hope that Object Collection Operations node or something similar is added to fix this (collection) issue.

Best regards,
Radif

OmarSquircleArt:
Thanks for bringing this to our attention! We will look into it.

dragere:
I simply copied the [Group Operations Node](https://github.com/JacquesLucke/animation_nodes/blob/master/animation_nodes/nodes/object/object_group_operations.py) to 2.8 and came across these issues: 
- removing a object from it's sole collection will remove it from the view layer and view port:
 ![inCollection](https://user-images.githubusercontent.com/26360562/56518172-0b64a100-653f-11e9-8b03-264eacf43e39.gif)
 - adding an object to a collection causes it to appear twice:
![toCollection](https://user-images.githubusercontent.com/26360562/56518204-29320600-653f-11e9-8113-f629e67fe33a.gif)

It'll just be a design decision on how to deal with it.

OmarSquircleArt:
@dragere I think this is the user's responsibility. Not sure if Jacques would agree.

dragere:
I think so, too. Should I then create a pull request?

OmarSquircleArt:
@dragere Sure, but better wait for the pull request you have already to be merged.

OmarSquircleArt:
Added in #1088.

PR #1088, opened 2019-04-26
Collection operation node

dragere:
Added *Collection operation* node as it can be seen in #1086.

### DIFF
--- a/animation_nodes/nodes/object/collection_operations.py
+++ b/animation_nodes/nodes/object/collection_operations.py
@@ -4,30 +4,46 @@ from ... base_types import AnimationNode, VectorizedSocket
 class CollectionOperationsNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_CollectionOperationsNode"
     bl_label = "Collection Operations"
+    codeEffects = [VectorizedSocket.CodeEffect]
 
     useObjectList: VectorizedSocket.newProperty()
 
     def create(self):
         self.newInput("Collection", "Collection", "collection", defaultDrawType = "PROPERTY_ONLY")
         self.newInput(VectorizedSocket("Object", "useObjectList",
-            ("Object", "object"),
-            ("Objects", "objects")))
+            ("Object", "object", dict(defaultDrawType = "PROPERTY_ONLY")),
+            ("Objects", "objects", dict(defaultDrawType = "PROPERTY_ONLY"))))
             
         self.newInput("Boolean", "Linked", "linked")
         self.newOutput("Collection", "Collection", "collection")
 
-    def execute(self, collection, objects, linked):
+    def getExecutionFunctionName(self):
+        if self.useObjectList:
+            return "executeMultipleObject"
+        else:
+            return "executeSingleObject"
+
+    def executeMultipleObject(self, collection, objects, linked):
         if collection is None: return collection
         if objects is None: return collection
         
-        if not self.useObjectList: objects = [objects]
-
         for object in objects:
             if object is None: continue
-            if object.name in collection.objects:
-                if not linked: collection.objects.unlink(object)
-            else:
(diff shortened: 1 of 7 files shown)


dragere:
I'll try, but probably not until next week.

OmarSquircleArt:
If you want, we can merge this for now and do the vectorization later. Should we?

dragere:
No, I think I can do this today.

animation_nodes/nodes/object/collection_operations.py
```diff
@@ -0,0 +1,33 @@
...
+class CollectionOperationsNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_CollectionOperationsNode"
+    bl_label = "Collection Operations"
+
+    useObjectList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput("Collection", "Collection", "collection", defaultDrawType = "PROPERTY_ONLY")
+        self.newInput(VectorizedSocket("Object", "useObjectList",
+            ("Object", "object"),
```
OmarSquircleArt: You can set the `defaultDrawType`—or other properties—like this:
```python
("Object", "object", dict(defaultDrawType = "PROPERTY_ONLY"))
```

animation_nodes/nodes/object/collection_operations.py
```diff
@@ -0,0 +1,33 @@
+import bpy
+from ... base_types import AnimationNode, VectorizedSocket
+
+class CollectionOperationsNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_CollectionOperationsNode"
+    bl_label = "Collection Operations"
+
```
OmarSquircleArt: You can use the `VectorizedSocket.CodeEffect` code effect to perform the vectorization automatically. To know how this works, see the `is_inside_volume.py` node for instance.

dragere: I'm not sure if this is what you expected, but I tried to implement it in 827cae1.

animation_nodes/nodes/object/collection_operations.py
dragere: Done in 827cae1.

animation_nodes/nodes/object/collection_operations.py
OmarSquircleArt: Not exactly. The point of the `VectorizedSocket.CodeEffect` is to let Animation Nodes make us the `executeMultiple` function automatically. So we only have to provide the `executeSingle` function. However, since it is a code effect, the node have to be implemented using the `getExecutionCode` function.

Look at the `is_inside_volume.py` again, there is no `executeMultiple`.

dragere: Like this: c515eff?

animation_nodes/nodes/object/collection_operations.py
```diff
@@ -0,0 +1,33 @@
...
+        self.newInput("Collection", "Collection", "collection", defaultDrawType = "PROPERTY_ONLY")
+        self.newInput(VectorizedSocket("Object", "useObjectList",
+            ("Object", "object", dict(defaultDrawType = "PROPERTY_ONLY")),
+            ("Objects", "objects", dict(defaultDrawType = "PROPERTY_ONLY"))))
+            
+        self.newInput("Boolean", "Linked", "linked")
+        self.newOutput("Collection", "Collection", "collection")
+
+    def getExecutionCode(self, required):
+        return "self.executeSingleObject(collection, object, linked)"
```
OmarSquircleArt: You forgot to assign the return, it should be `collection = self.self.executeSingleObject(...)`. Also, please add a new line at the end of the file, you can setup your editor to add this automatically.

OmarSquircleArt:
Thanks!

PR #1156, opened 2019-07-24
Add ChainActionListNode to chain a list of Actions

MagicMau:
While using a loop to generate a list of Actions, I was missing a way to chain the list into a single Action. I have modified the existing ChainAction to allow a list as input to handle this use case.

### DIFF
--- a/animation_nodes/nodes/action/chain_action_list.pyx
+++ b/animation_nodes/nodes/action/chain_action_list.pyx
@@ -1,0 +1,131 @@
+import bpy
+from bpy.props import *
+from ... events import propertyChanged
+from ... base_types import AnimationNode
+from ... graphics import Rectangle
+from ... graphics.drawing_2d import drawVerticalLine
+from ... data_structures cimport (
+    BoundedAction, BoundedActionEvaluator,
+    FloatList, BooleanList
+)
+
+class ChainActionListNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ChainActionListNode"
+    bl_label = "Chain Action List"
+    errorHandlingType = "EXCEPTION"
+
+    __annotations__ = {}
+    __annotations__["relative"] = BoolProperty(name = "Relative", default = True,
+        description = "Use last frame of first action as starting point of the second",
+        update = propertyChanged)
+
+    def create(self):
+        self.newInput("Action List", "Action list", "list")
+        self.newOutput("Action", "Chained Action")
+
+    def draw(self, layout):
+        layout.prop(self, "relative")
+
+    def execute(self, list):
+        if list is None:
+            return list
+
+        length = len(list)
+
+        if length == 0:
+            return None
+
+        if length < 2:
+            return list[0]
+        
+        r = list[0]
+
+        if not isinstance(r, BoundedAction):
+            self.raiseErrorMessage("only bounded actions can be chained")
+
+        i = 1
+        while i < length:
+            b = list[i]
+            if isinstance(b, BoundedAction):
+                r = ChainAction(r, b, self.relative)
+            i += 1
+
+        return r
+
+cdef class ChainAction(BoundedAction):
+    cdef BoundedAction a, b
+    cdef bint relative
+    cdef set channels
+
+    def __cinit__(self, BoundedAction a, BoundedAction b, bint relative):
+        self.a = a
+        self.b = b
+        self.relative = relative
+        self.channels = a.getChannelSet() | b.getChannelSet()
+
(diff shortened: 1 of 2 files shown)


animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+    bl_label = "Chain Action List"
+    errorHandlingType = "EXCEPTION"
+
+    __annotations__ = {}
+    __annotations__["relative"] = BoolProperty(name = "Relative", default = True,
+        description = "Use last frame of first action as starting point of the second",
+        update = propertyChanged)
+
+    def create(self):
+        self.newInput("Action List", "Action list", "list")
```
OmarEmaraDev: Use more explicit names, `list` -> `actions`, `r` -> `chainedAction`, and so on.

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+        self.newOutput("Action", "Chained Action")
+
+    def draw(self, layout):
+        layout.prop(self, "relative")
+
+    def execute(self, list):
+        if list is None:
```
OmarEmaraDev: I don't think `list` can ever be `None`, so this is redundant.

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+
+        if length == 0:
+            return None
+
+        if length < 2:
+            return list[0]
+        
+        r = list[0]
+
+        if not isinstance(r, BoundedAction):
```
OmarEmaraDev: Check the whole list upfront.

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+            self.raiseErrorMessage("only bounded actions can be chained")
```
OmarEmaraDev: Capitalize and end with a period.

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+
+        i = 1
+        while i < length:
```
OmarEmaraDev: Use a `for` loop instead.

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+            b = list[i]
+            if isinstance(b, BoundedAction):
+                r = ChainAction(r, b, self.relative)
+            i += 1
+
+        return r
+
+cdef class ChainAction(BoundedAction):
```
OmarEmaraDev: Aren't those classes identical to those in the Chain Actions node? Can't we just use the already defined classes?

animation_nodes/ui/node_menu.py
```diff
@@ -472,6 +472,8 @@ def draw(self, context):
         insertNode(layout, "an_WiggleActionNode", "Wiggle Action")
         insertNode(layout, "an_FollowSplineActionNode", "Follow Spline Action")
         layout.separator()
+        insertNode(layout, "an_ChainActionListNode", "Chain Action List")
```
OmarEmaraDev: Wouldn't it be better to put that below the *Chain Actions* node?

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+)
+
+class ChainActionListNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ChainActionListNode"
+    bl_label = "Chain Action List"
+    errorHandlingType = "EXCEPTION"
+
+    __annotations__ = {}
+    __annotations__["relative"] = BoolProperty(name = "Relative", default = True,
+        description = "Use last frame of first action as starting point of the second",
```
OmarEmaraDev: Update the description to fit the fact that we are dealing with a list.

animation_nodes/nodes/action/chain_action_list.pyx
```diff
@@ -0,0 +1,131 @@
...
+    def execute(self, list):
+        if list is None:
+            return list
+
+        length = len(list)
+
+        if length == 0:
+            return None
+
+        if length < 2:
```
OmarEmaraDev: Make this `if length == 1`, just to more readable.

JacquesLucke:

PR #1157, opened 2019-07-24
add option to FollowSplinAction to prevent rotation while following the spline

MagicMau:
I wanted to use an Action to have a camera follow a spline. As I only wanted the translation and not the rotation of this action, I have added an option to determine if following a spline should also rotate the object.


OmarSquircleArt:
Thanks for contributing! Just wanted to let you know that we saw both your pull requests and they will be reviewed once we have the time.

JacquesLucke:

PR #1204, opened 2019-10-09
cleanup header ui

samytichadou:
not perfect but allows to keep alignment and readability of Subprograms menu and Remove An nodetree operator (which is now aligned to the right of the header)


samytichadou:
Before :
image

After : 
image

OmarEmaraDev:
Can you solve the conflicts here?

PR #1205, opened 2019-10-09
add information when no Viewport Input node

samytichadou:
tell the user there is no Viewport Input node instead of a blank panel

### DIFF
--- a/animation_nodes/ui/data_input_panel.py
+++ b/animation_nodes/ui/data_input_panel.py
@@ -15,19 +15,20 @@ class ViewportInputPanel(bpy.types.Panel):
         nodes = getNodesByType("an_ViewportInputNode")
         if len(nodes) == 0:
             layout.label(text="No Viewport Input node", icon="INFO")
-        else:
-            for node in nodes:
-                box = layout.box()
+            return
+        
+        for node in nodes:
+            box = layout.box()
 
-                row = box.row()
-                row.label(text = node.label + ":")
-                row.operator("an.toogle_viewport_input_box", text="",
-                    icon='TRIA_RIGHT' if node.hidden else 'TRIA_DOWN',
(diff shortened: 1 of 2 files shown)


samytichadou:
Before : 
image

After : 
image

animation_nodes/ui/data_input_panel.py
```diff
@@ -13,18 +13,21 @@ class ViewportInputPanel(bpy.types.Panel):
     def draw(self, context):
         layout = self.layout
         nodes = getNodesByType("an_ViewportInputNode")
-        for node in nodes:
-            box = layout.box()
+        if len(nodes) == 0:
```
OmarEmaraDev: Can you use a guard condition instead?

samytichadou: you mean like 
```
if len(nodes) == 0:
    layout.label(text="No Viewport Input node", icon="INFO")
    return
for node in nodes:
    box = layout.box()
```

samytichadou: done ! not a really good coder here, i'm not used to this syntax, is it just a readability issue, or does it improve code performance ?

OmarEmaraDev: It probably doesn't affect performance, but it improves readability.

animation_nodes/ui/data_input_panel.py
```diff
@@ -13,13 +13,17 @@ class ViewportInputPanel(bpy.types.Panel):
     def draw(self, context):
         layout = self.layout
         nodes = getNodesByType("an_ViewportInputNode")
+        if len(nodes) == 0:
+            layout.label(text="No Viewport Input node", icon="INFO")
```
OmarEmaraDev: Can you add a period at the end?

PR #1226, opened 2019-11-07
Set Vertex Weights node

3DSinghVFX:
I have added the **Set Vertex Weights** node which allows us to set or control the vertex weights with animation nodes. I have also updated the node-menu and this node is available under: **Add (node-menu) -> Mesh -> Set Vertex Weights**.

Inputs of this nodes are:
Object <-- Mesh Object
Vertex Group Index <-- 0, 1, 2, etc.
Weight/Weights <-- float value or list (length is equal to the number of vertex locations)
Mode <-- REPLACE, ADD or SUBTRACT

Screenshot from 2019-11-08 00-07-30


OmarEmaraDev:
Hi, thanks for contributing!

What does different modes do in this node?

3DSinghVFX:
@OmarEmaraDev Actually, the blender has these modes for vertex weight.
**Replace ->** this mode allows to completely replace the existing vertex weight of the mesh
**Add ->** this mode allows to add vertex weight on top of the existing vertex weight () of the mesh 
**Subtract ->** this mode allows to subtract vertex weight on top of the existing vertex weight of the mesh

OmarEmaraDev:
@3DSinghVFX  I see. I think we should just remove the mode option and make it REPLACE all the time. Do you agree?

3DSinghVFX:
Yes, I agree. I'll make the REPLACE mode all the time and remove the other modes. I have just noticed that these extra modes (ADD and SUBTRACT) only work properly when Auto Execution is set to Frame Changed. Thanks for your suggestions.

3DSinghVFX:
@OmarEmaraDev I have refined the Set Vertex Weights node code as you suggested. Now, the Set Vertex Weights node has similar settings as **Vertex Group Input** node has:
Screenshot from 2019-11-08 18-25-59

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,145 @@
...
+    errorHandlingType = "EXCEPTION"
+
+    mode: EnumProperty(name = "Mode", default = "ALL",
+        items = modeItems, update = AnimationNode.refresh)
+
+    groupIdentifierType: EnumProperty(name = "Group Identifier Type", default = "INDEX",
+        items = groupIdentifierTypeItems, update = AnimationNode.refresh)
+
+    useIndexList: VectorizedSocket.newProperty()
+    useFloatList: VectorizedSocket.newProperty()
```
OmarEmaraDev: Use a more meaningful name, like `useWeightsList`.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,145 @@
...
+    def execute_AllSingle(self, object, identifier, weight):
+        if object is None: return
+        if object.type != "MESH": 
+            self.raiseErrorMessage(noMeshMessage)
+        if object.mode == "EDIT":
+            self.raiseErrorMessage(editMeshMessage)
+
+        vertexGroup = self.getVertexGroup(object, identifier)
+
+        indices = list(range(0, len(object.data.vertices)))
```
OmarEmaraDev: The zero is redundant here.

animation_nodes/ui/node_menu.py
```diff
@@ -370,6 +370,9 @@ def draw(self, context):
         insertNode(layout, "an_VertexGroupInputNode", "Vertex Group Input")
         insertNode(layout, "an_MeshFromSplineNode", "Mesh From Spline")
         layout.separator()
+        insertNode(layout, "an_VertexGroupInputNode", "Vertex Group Input")
```
OmarEmaraDev: `an_VertexGroupInputNode` is already inserted above.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,145 @@
...
+        totalVertices = len(object.data.vertices)    
+        for i, weight in enumerate(weights):
+            if i >= totalVertices: return 
+            vertexGroup.add([i], weight, "REPLACE") 
+        object.data.update()
+        return object            
+     
+    def getVertexGroup(self, object, identifier):
+        try: return object.vertex_groups[identifier]
+        except: return object.vertex_groups.new()
```
OmarEmaraDev: Why create a new group? If no group exists, just raise an exception.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,145 @@
...
+        if object.type != "MESH": 
+            self.raiseErrorMessage(noMeshMessage)
+        if object.mode == "EDIT":
+            self.raiseErrorMessage(editMeshMessage)
+
+        vertexGroup = self.getVertexGroup(object, identifier)
+
+        totalVertices = len(object.data.vertices)    
+        for i, weight in enumerate(weights):
+            if i >= totalVertices: return
```
OmarEmaraDev: Just do a `for i in range(verticesCount):` to avoid this line.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,145 @@
...
+            self.raiseErrorMessage(noMeshMessage)
+        if object.mode == "EDIT":
+            self.raiseErrorMessage(editMeshMessage)
+
+        vertexGroup = self.getVertexGroup(object, identifier)
+
+        totalVertices = len(object.data.vertices)    
+        for i, weight in enumerate(weights):
+            if i >= totalVertices: return 
+            vertexGroup.add([i], weight, "REPLACE")
```
OmarEmaraDev: Use a `virtualDoubleList`.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: If there is no group then Set Vertex Weights node will create a new one otherwise we have to create it manually from the Object Vertex Tab. To avoid this manual step, I have added this option.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: The code below gives the error when _weights_ is equal not to _verticesCount_
```
verticesCount = len(object.data.vertices)    

for i in range(verticesCount):
    vertexGroup.add([i], weights[i], "REPLACE")
```
so I have to add another line for error handling, some thing like this:

```
verticesCount = len(object.data.vertices)    
if len(weights) < verticesCount: return object

for i in range(verticesCount):
    vertexGroup.add([i], weights[i], "REPLACE")
```
Is this correct?

animation_nodes/nodes/mesh/set_vertex_weights.py
OmarEmaraDev: But it is not intuitive and unexpected to the user. The node is supposed to **Set** Vertex Weights, not create vertex groups. If you think it is important to have a fast way to create vertex groups, then I think we should add an operator somewhere in the node, not add them implicitly.

animation_nodes/nodes/mesh/set_vertex_weights.py
OmarEmaraDev: This is what I had in mind:
```python
weights = VirtualDoubleList.create(weights, 0)
for i in range(len(object.data.vertices)):
    vertexGroup.add([i], weights[i], "REPLACE")
```

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: To create a virtual list, e.g., `list = virtualDoubleList.create()`, what arguments are required in _create()_? Can you please give an example of how to use _virtualDoubleList_?

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: Okay, I'll remove this and raise an exception.

animation_nodes/nodes/mesh/set_vertex_weights.py
OmarEmaraDev: The `create()` function takes two required arguments. The first argument is a list object and the second argument is a default value.
A virtual list can fill itself with the default value if it is empty. For instance:

```python
v = virtualDoubleList.create([], 5)
print(v[0]) # Will print 5
print(v[1]) # Will print 5
...
```
A virtual list can also repeat itself when you access elements that are out of bound. For instance:

```python
v = virtualDoubleList.create([0, 2], 5)
print(v[0]) # Will print 0
print(v[1]) # Will print 2
print(v[2]) # Will print 0
print(v[3]) # Will print 2
...

So a virtual lists is never out of bound, regardless of the length of the actual list.

File: animation_nodes/nodes/mesh/set_vertex_weights.py (hunk 5 above)
Comment (3DSinghVFX) [reply animation_nodes/nodes/mesh/set_vertex_weights.py]: 1) This code works fine when `len(object.data.vertices) = len(weights)`, however when `len(object.data.vertices) > len(weights)` then it gives an error and broke the tree. So, we have to raise an exception,
```
if len(weights) < len(object.data.vertices):
    self.raiseErrorMessage("Invalid Weights.")
weights = VirtualDoubleList.create(weights, 0)
for i in range(len(object.data.vertices)):
    vertexGroup.add([i], weights[i], "REPLACE")
```

2) The second thing is _VirtualDoubleList_ gives an error:
Screenshot from 2019-11-08 22-49-54

File: animation_nodes/nodes/mesh/set_vertex_weights.py (hunk 6 above)
Comment (3DSinghVFX) [reply animation_nodes/nodes/mesh/set_vertex_weights.py]: Thanks for the example.

File: animation_nodes/nodes/mesh/set_vertex_weights.py (hunk 5 above)
Comment (OmarEmaraDev) [reply animation_nodes/nodes/mesh/set_vertex_weights.py]: Can you try the `.get(index)` method of the virtual list?

Comment (3DSinghVFX) [reply animation_nodes/nodes/mesh/set_vertex_weights.py]: The code below gives an error
```
weights = VirtualDoubleList.create(weights, 0) 
print(weights.get(0))
```
and the error is:
_AttributeError: 'animation_nodes.data_structures.virtual_list.virtu' object has no attribute 'get'_

Comment (OmarEmaraDev) [reply animation_nodes/nodes/mesh/set_vertex_weights.py]: Ok, it appears I misunderstood things, sorry. You will need to materialize the list before using it. So your code should be something like this:
```python
weights = VirtualDoubleList.create(weights, 0).materialize(verticesCount)
for i, weight in enumerate(weights):
    vertexGroup.add([i], weight, "REPLACE")
```

3DSinghVFX: Thanks, It worked.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: I have renamed the _useFloatList_ to _useWeightsList_.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: Removed the zero from the _range_ function.

animation_nodes/ui/node_menu.py
3DSinghVFX: I have put the _Vertex Group Input_ and _Set Vertex Weights_ nodes into a separate category.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,157 @@
...
+
+        if self.groupIdentifierType == "INDEX":
+            self.newInput("Integer", "Group Index", "groupIndex")
+        elif self.groupIdentifierType == "NAME":
+            self.newInput("Text", "Name", "groupName")
+
+        if self.mode == "INDEX":
+            self.newInput(VectorizedSocket("Integer", "useIndexList",
+                ("Index", "index"), ("Indices", "indices")))
+            self.newInput(VectorizedSocket("Float", "useWeightsList",
```
OmarEmaraDev: The weights socket is shared been both conditions, so we can take it out of the if statement.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,157 @@
...
+    def execute_Index(self, object, identifier, index, weight):
+        if object is None: return
+        if object.type != "MESH": 
+            self.raiseErrorMessage(noMeshMessage)
+        if object.mode == "EDIT":
+            self.raiseErrorMessage(editMeshMessage)
+
+        vertexGroup = self.getVertexGroup(object, identifier)
+        if vertexGroup is None:
+            self.raiseErrorMessage(groupNotFoundMessage)
```
OmarEmaraDev: Since those lines are shared between all functions, can we move them to their own function?

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,157 @@
...
+
+        for i, index in enumerate(indices):
+            vertexGroup.add([index], weights[i], "REPLACE")
```
OmarEmaraDev: Use virtualDoubleList here as well.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: Yes, it is possible. Something like this,
```
    def execute_AllList(self, object, identifier, weights):
        if object is None: return
        vertexGroup = self.check(object, identifier)

        weights = VirtualDoubleList.create(weights, 0).materialize(len(object.data.vertices))
        for i, weight in enumerate(weights):
            vertexGroup.add([i], weight, "REPLACE") 
        object.data.update()
        return object            

    def check(self, object, identifier):
        if object.type != "MESH": 
            self.raiseErrorMessage(noMeshMessage)
        if object.mode == "EDIT":
            self.raiseErrorMessage(editMeshMessage)

        vertexGroup = self.getVertexGroup(object, identifier)
        if vertexGroup is None:
            self.raiseErrorMessage(groupNotFoundMessage)
        return vertexGroup

```

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: Okay. I'll add the virtualDoubleList here.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: It is done.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,131 @@
...
+]
+
+groupIdentifierTypeItems = [
+    ("INDEX", "Index", "Get vertex group based on the index", "NONE", 0),
+    ("NAME", "Name", "Get vertex group based on the name", "NONE", 1)
+]
+
+groupNotFoundMessage = "group not found"
+noMeshMessage = "no mesh object"
+editMeshMessage = "object is not in object mode"
```
OmarEmaraDev: Now that we grouped the exceptions together, there is no need to define those as variables, just put them in the function call. Also, capitalize the first letter and add a period at the end.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,131 @@
...
+
+        if self.groupIdentifierType == "INDEX":
+            self.newInput("Integer", "Group Index", "groupIndex")
+        elif self.groupIdentifierType == "NAME":
+            self.newInput("Text", "Name", "groupName")
+
+        if self.mode == "INDEX":
+            self.newInput(VectorizedSocket("Integer", "useIndexList",
+                ("Index", "index"), ("Indices", "indices")))
+        self.newInput(VectorizedSocket("Float", "useWeightsList",
```
OmarEmaraDev: A line break for readability.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,131 @@
...
+        layout.prop(self, "mode", text = "")
+
+    def drawAdvanced(self, layout):
+        layout.prop(self, "groupIdentifierType", text = "Type")
+
+    def getExecutionFunctionName(self):
+        if self.mode == "INDEX":
+            if self.useIndexList:
+                if self.useWeightsList:
+                    return "execute_IndicesList"
```
OmarEmaraDev: Perhaps using more explicit names would be better:

- `execute_IndicesList` -> `execute_Indices_WeightsList`
- `execute_IndicesSingle` -> `execute_Indices_SingleWeight`
- `execute_AllList` -> `execute_All_WeightsList`
- `execute_AllSingle` -> `execute_All_SingleWeights`

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,131 @@
...
+                else:
+                    return "execute_IndicesSingle"
+            else:
+                return "execute_Index"
```
OmarEmaraDev: What will happen if `useIndexList` is false and `useWeightsList` is true? The `execute_Index` function will be called and the `weight` argument would be a list, not a float.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,131 @@
...
+        if object is None: return
+        vertexGroup = self.check(object, identifier)
+
+        weights = VirtualDoubleList.create(weights, 0).materialize(len(object.data.vertices))
+        for i, weight in enumerate(weights):
+            vertexGroup.add([i], weight, "REPLACE") 
+        object.data.update()
+        return object            
+
+    def check(self, object, identifier):
```
OmarEmaraDev: The function name is not intuitive. The function validates the inputs, throws exceptions, and returns the vertex group. So perhaps the code should be moved to the `getVertexGroup` function.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: Then, it gives the error:
Screenshot from 2019-11-10 10-57-30

Yes, you are right, `weight` argument would be a list. So I think, we add another function which handles this:
```
def execute_Index_WeightsList(self, object, identifier, index, weights):
        if object is None: return
        vertexGroup = self.getVertexGroup(object, identifier)

        vertexGroup.add([index], weights[0], "REPLACE")
        object.data.update()    
        return object
```
or do you have any other idea?

OmarEmaraDev: I believe we can solve this by adding the `useIndexList` to the list of vectorized socket properties like so:

```
self.newInput(VectorizedSocket("Float", ["useWeightsList", "useIndexList"],
    ("Weight", "weight"), ("Weights", "weights")))
```
I think this will break the ALL case though, so we might need to branch again. But I haven't tried that, can you try that?

3DSinghVFX: Okay. I'll try and let you know.

3DSinghVFX: It worked without any error for all the modes (Index, Indices, and All).

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: Removed the error variables.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: I have added the line.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: I have renamed the "execute" functions.

animation_nodes/nodes/mesh/set_vertex_weights.py
3DSinghVFX: I have moved the check-conditions into the `getVertexGroup`.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Added missing nodes to the node menu
 - Added *Collection Operations* node.
 - Added *Rotations* output to the *Particles Data* node.
+- Added *Set Vertex Weights* node.
```
OmarEmaraDev: This should be added in the Unreleased section above.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,124 @@
...
+        object.data.update()    
+        return object
+
+    def execute_Indices_WeightsList(self, object, identifier, indices, weights):
+        if object is None: return
+        vertexGroup = self.getVertexGroup(object, identifier)
+
+        weights = VirtualDoubleList.create(weights, 0).materialize(len(indices))
+        for i, index in enumerate(indices):
+            vertexGroup.add([index], weights[i], "REPLACE")
```
OmarEmaraDev: I allowed python access in ab7992d, so you don't need to materialize anymore.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,124 @@
...
+
+        vertexGroup.add(indices, weight, "REPLACE")
+        object.data.update()    
+        return object
+
+    def execute_All_WeightsList(self, object, identifier, weights):
+        if object is None: return
+        vertexGroup = self.getVertexGroup(object, identifier)
+
+        weights = VirtualDoubleList.create(weights, 0).materialize(len(object.data.vertices))
```
OmarEmaraDev: No need to materialize anymore.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def getVertexGroup(self, object, identifier):
+        if object.type != "MESH": 
+            self.raiseErrorMessage("No mesh object.")
+        
+        if object.mode == "EDIT":
+            self.raiseErrorMessage("Object is not in object mode.")
+
+        try: return object.vertex_groups[identifier]
+        except: return self.raiseErrorMessage("Group is not found.")
```
OmarEmaraDev: No need to return in this case.

animation_nodes/nodes/mesh/set_vertex_weights.py
```diff
@@ -0,0 +1,124 @@
...
+    ("ALL", "All", "Set weight of every vertex", "NONE", 0),
+    ("INDEX", "Index", "Set weight of a specific vertex", "NONE", 1)
+]
+
+groupIdentifierTypeItems = [
+    ("INDEX", "Index", "Get vertex group based on the index", "NONE", 0),
+    ("NAME", "Name", "Get vertex group based on the name", "NONE", 1)
+]
+
+class VertexWeights(bpy.types.Node, AnimationNode):
```
OmarEmaraDev: Rename the class to `SetVertexWeight`, making `weight` singular in the file name and the `bl_` properties along the way.

3DSinghVFX: I have renamed it something like this:
```
class SetVertexWeightNode(bpy.types.Node, AnimationNode):
    bl_idname = "an_SetVertexWeightNode"
    bl_label = "Set Vertex Weight"
```
does it fine ?

OmarEmaraDev: Yes.

CHANGELOG.md
OmarEmaraDev: @3DSinghVFX Have you seen this comment?

OmarEmaraDev:
Thanks for contributing!

PR #1229, opened 2019-11-11
Vertex Color Input and Set Vertex Colors nodes

3DSinghVFX:
I have added the Vertex Color Input and Set Vertex Colors nodes.

**Vertex Color Input Node** allows us to get vertex-color with animation nodes. This node is available under: Add (node-menu) -> Mesh -> Vertex Color Input.

Inputs of this nodes are:
Object <-- Mesh Object
Color Index <-- 0, 1, 2, etc.
Index <-- 0, 1, 2, etc.
Vertex_Color_Input

**Set Vertex Colors Node** allows us to set or control the vertex colors with animation nodes. This node is available under: Add (node-menu) -> Mesh -> Set Vertex Colors.

Inputs of this nodes are:
Object <-- Mesh Object
Color Index <-- 0, 1, 2, etc.
Index <-- 0, 1, 2, etc.
Color/Colors <-- color or color list (length is equal to the number of vertices locations)
Screenshot from 2019-11-12 02-08-07


OmarEmaraDev:
The last time we discussed adding those nodes, there were concerns about the performance due to the fact that Color Lists in Animation Nodes are not CLists. I will talk to Jacques about it and get back to you.

3DSinghVFX:
Okay.

OmarEmaraDev:
@3DSinghVFX Alright. I will first add some data structures to Animation Nodes, then you will have to update your patch to use those data structures. I will give you more details later.

Might take sometime because I have midterm exams right now.

3DSinghVFX:
Okay, thanks.

3DSinghVFX:
Omar, In Set Vertex Color node, I have used `from itertools import chain` function to generate a flat list from the nested lists because it is needed for vertex color. It would be nice if the new data structure will have this functionality inbuild and faster than this method.

OmarEmaraDev:
Yes, I am aware. It is contiguous anyway.

3DSinghVFX:
In the last point, loops mode is fine, I can do that. But I do not know how to handle vertices or polygons, for example in the case of a cube object, 
```
loops = 24  
vertices = 8
polygons = 6
```
but the `vertex_colors['Col'].data.foreach_set()` needs a flat list of length 24*4.

3DSinghVFX:
Here is a function which I have defined to set vertex color per polygon:
```
def setVertexColorForPolygons(self, object, identifier, colors):
        if object is None: return
        colorLayer = self.getVertexColorLayer(object, identifier)
        
        colorsPolygons = [0] * len(colorLayer.data)
        for i, polygon in enumerate(object.data.polygons):
            for index in polygon.loop_indices:
                colorsPolygons[index] = colors[i]

        colorLayer.data.foreach_set("color", colorsFlatList(self, colorsPolygons))
        object.data.update()
        return object
```
Is it the correct way?

OmarEmaraDev:
Pseudocode code to generate `loopColors` from `verticesColors`:

```python
loopColors = []
for polygon in polygons:
    for index in polygon.indices:
        loopsColors.append(verticesColors[index])
```

Pseudocode code to generate `loopColors` from `polygonColors`:

```python
loopColors = []
for i, polygon in enumerate(polygons):
    loopsColors.extend(list(polygonColors[i]) * len(polygon.indices))
```

OmarEmaraDev:
Notice that you will eventually rewrite the code in Cython using the data from the `getPolygonIndices` function, so make sure to try to understand how those works.

3DSinghVFX:
Thanks for the examples. I have modified the code of Set Vertex Color node based on your suggestions, and also refined the code of Vertex Color Input node.

3DSinghVFX:
Omar, Is there a way to generate a flat color list with the help of VirtualColorList or CList?

OmarEmaraDev:
A CList have methods `asMemoryView` and `asNumpyArray`, you can use either to set the colors using the `foreach_set` method.

3DSinghVFX:
Thanks, it works. 

1. I have defined a function (below) to set colors for vertices. It works fine but when colors /= vertices then it brakes the node-tree,

```
def setVertexColorForVertices(self, object, identifier, colors):
colorLayer = self.getVertexColorLayer(object, identifier)

colorLayer.data.foreach_set("color", colors.asNumpyArray())

object.data.update()
return object
```
Should I add the If-condition like `if len(colors) < len(colorLayer.data): return object` above the `foreach_set` to avoid the error?

2. Is it possible to convert a Flat Colors list into a normal Colors list?

OmarEmaraDev:
That's why you should use a virtual list. Create a virtual list from the color list, materialize it, then write it.

I assume you want to create a float list from a memory view, in this case, you can use the `fromNumpyArray` method.

3DSinghVFX:
Okay. To implement that in Cython, should I write `c_utils.pyx` file or write the Set Vertex Color node in Cython?

The other point is, `setVertexColorForVertices` function does not have any **for loop**. However, `setVertexColorForLoops` has a **for loop**. So, I think this function should be in Cython.

OmarEmaraDev:
It seems you switched the implementation of the Vertex and Loops functions. Read my comment above again.

You should add a `c_utils.pyx` file.

3DSinghVFX:
I thought Vertices mode will have direct colors and Loop mode will have loop-colors. But anyway, I'll fix it.  

For `c_utils.pyx` file, I have written this function,
```
def loopsColorsList(loops, ColorList colors):
    
    cdef VirtualColorList colorsList = VirtualColorList.create(colors, 0) 
    cdef VirtualColorList loopsColors = ColorList(length = len(loops)) 
    for i in range(len(loops)):
        loopsColors[i] = colorsList[loops[i].vertex_index]
    return loopsColors
```
Is it correct?

OmarEmaraDev:
When writing Vertex Colors, you are writing to loops not vertices. That's why you have to do more work to write to vertices than to write to loops.

The function should take two arguments of type `PolygonIndicesList` and `VirtualColorList`. In other words, you can get the polygon indices using the `getPolygonIndices` method of meshes. The Virtual list should be created in python not Cython.

3DSinghVFX:
I have fixed the code of `setVertexColorForLoops` and `setVertexColorForVertices` functions.

3DSinghVFX:
> Pseudocode code to generate `loopColors` from `colorsList = verticesColors`:
> 
> ```python
> loopColors = []
> for polygon in polygons:
>     for index in polygon.indices:
>         loopsColors.append(colorsList[index])
> ```
Screenshot from 2019-11-25 12-47-43


and the code which I have written for loops (for setVertexColorForVertices) gives different results,

```
loops = object.data.loops
loopsColors = []
for loop in loops:
    loopsColors.extend(colorsList[loop.vertex_index])
```
Screenshot from 2019-11-25 12-47-30

OmarEmaraDev:
I would expect them to be different. Why do you want them to be identical?

3DSinghVFX:
I do not want them to be identical. My question is, which code is correct for loopsColors for setVertexColorForVertices?

Here is the mochup of the codes for c_utils.pyx file:
```
# cython: profile=True
cimport cython

from ... data_structures ColorList, VirtualColorList

# LoopsColors for Vertices
###########################################
def loopsColorsList(loops, colors):
    
    colorsList = VirtualColorList.create(colors, 0) 
    loopsColors = ColorList(length = len(loops))
     
    for i in range(len(loops)):
        loopsColors[i] = colorsList[loops[i].vertex_index]
    return loopsColors

# PolygonsColors for Vertices
###########################################
def polygonsColorsList(polygons, colors):

    colorsList = VirtualColorList.create(colors, 0)

    polygonsColors = []
    for i, polygon in enumerate(object.data.polygons):
        polygonsColors.extend(list(colorsList[i]) * len(polygon.loop_indices))
    return polygonsColors
```

OmarEmaraDev:
For vertices, your second image is the correct one.

3DSinghVFX:
Omar, I have refined the code for `c_utils.pyx` file and it is working,

```
# cython: profile=True
cimport cython

from ... data_structures import ColorList, VirtualColorList

def loopsColorsList(polygons, colors):
    
    colorsList = VirtualColorList.create(colors, 0) 
    loops = polygons.data.loops
    loopsColors = ColorList(length = len(loops))
     
    for i in range(len(loops)):
        loopsColors[i] = colorsList[loops[i].vertex_index]
    return loopsColors
    
def polygonsColorsList(polygons, colors):

    colorsList = VirtualColorList.create(colors, 0)
    polygonsColors = ColorList(length = len(polygons.data.loops))
    
    for i, polygon in enumerate(polygons):
        for index in polygon.loop_indices:
            polygonsColors[index] = colorsList[i]
    return polygonsColors
```
and in `set_vertex_color.py`, it looks like this for the set vertices colors case, 
```
def setVertexColorForVertices(self, object, identifier, colors):
        colorLayer = self.getVertexColorLayer(object, identifier)
        
        if len(colors) == 0: return object

        polygons = object.data.polygons
        
        colorLayer.data.foreach_set("color", loopsColorsList(polygons, colors).asNumpyArray())
        object.data.update()
        return object
```
Is it fine or needs some improvement?

3DSinghVFX:
> For vertices, your second image is the correct one.

Okay.

OmarEmaraDev:
One of the reasons we use Cython is typing, so don't leave your variables without a type. Including the arguments of the function. The overhead is too high right now. Read my previous comments again https://github.com/JacquesLucke/animation_nodes/pull/1229#issuecomment-557921594.

3DSinghVFX:
Okay, Omar. I have tried to improve the cython code. Here is the code for loops colors:
```
def loopsColorsList(PolygonIndicesList polygons, colors):
    cdef long i
    cdef long Amount = len(polygons.indices)
    
    colorsList = VirtualColorList.create(colors, 0)
    loopsColors = ColorList(length = Amount)
    
    for i in range(Amount):
        loopsColors[i] = colors[polygons.indices[i]]
    return loopsColors
```

OmarEmaraDev:
This looks better, but `colors`, `colorsList`, and `loopsColors` are still not typed. Create the virtual list in python and pass it to the Cython function.

3DSinghVFX:
> Create the virtual list in python and pass it to the Cython function.

Sorry, Omar. I do not know how to do that. Can you please provide an example or node name in which this is done?

OmarEmaraDev:
See any of the geometry nodes. For instance, the virtual lists are created in [python](https://github.com/JacquesLucke/animation_nodes/blob/master/animation_nodes/nodes/geometry/intersect_line_sphere.py#L65), then they are passed to the [cython function](https://github.com/JacquesLucke/animation_nodes/blob/master/animation_nodes/nodes/geometry/c_utils.pyx#L194-L198). Also take a look at how the cython functions are implemented, take notes and try to apply in your functions.

3DSinghVFX:
Thanks for the suggestions. 

Here is the Cython code for loops colors,
```
# cython: profile=True
cimport cython

from ... data_structures cimport (
    PolygonIndicesList
)

def loopsColorsList(long Amount, PolygonIndicesList polygons, colors, loopsColors):
    cdef long i

    for i in range(Amount):
        loopsColors[i] = colors[polygons.indices[i]]
    return loopsColors
```
This is the python function in which I have created the virtual list and passes to the cython function, 
```
def setVertexColorForVertices(self, object, identifier, colors):
        colorLayer = self.getVertexColorLayer(object, identifier)
        
        if len(colors) == 0: return object

        sourceMesh = object.an.getMesh(False)
        polygonIndices = sourceMesh.an.getPolygonIndices()
        Amount = len(polygonIndices.indices)
        
        colorsList = VirtualColorList.create(colors, 0)
        loopsColors = ColorList(length = Amount)
    
        verticesColors = loopsColorsList(Amount, polygonIndices, colorsList, loopsColors)  
        colorLayer.data.foreach_set("color", verticesColors.asNumpyArray())
        object.data.update()
        return object
```

What is the type for 'colorsList' or `loopsColors` in Cython code? I tried a `VirtualColorList` as type but it did not work, gave an error.

OmarEmaraDev:
The type of color lists is `ColorList`. Also, I feel like you are still missing the point of the function. The function takes vertices colors and computes loops colors, not the other way around. "Vertex Colors" are actually not colors of vertices, they are colors of loops.

3DSinghVFX:
1. When I tried to `ColorList`, it gave an error,
```
from ... data_structures cimport (
    ColorList,
    PolygonIndicesList
)
```
Screenshot from 2019-11-27 00-30-08


2. My bad, I forgot to change the names in ` setVertexColorForVertices` function, 
```
def setVertexColorForVertices(self, object, identifier, colors):
        colorLayer = self.getVertexColorLayer(object, identifier)
        
        if len(colors) == 0: return object

        sourceMesh = object.an.getMesh(False)
        polygonIndices = sourceMesh.an.getPolygonIndices()
        Amount = len(polygonIndices.indices)
        
        verticesColors = VirtualColorList.create(colors, 0)
        loopsColors = ColorList(length = Amount)

        colorsList = loopsColorsList(Amount, polygonIndices, verticesColors, loopsColors)  
        colorLayer.data.foreach_set("color", colorsList.asNumpyArray())
        object.data.update()
        return object
```

OmarEmaraDev:
Can you make a clean recompile? Its seems your `base_lists.pyx` don't include the new ColorList structure.

3DSinghVFX:
Thanks, it worked. 

Here is the Cython code with all typed variables,
```
# cython: profile=True
cimport cython

from ... data_structures cimport (
    ColorList,
    VirtualColorList,
    PolygonIndicesList
)

def loopsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors, 
                    ColorList loopsColors):
    cdef long i
    for i in range(Amount):
        loopsColors[i] = colors[polygons.indices[i]]
    return loopsColors
    
def polygonsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors,
                       ColorList polygonsColors):
    cdef long i, j, index
    cdef long polygonsAmount = len(polygons)
    index = 0    
    for i in range(polygonsAmount):
        for j in range(len(polygons[i])):
            polygonsColors[index] = colors[i]
            index += 1
    return polygonsColors
```

OmarEmaraDev:
Great! Just commit it and I will add a review.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
+# cython: profile=True
+cimport cython
+
+from ... data_structures cimport (
+    ColorList,
+    VirtualColorList,
+    PolygonIndicesList
+)
+
+def loopsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors,
```
OmarEmaraDev: Use a more descriptive name, `getLoopColorsFromVertexColors`. The same for the other function.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
...
+cimport cython
+
+from ... data_structures cimport (
+    ColorList,
+    VirtualColorList,
+    PolygonIndicesList
+)
+
+def loopsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors, 
+                    ColorList loopsColors):
```
OmarEmaraDev: Don't pass the output list, just create it in Cython and return it.

animation_nodes/nodes/color/c_utils.pyx
OmarEmaraDev: No need to pass the amount, you can just get it from the polygons list.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
...
+    cdef long i
+    for i in range(Amount):
+        loopsColors[i] = colors[polygons.indices[i]]
```
OmarEmaraDev: Use the `get` method of VirtualList.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
...
+    return loopsColors
+    
+def polygonsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors,
+                       ColorList polygonsColors):
+    cdef long i, j, index
+    cdef long polygonsAmount = len(polygons)
```
OmarEmaraDev: Since it is only used once, just inline it.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
...
+    index = 0    
+    for i in range(polygonsAmount):
+        for j in range(len(polygons[i])):
+            polygonsColors[index] = colors[i]
+            index += 1
```
OmarEmaraDev: I think you can just use the `polyLengths` list of polygons in this case.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
...
+
+def loopsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors, 
+                    ColorList loopsColors):
+    cdef long i
+    for i in range(Amount):
+        loopsColors[i] = colors[polygons.indices[i]]
+    return loopsColors
+    
+def polygonsColorsList(long Amount, PolygonIndicesList polygons, VirtualColorList colors,
+                       ColorList polygonsColors):
```
OmarEmaraDev: Same as above.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
...
+    cdef long i, j, index
+    cdef long polygonsAmount = len(polygons)
+    index = 0    
+    for i in range(polygonsAmount):
+        for j in range(len(polygons[i])):
+            polygonsColors[index] = colors[i]
+            index += 1
+    return polygonsColors
+
```
OmarEmaraDev: New line at the end of the file.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,27 @@
+# cython: profile=True
+cimport cython
```
OmarEmaraDev: Why do we need cython?

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Omar, I tired something like this `colors.get(polygons.indices[i])`. It gave an error `animation_nodes/nodes/color/c_utils.pyx:14:35: Cannot convert 'Color *' to Python object`.

Is there a way to find all the possible attributes of any object (e.g., possible attributes of Virtual List)? Maybe I'm wrong but the `dir()` function does not show any `get` attribute of Virtual Color List.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Now, it looks like this,
```
def getLoopColorsFromPolygonColors(PolygonIndicesList polygons, VirtualColorList colors):
    cdef long i, j, index
    cdef ColorList loopsColors = ColorList(length = len(polygons.indices))
    
    index = 0 
    for i in range(len(polygons)):
        for j in range(polygons.polyLengths[i]):
            loopsColors[index] = colors[i]
            index += 1
    return loopsColors
```

animation_nodes/nodes/color/c_utils.pyx
OmarEmaraDev: `get` is a c method, you wouldn't see it with `dir` because it is not callable from Python. The error has nothing to do with the `get` method. It is probably caused by a wrong assignment, since `loopsColors[i]` is a python object. To assign to a `ColorList`, do `loopsColors.data[0] = colors.get(something)`.

animation_nodes/nodes/color/c_utils.pyx
OmarEmaraDev: Looks correct.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Here is the complete code,
```
from ... data_structures cimport (
    ColorList,
    VirtualColorList,
    PolygonIndicesList
)

def getLoopColorsFromVertexColors(PolygonIndicesList polygons, VirtualColorList colors):
    cdef long i
    cdef Amount = len(polygons.indices)
    cdef ColorList loopsColors = ColorList(length = Amount)

    for i in range(Amount):
        loopsColors.data[i] = colors.get(polygons.indices[i])
    return loopsColors
```
Stil, I'm getting the error,
Screenshot from 2019-11-27 17-42-23

I have also tried this syntax but no success,
`loopsColors[i] = colors.get(polygons.indices[i])`

OmarEmaraDev: Just dereference the right hand side. Dereferencing in Cython is done by `[0]`. For instance : `colors.get(polygons.indices[i])[0]`.

3DSinghVFX: Thanks, it worked. Should I also use this in second function like this `loopsColors.data[index] = colors.get(i)[0]`?

OmarEmaraDev: Yes.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Removed it.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: A new line is added.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: It is refined.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Replaced it with the inline code.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Removed the Amount from the inputs.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Now, it is in the Cython code.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: I have renamed both of the functions.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,129 @@
...
+            return "execute_ColorsList"
+        else:
+            return "execute_SingleColor"
+
+    def execute_SingleColor(self, object, identifier, color):
+        if object is None: return
+        colorLayer = self.getVertexColorLayer(object, identifier)
 
-        mesh = object.data
+        colorsList = VirtualColorList.create(color, 0).materialize(len(colorLayer.data))
```
OmarEmaraDev: The second argument of `create` should be a color not a number.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,129 @@
...
 
-        newColorAsArray = numpy.array(newColor, dtype = "f")
-        colors = numpy.tile(newColorAsArray, len(colorLayer.data))
-        colorLayer.data.foreach_set("color", colors)
+        elif self.mode == "POLYGONS":
+            return self.setVertexColorForPolygons(object, identifier, colors)
 
-        # without this line the 3d view doesn't update
-        colorLayer.data[0].color = newColor
+    def setVertexColorForLoops(self, object, identifier, colors):
```
OmarEmaraDev: You have a lot of repeated code here. I would say just implement them in a single function.

animation_nodes/nodes/color/vertex_color_input.py
```diff
@@ -0,0 +1,101 @@
...
+    ("ALL", "All", "Get color of every vertex", "NONE", 0),
+    ("INDEX", "Index", "Get color of a specific vertex", "NONE", 1)
+]
+
+colorLayerIdentifierTypeItems = [
+    ("INDEX", "Index", "Get color layer based on the index", "NONE", 0),
+    ("NAME", "Name", "Get color layer based on the name", "NONE", 1)
+]
+
+class VertexColorInputNode(bpy.types.Node, AnimationNode):
```
OmarEmaraDev: This should behave similar to the other node. There should be an anum to choose the mode (Loop, Vertex, or Polygon). The Vertex and Polygon outputs should be the average of the corresponding loops.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: Do you mean something like this,
```
def execute_ColorsList(self, object, identifier, colors):
        if object is None: return

        colorLayer = self.getVertexColorLayer(object, identifier)
        
        if self.mode == "LOOPS":
            colorsList = VirtualColorList.create(colors, 0).materialize(len(colorLayer.data))
            
        elif self.mode == "VERTICES":
            sourceMesh = object.an.getMesh(False)
            polygonIndices = sourceMesh.an.getPolygonIndices()
        
            verticesColors = VirtualColorList.create(colors, 0)
            colorsList = getLoopColorsFromVertexColors(polygonIndices, verticesColors) 

        elif self.mode == "POLYGONS":
            sourceMesh = object.an.getMesh(False)
            polygonIndices = sourceMesh.an.getPolygonIndices()
        
            polygonsColors = VirtualColorList.create(colors, 0)
            colorsList = getLoopColorsFromPolygonColors(polygonIndices, polygonsColors)
        
        colorLayer.data.foreach_set("color", colorsList.asNumpyArray())
        object.data.update()
        return object
```

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: Is this correct way?
```
color = Color((0., 0., 0., 1.))
colorsList = VirtualColorList.create(colors, color).materialize(len(colorLayer.data))
```

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,26 @@
+from ... data_structures cimport (
+    ColorList,
+    VirtualColorList,
+    PolygonIndicesList
+)
+
+def getLoopColorsFromVertexColors(PolygonIndicesList polygons, VirtualColorList colors):
+    cdef long i
```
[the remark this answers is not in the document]
3DSinghVFX: Is it correct?
```
def getLoopColorsFromVertexColors(PolygonIndicesList polygons, VirtualColorList colors):
    cdef loopsCount = len(polygons.indices)
    cdef ColorList loopsColors = ColorList(length = loopsCount)

    cdef long i
    for i in range(loopsCount):
        loopsColors.data[i] = colors.get(polygons.indices[i])[0]
    return loopsColors
```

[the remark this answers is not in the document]
OmarEmaraDev: Yes.

animation_nodes/nodes/color/set_vertex_color.py
OmarEmaraDev: Yes. But name the variable `defaultColor` and lose the point in the arguments, Python doesn't really care.

animation_nodes/nodes/color/set_vertex_color.py
OmarEmaraDev: Yes, but you can also move the VirtualColorList creation outside of the if statement.

3DSinghVFX: Now, it looks like this,
```
def execute_ColorsList(self, object, identifier, colors):
        if object is None: return
        
        defaultColor = Color((0, 0, 0, 1))
        colorsList = VirtualColorList.create(colors, defaultColor)
        colorLayer = self.getVertexColorLayer(object, identifier)
        
        if self.mode == "LOOP":
            colorsList = colorsList.materialize(len(colorLayer.data))
            
        elif self.mode == "VERTEX":
            sourceMesh = object.an.getMesh(False)
            polygonIndices = sourceMesh.an.getPolygonIndices()

            colorsList = getLoopColorsFromVertexColors(polygonIndices, colorsList) 

        elif self.mode == "POLYGON":
            sourceMesh = object.an.getMesh(False)
            polygonIndices = sourceMesh.an.getPolygonIndices()
        
            colorsList = getLoopColorsFromPolygonColors(polygonIndices, colorsList)
        
        colorLayer.data.foreach_set("color", colorsList.asNumpyArray())
        object.data.update()
        return object
```

3DSinghVFX: I have also changed the name of modes,
Loops -> Loop
Vertices -> Vertex
Polygons -> Polygon

OmarEmaraDev: Looks good, but maybe remove the newline before the elifs.

animation_nodes/nodes/color/vertex_color_input.py
3DSinghVFX: Omar, I did not understand the _Vertex outputs should be the average of the loops_?

OmarEmaraDev: When you are in Vertex mode, multiple loops will share the same vertex. In which case, the output color for each vertex should be the average of all vertices that share it. The same goes for polygons.

3DSinghVFX: Here is a code to get average colors for polygons,
```
sourceMesh = object.an.getMesh(False)
polygonIndices = sourceMesh.an.getPolygonIndices()

colorsList = self.getVertexColors(object, identifier)
polygonsColors = ColorList(length = len(polygonIndices))

index = 0 
for i in range(len(polygonIndices)):
    r = 0
    g = 0
    b = 0
    a = 0
    polyLength = polygonIndices.polyLengths[i]
    for j in range(polyLength):
        color = colorsList[index]
        r += color.r/polyLength
        b += color.b/polyLength
        g += color.g/polyLength
        a += color.a/polyLength
        index += 1
    polygonsColors[i] = Color((r, g, b, a))
```

OmarEmaraDev: Sure. But it should also be implemented in Cython. You can add utility functions in `math/Color.pyx` if you need.

3DSinghVFX: Yes, after finalizing it, I'll implement in the Cython. I think it would be convenient if we have simple math operations for colors e.g., `colorA = ColorB + ColorC` or 'ColorA = ColorB/(any number)'.

Here is code for Vertex average,
```
sourceMesh = object.an.getMesh(False)
polygonIndices = sourceMesh.an.getPolygonIndices()
vertexCount = len(sourceMesh.vertices)

colorsList = self.getVertexColors(object, identifier)
vertexsColors = ColorList(length = vertexCount)

for i in range(vertexCount):
    r = 0
    g = 0
    b = 0
    a = 0
    shareCount = 0
    for j in range(len(polygonIndices.indices)):
        polyIndex = polygonIndices.indices[j]
        if polyIndex == i:
            color = colorsList[j]
            r += color.r
            b += color.b
            g += color.g
            a += color.a
            shareCount += 1
    vertexsColors[i] = Color((r/shareCount, g/shareCount, b/shareCount, a/shareCount))
return vertexsColors
```

3DSinghVFX: Omar, these are the functions for Cython code,

For vertex colors from loops colors:
```
def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
    cdef Color color
    cdef double r, g, b, a
    cdef long i, j, shareCount
    cdef ColorList vertexsColors = ColorList(length = vertexCount)
            
    for i in range(vertexCount):
        r = 0
        g = 0
        b = 0
        a = 0
        shareCount = 0
        for j in range(len(polygons.indices)):
            if polygons.indices[j] == i:
                color = colors.get(polygons.indices[j])[0]
                r += color.r
                b += color.b
                g += color.g
                a += color.a
                shareCount += 1
        vertexsColors.data[i] = Color(r/shareCount, g/shareCount, b/shareCount, a/shareCount)
    return vertexsColors

```
For polygons colors from loops colors:
```
def getPolygonColorsFromLoopColors(PolygonIndicesList polygons, VirtualColorList colors):
    cdef Color color
    cdef double r, g, b, a
    cdef long i, j, index, polyLength
    cdef polygonsCount = len(polygons)
    cdef ColorList polygonsColors = ColorList(length = polygonsCount)
    
    index = 0 
    for i in range(polygonsCount):
        r = 0
        g = 0
        b = 0
        a = 0
        polyLength = polygons.polyLengths[i]
        for j in range(polyLength):
            color = colors.get(polygons.indices[index])[0]
            r += color.r/polyLength
            b += color.b/polyLength
            g += color.g/polyLength
            a += color.a/polyLength
            index += 1
        polygonsColors.data[i] = Color(r, g, b, a)
    return polygonsColors
```

OmarEmaraDev: The `getVertexColorsFromLoopColors` function looks inefficient. Maybe something like this would be more efficient:
```
verticesColors = a list of zero colors of length verticesCount
loopCounts = a list of zero integers of length verticesCount
for i in indices:
    verticesColors[i] += loopColors[i]
    loopCounts[i] += 1

for i in range(verticesCount):
    verticesColors[i] /= loopCounts[i]

return verticesColors
```

3DSinghVFX: Omar, How I can create ColorList with zero color `vertexsColors = ColorList(length = vertexCount)`?

3DSinghVFX: Ahh, I found the syntax to create  ColorList with zero color `vertexsColors = ColorList.fromValue(defaultColor, length = vertexCount)`, is it correct way?

3DSinghVFX: Here is code of `getVertexColorsFromLoopColors` function for Cython,
```
def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
    cdef long i, index
    cdef double r, g, b, a
    cdef Color loopColor, vertexColor
    cdef ColorList vertexsColors = ColorList.fromValue((0, 0, 0, 0), length = vertexCount)
    cdef LongList loopCounts = LongList.fromValue(0, length = vertexCount)        
    
    for i in range(len(polygons.indices)):
        index = polygons.indices[i]
        vertexColor = vertexsColors.data[index]
        loopColor = colors.get(polygons.indices[i])[0]
        vertexColor.r += loopColor.r
        vertexColor.b += loopColor.b
        vertexColor.g += loopColor.g
        vertexColor.a += loopColor.a
        vertexsColors.data[index] = vertexColor
        loopCounts[index] += 1

    for i in range(vertexCount):
        vertexColor = vertexsColors.data[i]
        vertexColor.r = vertexColor.r/loopCounts[i]
        vertexColor.b = vertexColor.b/loopCounts[i]
        vertexColor.g = vertexColor.g/loopCounts[i]
        vertexColor.a = vertexColor.a/loopCounts[i]
        vertexsColors.data[i] = vertexColor
    return vertexsColors
```

OmarEmaraDev: Nice. Now add some utility functions in `math/color.pyx` to handle color addition and division.

3DSinghVFX: Okay. I have a question,
- What are the basic points I have to take into account while adding any utility function?

Here is the mockup of the code, 
```
from . color cimport Color

cdef addColor(Color colorA, Color colorB):
    cdef double r, g, b, a
    r = colorA.r + colorB.r
    g = colorA.g + colorB.g
    b = colorA.b + colorB.b
    a = colorA.a + colorB.a
    return Color(r, g, b, a)

cdef divideColor(Color color, double x):
    color.r = color.r/x
    color.g = color.g/x
    color.b = color.b/x
    color.a = color.a/x
    return color
```
It compiled but gave a warning `math/color.c:1154:18: warning: ‘__pyx_f_15animation_nodes_4math_5color_divideColor’ defined but not used [-Wunused-function]`

OmarEmaraDev: - Notice that there is a `color.pxd` file. This is like a header file, where you put definitions and function prototypes.
- Functions that operates on structs (Like Color), should take pointers to structs as arguments, including the output. Then you should write the output to that color instead of returning it. For instance:
  ```cython
  cdef void addColor(Color* target, Color* a, Color* b):
    target.r = a.r + b.r
    target.g = a.g + b.g
    target.b = a.b + b.b
    target.a = a.a + b.a
  ```
- Some functions are said to operate "Inplace". Those functions write the computed values to the input directly. For instance:
  ```cython
  cdef void scaleColor_Inplace(Color* c, float factor):
    c.r *= factor
    c.g *= factor
    c.b *= factor
    c.a *= factor
  ```

The warning you attached will be gone as soon as you use the function.

3DSinghVFX: Thanks for the explanation. Is this how they will be defined in `color.pxd` file,
```
from . color cimport Color
cdef void addColor(Color target, Color a, Color b)
cdef void scaleColor_Inplace(Color c, float factor)
```

3DSinghVFX: Okay. I have found the correct way,
```
cdef struct Color:
    float r, g, b, a

cdef void addColor(Color* target, Color* a, Color* b)
cdef void scaleColor_Inplace(Color* c, float factor)
```
and then I call in the `color/c_utils.pyx` file,
```
from ... math cimport (
    addColor
)
```
But it gives an error,
```
File "animation_nodes/nodes/color/c_utils.pyx", line 1, in init animation_nodes.nodes.color.c_utils
AttributeError: module 'animation_nodes.math.color' has no attribute '__pyx_capi__'

```

OmarEmaraDev: It is not immediately clear to me why this error occurs. Perhaps the code didn't compile correctly. Can you recompile?

3DSinghVFX: There is no error while compiling but when I run the blender and add the Vertex Input Node (have an object with a Col layer) then that error shows up. Here is the full screenshot, 
Screenshot from 2019-11-30 00-31-14

OmarEmaraDev: Can you commit the changes so that I can investigate this better?

3DSinghVFX: My bad. I just found the cause. Actually, I forgot to copy the newly compiled `color.cpython-37m-x86_64-linux-gnu.so` in `animation_nodes/math/` folder (using in the blender).

OmarEmaraDev: You are not using the `--copy` flag?

3DSinghVFX: Now, I'm using that flag.

Omar, this function, can be used for `multiply` as well as `divide` color with a number,
```
cdef void scaleColor_Inplace(Color* c, float factor):
  c.r *= factor
  c.g *= factor
  c.b *= factor
  c.a *= factor
```
Questions:
- Should we need to create separate functions for `multiply` and `divide` operations or use the above function?
- If we are going to keep the above function then what will be its name, `scaleColor_Inplace` or `scaleColor` or `multipy_divide_Color` or sth else?

OmarEmaraDev: Keep the name. Use it for division.

3DSinghVFX: Here are functions of `c_utils.pyx` file with `addColor` and `scaleColor_Inplace`,
```
def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
    cdef long i, index
    cdef Color vertexColor
    cdef ColorList vertexsColors = ColorList.fromValue((0, 0, 0, 0), length = vertexCount)
    cdef LongList loopCounts = LongList.fromValue(0, length = vertexCount)        
    
    for i in range(len(polygons.indices)):
        index = polygons.indices[i]
        vertexColor = vertexsColors.data[index]
        addColor(&vertexColor, &vertexColor, &colors.get(polygons.indices[i])[0])
        vertexsColors.data[index] = vertexColor
        loopCounts[index] += 1

    for i in range(vertexCount):
        scaleColor_Inplace(&vertexsColors.data[i], 1/loopCounts[i])
    return vertexsColors
```
and 

```
def getPolygonColorsFromLoopColors(PolygonIndicesList polygons, VirtualColorList colors):
    cdef Color color
    cdef long i, j, index, polyLength
    cdef polygonsCount = len(polygons)
    cdef ColorList polygonsColors = ColorList(length = polygonsCount)
    
    index = 0 
    for i in range(polygonsCount):
        color = Color(0, 0, 0, 0)
        polyLength = polygons.polyLengths[i]
        for j in range(polyLength):
            addColor(&color, &color, &colors.get(polygons.indices[index])[0])
            index += 1
        scaleColor_Inplace(&color, 1/polyLength)
        polygonsColors.data[i] = color
    return polygonsColors
```

OmarEmaraDev: Ok, just commit directly and I will ad a review there.

animation_nodes/nodes/color/vertex_color_input.py
```diff
@@ -0,0 +1,132 @@
+import bpy
+from bpy.props import *
+from ... events import propertyChanged
+from ... base_types import AnimationNode, VectorizedSocket
+from ... data_structures import Color, ColorList, VirtualColorList, LongList
+from . c_utils import getVertexColorsFromLoopColors, getPolygonColorsFromLoopColors
+
+modeItems = [
+    ("ALL", "All", "Get color of every vertex", "NONE", 0),
+    ("INDEX", "Index", "Get color of a specific vertex", "NONE", 1)
```
OmarEmaraDev: Remove the index mode.

animation_nodes/math/color.pyx
```diff
@@ -0,0 +1,15 @@
...
+  target.b = a.b + b.b
+  target.a = a.a + b.a
+
+cdef void scaleColor_Inplace(Color* c, float factor):
+  c.r *= factor
+  c.g *= factor
+  c.b *= factor
+  c.a *= factor
+
+
```
OmarEmaraDev: New line at the end of file.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,66 @@
...
+    
+    for i in range(len(polygons.indices)):
+        index = polygons.indices[i]
+        vertexColor = vertexsColors.data[index]
+        addColor(&vertexColor, &vertexColor, &colors.get(polygons.indices[i])[0])
+        vertexsColors.data[index] = vertexColor
+        loopCounts[index] += 1
+
+    for i in range(vertexCount):
+        scaleColor_Inplace(&vertexsColors.data[i], 1/loopCounts[i])
```
OmarEmaraDev: Spaces around the `/` and `1.0` instead of `1`.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,100 @@
...
+from . c_utils import getLoopColorsFromVertexColors, getLoopColorsFromPolygonColors
+
+colorModeItems = [
+    ("LOOP", "Loop", "Set color of every loop vertex", "NONE", 0),
+    ("VERTEX", "Vertex", "Set color of every vertex", "NONE", 1),
+    ("POLYGON", "Polygon", "Set color of every polgon", "NONE", 2)    
+]
+
+colorLayerIdentifierTypeItems = [
+    ("INDEX", "Index", "Get vertex color layer based on the index", "NONE", 0),
```
OmarEmaraDev: Remove index mode.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,66 @@
...
+    cdef ColorList loopsColors = ColorList(length = len(polygons.indices))
+    
+    index = 0 
+    for i in range(len(polygons)):
+        for j in range(polygons.polyLengths[i]):
+            loopsColors.data[index] = colors.get(i)[0]
+            index += 1
+    return loopsColors
+
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
```
OmarEmaraDev: Use the `@cython.cdivision(True)` directive here.

animation_nodes/nodes/color/vertex_color_input.py
3DSinghVFX: Omar, this `Index` mode is to _Get color of a specific vertex_ but not for _Color Layer_. Which one do you want to remove?

OmarEmaraDev: The node should return a list of colors all the time. So yes, the Index mode is not needed.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: Omar, I think `Index` mode is handier than the `Name` mode to get the `Vertex Color Layer`, is it not? The `Set Vertex Weight` and `Vertex Group Input` nodes also have this mode.

3DSinghVFX:
> I will probably be gone for couple of days. So I will get back to you then.
Okay. In the meantime, I'll try to improve the code based on your suggestions.

animation_nodes/nodes/color/set_vertex_color.py
OmarEmaraDev: Oh, my bad, I confused it with the other mode type. This can be left as is then.

animation_nodes/nodes/color/vertex_color_input.py
OmarEmaraDev: I want to remove the one that "Get color of a specific vertex".

3DSinghVFX: Okay.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: It is fine :)

animation_nodes/nodes/color/vertex_color_input.py
3DSinghVFX: I have removed the `Index` mode and put the remaing code into a `execute` function,
```
def execute(self, object, identifier):
        if object is None: return
        
        colorLayer = self.getVertexColorLayer(object, identifier)
        defaultColor = Color((0, 0, 0, 1))
        
        colorsList = ColorList(length = len(colorLayer.data))
        colorLayer.data.foreach_get("color", colorsList.asNumpyArray())

        if self.colorMode == "LOOP":
            return colorsList

        elif self.colorMode == "VERTEX":
            sourceMesh = object.an.getMesh(False)
            polygonIndices = sourceMesh.an.getPolygonIndices()
            vertexCount = len(sourceMesh.vertices)
            
            colorsList = VirtualColorList.create(colorsList, defaultColor)
            return getVertexColorsFromLoopColors(vertexCount, polygonIndices, colorsList)

        elif self.colorMode == "POLYGON":
            sourceMesh = object.an.getMesh(False)
            polygonIndices = sourceMesh.an.getPolygonIndices()

            colorsList = VirtualColorList.create(colorsList, defaultColor)
            return getPolygonColorsFromLoopColors(polygonIndices, colorsList)
```

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,68 @@
...
+        vertexColor = vertexsColors.data[index]
+        addColor(&vertexColor, &vertexColor, &colors.get(polygons.indices[i])[0])
+        vertexsColors.data[index] = vertexColor
+        loopCounts[index] += 1
+
+    for i in range(vertexCount):
+        scaleColor_Inplace(&vertexsColors.data[i], 1.0 / loopCounts[i])
+    return vertexsColors
+
+def getPolygonColorsFromLoopColors(PolygonIndicesList polygons, VirtualColorList colors):
```
OmarEmaraDev: `@cython.cdivision(True)` here as well.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: It is added.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Replaced the `1` -> `1.0` and `0` -> `0.0`.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: `@cython.cdivision(True)` is added.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    cdef loopsCount = len(polygons.indices)
+    cdef ColorList loopsColors = ColorList(length = loopsCount)
+
+    cdef long i
+    for i in range(loopsCount):
+        loopsColors.data[i] = colors.get(polygons.indices[i])[0]
+    return loopsColors
+    
+def getLoopColorsFromPolygonColors(PolygonIndicesList polygons, VirtualColorList colors):
+    cdef long i, j, index
```
OmarEmaraDev: Move this line just before the for loop. To make declaration and use of the variable closer together.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    index = 0 
+    for i in range(len(polygons)):
+        for j in range(polygons.polyLengths[i]):
+            loopsColors.data[index] = colors.get(i)[0]
+            index += 1
+    return loopsColors
+
+@cython.cdivision(True)
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
+    cdef long i, index
```
OmarEmaraDev: Move this to just before the for loop.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    cdef Color vertexColor
+    cdef ColorList vertexsColors = ColorList.fromValue((0.0, 0.0, 0.0, 0.0), length = vertexCount)
+    cdef LongList loopCounts = LongList.fromValue(0, length = vertexCount)        
+    
+    for i in range(len(polygons.indices)):
+        index = polygons.indices[i]
+        vertexColor = vertexsColors.data[index]
+        addColor(&vertexColor, &vertexColor, &colors.get(polygons.indices[i])[0])
```
OmarEmaraDev: For the third argument, you don't need the `&` and `[0]`.

animation_nodes/ui/node_menu.py
```diff
@@ -368,9 +368,12 @@ def draw(self, context):
         insertNode(layout, "an_ObjectBoundingBoxNode", "Get Bounding Box")
         insertNode(layout, "an_CombineMeshNode", "Combine Mesh")
         insertNode(layout, "an_MeshFromSplineNode", "Mesh From Spline")
-        layout.separator()
+        layout.separator()
```
OmarEmaraDev: Why the extra white space?

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Moved the declaration of variables.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: It is done.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Removed the `&` and `[0]` for the third argument.

animation_nodes/ui/node_menu.py
3DSinghVFX: Removed the extra white space.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    cdef ColorList polygonsColors = ColorList(length = polygonsCount)
+    
+    cdef Color color
+    cdef long i, j, index, polyLength
+    index = 0 
+    for i in range(polygonsCount):
+        color = Color(0.0, 0.0, 0.0, 0.0)
+        polyLength = polygons.polyLengths[i]
+        for j in range(polyLength):
+            addColor(&color, &color, &colors.get(polygons.indices[index])[0])
```
OmarEmaraDev: No need for `&` and `[0]` in the third argument. They do opposite jobs, one takes a pointer and the other dereference, so they cancel each other.

3DSinghVFX: Okay, thanks for the info. I removed `&` and `[0]`.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    return loopsColors
+    
+def getLoopColorsFromPolygonColors(PolygonIndicesList polygons, VirtualColorList colors):
+    cdef ColorList loopsColors = ColorList(length = len(polygons.indices))
+    
+    cdef long i, j, index
+    index = 0 
+    for i in range(len(polygons)):
+        for j in range(polygons.polyLengths[i]):
+            loopsColors.data[index] = colors.get(i)[0]
```
OmarEmaraDev: Get the color in the outer loop and store it in a variable.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+@cython.cdivision(True)
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
+    cdef ColorList vertexsColors = ColorList.fromValue((0.0, 0.0, 0.0, 0.0), length = vertexCount)
+    cdef LongList loopCounts = LongList.fromValue(0, length = vertexCount)        
+    
+    cdef Color vertexColor
+    cdef long i, index
+    for i in range(len(polygons.indices)):
+        index = polygons.indices[i]
+        vertexColor = vertexsColors.data[index]
```
OmarEmaraDev: Perhaps an inplace function can be used here. `addColor_Inplace(vertexsColors.data + index,  colors.get(polygons.indices[i]))`.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+        addColor(&vertexColor, &vertexColor, colors.get(polygons.indices[i]))
+        vertexsColors.data[index] = vertexColor
+        loopCounts[index] += 1
+
+    for i in range(vertexCount):
+        scaleColor_Inplace(&vertexsColors.data[i], 1.0 / loopCounts[i])
```
OmarEmaraDev: Same as before, `&` and `[i]` are doing opposite jobs. You can do `vertexsColors.data + i`, this is called pointer arithmetic.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    cdef long i, j, index, polyLength
+    index = 0 
+    for i in range(polygonsCount):
+        color = Color(0.0, 0.0, 0.0, 0.0)
+        polyLength = polygons.polyLengths[i]
+        for j in range(polyLength):
+            addColor(&color, &color, colors.get(polygons.indices[index]))
+            index += 1
+        scaleColor_Inplace(&color, 1.0 / polyLength)
+        polygonsColors.data[i] = color
```
OmarEmaraDev: Same as above.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: But we haven't added the `addColor_Inplace` function in `color.pyx` file. Should I add that function?

OmarEmaraDev: Yes.

3DSinghVFX: Okay.

3DSinghVFX:
> Also, please add your changes to the CHANGELOG file.

I have mentioned these points in the CHANGELOG file,
```
## Unreleased
- Added Vertex Color Input node.
- Added Set Vertex Color node.
- Added c_utils.pyx for colors.
```
and what else should I mention in this file?

OmarEmaraDev:
@3DSinghVFX You don't need to mention the addition of the `c_utils` file, the CHANGELOG is mostly for changes that affects the user, the user won't see `c_utils`. Also the *Set Vertex Color* node technically got updated, so move it to the Changed section `- Vectorized Set Vertex Color node.` or something.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: It is done.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Removed the `&` and `[i]`.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: It is changed.

CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## Unreleased
+- Added Vertex Color Input node.
```
OmarEmaraDev: Move this to the Added section.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+
+@cython.cdivision(True)
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
+    cdef ColorList vertexsColors = ColorList.fromValue((0.0, 0.0, 0.0, 0.0), length = vertexCount)
+    cdef LongList loopCounts = LongList.fromValue(0, length = vertexCount)        
+    
+    cdef long i, index
+    for i in range(len(polygons.indices)):
+        index = polygons.indices[i]
+        addColor_Inplace(vertexsColors.data + index, colors.get(polygons.indices[i]))
```
OmarEmaraDev: Isn't `polygons.indices[i]` equivalent to `index`?

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+    cdef ColorList polygonsColors = ColorList(length = polygonsCount)
+    
+    cdef Color color
+    cdef long i, j, index, polyLength
+    index = 0 
+    for i in range(polygonsCount):
+        color = Color(0.0, 0.0, 0.0, 0.0)
+        polyLength = polygons.polyLengths[i]
+        for j in range(polyLength):
+            addColor_Inplace(&color, colors.get(polygons.indices[index]))
```
OmarEmaraDev: You can write to the list directly just as you did above. Zero initialize the list first, of course.

3DSinghVFX: Like this,
```
@cython.cdivision(True)
def getPolygonColorsFromLoopColors(PolygonIndicesList polygons, VirtualColorList colors):
    cdef polygonsCount = len(polygons)
    cdef ColorList polygonsColors = ColorList.fromValue((0.0, 0.0, 0.0, 0.0), length = polygonsCount)
    
    cdef long i, j, index, polyLength
    index = 0 
    for i in range(polygonsCount):
        polyLength = polygons.polyLengths[i]
        for j in range(polyLength):
            addColor_Inplace(polygonsColors.data + i, colors.get(polygons.indices[index]))
            index += 1
        scaleColor_Inplace(polygonsColors.data + i, 1.0 / polyLength)
    return polygonsColors
```

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Yes, I have replaced that.

animation_nodes/nodes/color/c_utils.pyx
OmarEmaraDev: Yes.

CHANGELOG.md
3DSinghVFX: It is changed.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,100 @@
...
-        if object is None: return object
-        if object.type != "MESH": return object
-        if object.mode == "EDIT":
-            self.raiseErrorMessage("Object is in edit mode")
-            return object
+        self.newOutput("Object", "Object", "outObject")
 
-        mesh = object.data
+    def draw(self, layout):
+        layout.prop(self, "colorMode", text = "")
```
OmarEmaraDev: Only draw the enum if `useColorList` is true.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,100 @@
...
     def create(self):
-        self.newInput("Object", "Object", "object").defaultDrawType = "PROPERTY_ONLY"
-        self.newInput("Color", "Color", "color").defaultDrawType = "PROPERTY_ONLY"
-        self.newOutput("Object", "Object", "outObject")
+        self.newInput("Object", "Object", "object", defaultDrawType = "PROPERTY_ONLY")
 
-    def draw(self, layout):
-        layout.prop(self, "vertexColorName", text = "", icon = "GROUP_VCOL")
+        if self.colorLayerIdentifierType == "INDEX":
+            self.newInput("Integer", "Color Index", "colorLayerIndex")
```
OmarEmaraDev: Rename this to something more clear. *Vertex Color Index* maybe.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,66 @@
...
+
+@cython.cdivision(True)
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
+    cdef ColorList vertexsColors = ColorList.fromValue((0.0, 0.0, 0.0, 0.0), length = vertexCount)
+    cdef LongList loopCounts = LongList.fromValue(0, length = vertexCount)        
+    
+    cdef long i, index
+    for i in range(len(polygons.indices)):
+        index = polygons.indices[i]
+        addColor_Inplace(vertexsColors.data + index, colors.get(index))
```
OmarEmaraDev: `colors.get(index)` should be `colors.get(i)`, I think?

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,66 @@
...
+    for i in range(len(polygons)):
+        polygonColor = colors.get(i)[0]
+        for j in range(polygons.polyLengths[i]):
+            loopsColors.data[index] = polygonColor
+            index += 1
+    return loopsColors
+
+@cython.cdivision(True)
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
+    cdef ColorList vertexsColors = ColorList.fromValue((0.0, 0.0, 0.0, 0.0), length = vertexCount)
```
OmarEmaraDev: Instead of using `fromValue` to zero initialize the list, you can create the list, then use the fill function, like so:
```
    cdef ColorList vertexColors = ColorList(length = vertexCount)
    cdef LongList loopCounts = LongList(length = vertexCount)        
    vertexColors.fill(0)
    loopCounts.fill(0)
```

Also, `vertexsColors` should be `vertexColors`, there is no `s`.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: It is added.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: Renamed to _Vertex Color Index_.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: Yes, I have corrected that.

animation_nodes/nodes/color/c_utils.pyx
3DSinghVFX: I have changed that.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,101 @@
...
 from ... events import propertyChanged
-from ... base_types import AnimationNode
+from ... base_types import AnimationNode, VectorizedSocket
+from ... data_structures import Color, ColorList, VirtualColorList
+from . c_utils import getLoopColorsFromVertexColors, getLoopColorsFromPolygonColors
+
+colorModeItems = [
+    ("LOOP", "Loop", "Set color of every loop vertex", "NONE", 0),
+    ("VERTEX", "Vertex", "Set color of every vertex", "NONE", 1),
+    ("POLYGON", "Polygon", "Set color of every polgon", "NONE", 2)
```
OmarEmaraDev: Typo in `polygon`.

animation_nodes/nodes/color/set_vertex_color.py
```diff
@@ -1,62 +1,101 @@
...
 
 class SetVertexColorNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_SetVertexColorNode"
     bl_label = "Set Vertex Color"
     errorHandlingType = "EXCEPTION"
 
-    vertexColorName: StringProperty(name = "Vertex Color Group", default = "Col", update = propertyChanged)
-    checkIfColorIsSet: BoolProperty(default = True)
+    colorMode: EnumProperty(name = "Color Mode", default = "LOOP",
+        items = colorModeItems, update = AnimationNode.refresh)
```
OmarEmaraDev: The update function should be `propertyChanged`. `AnimationNode.refresh` will trigger a redraw of the node, which is not needed in this case.

animation_nodes/nodes/color/vertex_color_input.py
```diff
@@ -0,0 +1,84 @@
...
+        self.newOutput("Color List", "Colors", "colors")        
+
+    def draw(self, layout):
+        layout.prop(self, "colorMode", text = "")
+
+    def drawAdvanced(self, layout):
+        layout.prop(self, "colorLayerIdentifierType", text = "Type")
+
+    def execute(self, object, identifier):
+        if object is None: return
```
OmarEmaraDev: You should return an empty ColorList in this case.

animation_nodes/nodes/color/vertex_color_input.py
```diff
@@ -0,0 +1,84 @@
...
+        
+        colorLayer = self.getVertexColorLayer(object, identifier)
+        defaultColor = Color((0, 0, 0, 1))
+        
+        colorsList = ColorList(length = len(colorLayer.data))
+        colorLayer.data.foreach_get("color", colorsList.asNumpyArray())
+
+        if self.colorMode == "LOOP":
+            return colorsList
+
```
OmarEmaraDev: Remove the new line between the `if` and `elif`.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
+import cython
+from ... math cimport (
+    addColor,
```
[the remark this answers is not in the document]
3DSinghVFX: Removed.

animation_nodes/nodes/color/c_utils.pyx
```diff
@@ -0,0 +1,69 @@
...
+            loopsColors.data[index] = polygonColor
+            index += 1
+    return loopsColors
+
+@cython.cdivision(True)
+def getVertexColorsFromLoopColors(long vertexCount, PolygonIndicesList polygons, VirtualColorList colors):
+    cdef ColorList vertexColors = ColorList(length = vertexCount)
+    cdef LongList loopCounts = LongList(length = vertexCount)        
+    loopCounts.fill(0)
+    vertexColors.fill(0.0)
```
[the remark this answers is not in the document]
3DSinghVFX: Okay.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: Corrected in both nodes.

animation_nodes/nodes/color/set_vertex_color.py
3DSinghVFX: Okay. Thanks for the info :) It is changed.

animation_nodes/nodes/color/vertex_color_input.py
3DSinghVFX: Now, it is returning an empty ColorList.

animation_nodes/nodes/color/vertex_color_input.py
3DSinghVFX: Removed the new line.

3DSinghVFX:
Thank you for helping me to understand the different concepts and codes in Animation Nodes.

PR #1256, opened 2019-12-18
Vertex color layer and nodes

3DSinghVFX:
1) This branch has Vertex Color Layer as mesh property like UVs. 
2) Two new nodes, **Get Mesh Vertex Color** and **Add Mesh Vertex Color**.

**Get Mesh Vertex Color**: This node takes mesh and the name of the vertex color layer as inputs. It gives the colors based on the color-mode (Loop, Vertex, Polygon).
Screenshot from 2019-12-18 22-56-35

**Add Mesh Vertex Color**: This node allows you to add a vertex color layer to the mesh. This node takes mesh, vertex color layer name, and color/colors (based on the color-mode e.g., Loop, Vertex, Polygon) as inputs. 
Screenshot from 2019-12-18 22-57-02

3) The Mesh Info, Mesh Object Input, and Mesh Object Output nodes are updated.


animation_nodes/nodes/mesh/add_mesh_vertex_color.py
```diff
@@ -0,0 +1,69 @@
...
+from ... data_structures import Color, ColorList, VirtualColorList
+from .. color.c_utils import getLoopColorsFromVertexColors, getLoopColorsFromPolygonColors
+
+colorModeItems = [
+    ("LOOP", "Loop", "Get color of every loop vertex", "NONE", 0),
+    ("VERTEX", "Vertex", "Get color of every vertex", "NONE", 1),
+    ("POLYGON", "Polygon", "Get color of every polygon", "NONE", 2)    
+]
+
+class AddMeshVertexColorNode(bpy.types.Node, AnimationNode):
```
OmarEmaraDev: Perhaps we should change the name to `InsertVertexColorLayerNode`, to be more uniform and to make it easier to search for. Do you agree?

animation_nodes/nodes/mesh/add_mesh_vertex_color.py
```diff
@@ -0,0 +1,69 @@
...
+            layout.prop(self, "colorMode", text = "")
+
+    def getExecutionFunctionName(self):
+        if self.useColorList:
+            return "execute_ColorsList"
+        else:
+            return "execute_SingleColor"
+
+    def execute_SingleColor(self, mesh, colorLayerName, color):
+        if mesh is None: return
```
OmarEmaraDev: return None

animation_nodes/nodes/mesh/add_mesh_vertex_color.py
```diff
@@ -0,0 +1,69 @@
...
+
+        defaultColor = Color((0, 0, 0, 1))
+        colorsList = VirtualColorList.create(color, defaultColor).materialize(len(mesh.polygons.indices))
+        
+        mesh.insertVertexColorLayer(colorLayerName, colorsList)
+        return mesh
+        
+    def execute_ColorsList(self, mesh, colorLayerName, colors):
+        if mesh is None: return None
+        if colorLayerName == "": return mesh
```
OmarEmaraDev: We should probably inform the user using exception error handling.

animation_nodes/nodes/mesh/get_mesh_vertex_color.py
```diff
@@ -0,0 +1,56 @@
...
+        self.newInput("Mesh", "Mesh", "mesh")
+        self.newInput("Text", "Name", "colorLayerName")
+
+        self.newOutput("Color List", "Colors", "colors")        
+
+    def draw(self, layout):
+        layout.prop(self, "colorMode", text = "")
+
+    def execute(self, mesh, colorLayerName):
+        if mesh is None or colorLayerName not in mesh.getVertexColorLayerNames(): return ColorList()
```
OmarEmaraDev: Inform the user using exception error handling here as well.

animation_nodes/nodes/mesh/get_mesh_vertex_color.py
```diff
@@ -0,0 +1,56 @@
...
+
+        defaultColor = Color((0, 0, 0, 1))
+        
+        colorsList = ColorList(length = len(mesh.polygons.indices))
+        
+        for name, data in mesh.getVertexColorLayers():
+            if name == colorLayerName:
+                colorsList = data
+
```
OmarEmaraDev: Add a method to the mesh that returns the vertex color layer for the given name.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -141,13 +154,18 @@ cdef class Mesh:
                 Vertices: {}
                 Edges: {}
```
OmarEmaraDev: Try replacing this with an f-string.

animation_nodes/nodes/mesh/add_mesh_vertex_color.py
3DSinghVFX: Omar, `Add Mesh Vertex Color` and  `Get Mesh Vertex Color` nodes do similar jobs, i.e., one node insert the vertex color layer and its colors, and the other gets the colors of a vertex color layer. I think if their names are similar then it would be nice. 

Add Mesh Vertex Color -> Insert Vertex Color Layer
Get Mesh Vertex Color -> Get Vertex Color Layer 

or

Add Mesh Vertex Color -> Mesh Vertex Color Output  or  Vertex Color Layer Output
Get Mesh Vertex Color -> Mesh Vertex Color Input  or  Vertex Color Layer Input

OmarEmaraDev: Yes, sure. This looks good:

Add Mesh Vertex Color -> Insert Vertex Color Layer
Get Mesh Vertex Color -> Get Vertex Color Layer

3DSinghVFX: Okay.

3DSinghVFX:
Omar, I'll remove the Vertex Color Input node. The reason I kept this node is because there is a bug in `color/c_utils.pyx`, so the Vertex Color Input node gives different colors,
RK1HfKVU jpg large


https://github.com/JacquesLucke/animation_nodes/blob/af8354fffcb9a05290fe81cebd3ad9ccbcf0c8d7/animation_nodes/nodes/color/c_utils.pyx#L64

When I change this to `addColor_Inplace(polygonColors.data + i, colors.get(index))` then Vertex Color Input node gives the same colors.

OmarEmaraDev:
Sure, you may fix the bug in this pull request as well.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Omar, I have checked with the fresh AN compile build, there is no `runtime` warning. It is working fine.

OmarEmaraDev: Good. To clarify, I wanted you to using python's f-strings here. See https://www.python.org/dev/peps/pep-0498/

animation_nodes/nodes/mesh/add_mesh_vertex_color.py
3DSinghVFX: Added.

animation_nodes/nodes/mesh/add_mesh_vertex_color.py
3DSinghVFX: I have added the exception error handling.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -173,6 +191,13 @@ cdef class Mesh:
         for i in range(meshData.polygons.indices.length):
             self.polygons.indices.data[polygonIndicesOffset + i] += vertexOffset
 
+    def getVertexColorLayer(self, str colorLayerName):
+        cdef long i    
+        for i in range(len(self.getVertexColorLayers())):
+            if self.getVertexColorLayers()[i][0] == colorLayerName:
+                return self.getVertexColorLayers()[i]    
+            else:
+                return None
```
OmarEmaraDev: `vertexColorLayers` is an ordered dictionary, you can simply get the vertex color layer using subscript notation. Also, move this function below the `getVertexColorLayerNames`.

animation_nodes/nodes/mesh/get_mesh_vertex_color.py
3DSinghVFX: Here, it is also added the exception error handling.

animation_nodes/nodes/mesh/get_mesh_vertex_color.py
3DSinghVFX: Now this is a method which gives vertex color layer for the given name.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Is this correct way,
```
def getVertexColors(self, str colorLayerName):
        return self.vertexColorLayers.get(colorLayerName, None)
```
This function return the colors corresponds to the colorLayerName.

OmarEmaraDev: Yes, looks good.

3DSinghVFX:
Omar, I have used the f-strings, here is the code:
```
def __repr__(self):
        return textwrap.dedent(
        f"""AN Mesh Object:
        Vertices: {len(self.vertices)}
        Edges: {len(self.edges)}
        Polygons: {len(self.polygons)}
        UV Maps: {list(self.uvMaps.keys())}
        Vertex Color Layers: {list(self.vertexColorLayers.keys())}""")
```
I think we should change the `AN Mesh Object:` to `AN Mesh:`, is it not?.

OmarEmaraDev:
I think *Object* here denotes an instance of the AN Mesh class, not a Blender object. So just leave it for now.

3DSinghVFX:
> I think _Object_ here denotes an instance of the AN Mesh class, not a Blender object. So just leave it for now.

Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: I have added this function.

3DSinghVFX:
Omar, In Insert Vertex Color Layer node, if we add a color layer of the same name which already there then does not notify the user. Should we add the exception for that,
screen
Or the viewer node is enough for user to see whether that color layer is already there or not.

OmarEmaraDev:
I think we should do that. Raise an exception informing the user about the existing layer.

I will continue the review on Sunday as I don't have much time right now.

3DSinghVFX:
Okay.

No problem :)

3DSinghVFX:
Omar, I have noticed another problem with the order of color layers arises if we use Set Vertex Color with Color Layer Index option.

In this example, I have inserted two color layers. Then try setting the color to color layer _AN-Col2_ which has index zero with the Set Vertex Color node, 
Vertex_Color_Layer3

However, in this case, if include the existing color layer (_Col_) or insert a color layer before the _AN-Col2_ then the Index get change for _AN-Col2_ color layer for Set Vertex Color node,
Vertex_Color_Layer4

So, I think we should also use the name of the color layer in the Set Vertex Color node to resolve this problem. Do have any other idea?

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -136,18 +152,21 @@ cdef class Mesh:
...
-                Polygons: {}
-                UV Maps: {}\
-            """.format(len(self.vertices), len(self.edges), len(self.polygons), list(self.uvMaps.keys())))
+        return textwrap.dedent(
+        f"""AN Mesh Object:
+        Vertices: {len(self.vertices)}
+        Edges: {len(self.edges)}
+        Polygons: {len(self.polygons)}
+        UV Maps: {list(self.uvMaps.keys())}
+        Vertex Color Layers: {list(self.vertexColorLayers.keys())}""")
```
OmarEmaraDev: Can't we use `getVertexColorLayerNames` here?

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,60 @@
...
+colorModeItems = [
+    ("LOOP", "Loop", "Get color of every loop vertex", "NONE", 0),
+    ("VERTEX", "Vertex", "Get color of every vertex", "NONE", 1),
+    ("POLYGON", "Polygon", "Get color of every polygon", "NONE", 2)    
+]
+
+class GetVertexColorLayerNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetVertexColorLayerNode"
+    bl_label = "Get Vertex Color Layer"
+    bl_width_default = 155
```
OmarEmaraDev: I would rather we don't use custom widths here.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,60 @@
...
+        self.newInput("Mesh", "Mesh", "mesh", dataIsModified = True)
+        self.newInput("Text", "Name", "colorLayerName")
+
+        self.newOutput("Color List", "Colors", "colors")        
+
+    def draw(self, layout):
+        layout.prop(self, "colorMode", text = "")
+
+    def execute(self, mesh, colorLayerName):
+        if mesh is None: return ColorList()
```
OmarEmaraDev: We can use an exception here as well.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,60 @@
...
+        if colorLayerName == "":
+            self.raiseErrorMessage("No Vertex Color Layer Name.")
```
OmarEmaraDev: -> Vertex color layer name can't be empty.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,60 @@
...
+        
+        defaultColor = Color((0, 0, 0, 1))
+        colorsList = ColorList(length = len(mesh.polygons.indices))
+        
+        if mesh.getVertexColors(colorLayerName) is not None:
+            colorsList = mesh.getVertexColors(colorLayerName)
```
OmarEmaraDev: You are calling `mesh.getVertexColors` two times here.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,60 @@
...
+        else:
+            self.raiseErrorMessage(f"No Vertex Color Layer of Name '{colorLayerName}'."
+                                   f" Layers: {', '.join(mesh.getVertexColorLayerNames())}")
```
OmarEmaraDev: This might be a little long for an error message. Also, `of name` -> `with name`.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,60 @@
...
+            
+            colorsList = VirtualColorList.create(colorsList, defaultColor)
+            return getVertexColorsFromLoopColors(vertexCount, polygonIndices, colorsList)
+        elif self.colorMode == "POLYGON":
+            polygonIndices = mesh.polygons
+
+            colorsList = VirtualColorList.create(colorsList, defaultColor)
+            return getPolygonColorsFromLoopColors(polygonIndices, colorsList)
+        
+
```
OmarEmaraDev: New line at the end of file.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: I added the `f" Layers: {', '.join(mesh.getVertexColorLayerNames())}"` to avoid guesswork or to check names with Viewer Node. Should I remove this line to make the error message short?

OmarEmaraDev: Yes. The Mesh repr method already provide those names.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: Omar, For input mesh,
https://github.com/JacquesLucke/animation_nodes/blob/419e1e73431b17cb5cfa6cce76acc588321c5092/animation_nodes/nodes/mesh/get_vertex_color_layer.py#L24
due to this, I think mesh can't none, is it not?

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: It is replaced.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
OmarEmaraDev: An input mesh can be None. But if it is None, I would consider this an issue with the node the produced the None value. So maybe you are right, the None check is not needed here.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: Now it looks like this `self.raiseErrorMessage(f"Mesh doesn't has '{colorLayerName}' Vertex Color Layer.")`. is it fine?

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: Removed.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
OmarEmaraDev: I think this is better:

> Mesh doesn't have a vertex color layer with the name '{colorLayerName}'.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Yes. I have changed it, `Vertex Colors: {self.getVertexColorLayerNames()}`.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: Okay, here it is
Screenshot from 2019-12-23 00-24-14

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: It is changed.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: Removed the extra line.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
OmarEmaraDev: Looks good.

3DSinghVFX:
Omar, How we are going to handle this problem: https://github.com/JacquesLucke/animation_nodes/pull/1256#issuecomment-568158138

OmarEmaraDev:
Well, the user can always identify the vertex color by name by changing the mode from the advanced node settings. So I don't really consider this an options.

3DSinghVFX:
Okay.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,57 @@
...
+            self.raiseErrorMessage("Vertex color layer name can't be empty.")
+        
+        defaultColor = Color((0, 0, 0, 1))
+        colorsList = ColorList(length = len(mesh.polygons.indices))
+        
+        getColorList = mesh.getVertexColors(colorLayerName)
+        if getColorList is not None:
+            colorsList = getColorList
+        else:
+            self.raiseErrorMessage(f"Mesh doesn't have a vertex color layer with the name '{colorLayerName}'.")
```
OmarEmaraDev:
```
colorsList = mesh.getVertexColors(colorLayerName)
if colorsList is None:
    raise
continue as usual without an else.
```

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,57 @@
...
+            vertexCount = len(mesh.vertices)
+            
+            colorsList = VirtualColorList.create(colorsList, defaultColor)
+            return getVertexColorsFromLoopColors(vertexCount, polygonIndices, colorsList)
+        elif self.colorMode == "POLYGON":
+            polygonIndices = mesh.polygons
+
+            colorsList = VirtualColorList.create(colorsList, defaultColor)
+            return getPolygonColorsFromLoopColors(polygonIndices, colorsList)
+
```
OmarEmaraDev: Setup your editor to add a new line at the end of file.

3DSinghVFX: Sorry. I'm using Visual Studio.

3DSinghVFX: I have setup a new line in my Visual Studio, this how it shows the code,
Screenshot from 2019-12-23 01-56-05

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: It is done.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
```diff
@@ -0,0 +1,54 @@
...
+    def draw(self, layout):
+        layout.prop(self, "colorMode", text = "")
+
+    def execute(self, mesh, colorLayerName):
+        if colorLayerName == "":
+            self.raiseErrorMessage("Vertex color layer name can't be empty.")
+        
+        defaultColor = Color((0, 0, 0, 1))
+        colorsList = ColorList(length = len(mesh.polygons.indices))
```
OmarEmaraDev: This is redundant.

animation_nodes/nodes/mesh/mesh_object_input.py
```diff
@@ -6,6 +6,7 @@
 class MeshObjectInputNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_MeshObjectInputNode"
     bl_label = "Mesh Object Input"
+    bl_width_default = 145
```
OmarEmaraDev: I would prefer if we keep the default width here as well.

animation_nodes/nodes/mesh/mesh_object_input.py
```diff
@@ -14,6 +15,7 @@ def create(self):
         self.newInput("Boolean", "Use World Space", "useWorldSpace")
         self.newInput("Boolean", "Use Modifiers", "useModifiers", value = False)
         self.newInput("Boolean", "Load UVs", "loadUVs", value = False)
+        self.newInput("Boolean", "Load Vertex Colors", "loadVertexColors", value = False)
```
OmarEmaraDev: Can you hide both of those inputs by default?

3DSinghVFX: Done.

animation_nodes/nodes/mesh/mesh_object_input.py
3DSinghVFX: Now it has the default width.

animation_nodes/nodes/mesh/get_vertex_color_layer.py
3DSinghVFX: Removed.

animation_nodes/nodes/mesh/mesh_object_input.py
```diff
@@ -33,7 +32,7 @@ def create(self):
         self.newOutput("Integer List", "Material Indices", "materialIndices")
 
         self.newOutput("Text", "Mesh Name", "meshName")
```
3DSinghVFX: Omar, Mesh name output has a bug.

OmarEmaraDev: What is the bug?

3DSinghVFX: When I connect the viewer node to the mesh name output then it breaks the node-tree:
Screenshot from 2019-12-23 15-45-23

OmarEmaraDev: Fixed in b442dfe.

3DSinghVFX: Okay :)

OmarEmaraDev:
Can you add the changes to the CHANGELOG file? Notice that we are still on the same sub version, so adjust the file appropriately. The node that was removed in this pull request should be removed from the CHANGELOG as well.

3DSinghVFX:
Okay, I'll update the CHANGELOG file.

3DSinghVFX:
I have added the changes about the insert and get vertex color layer nodes, and removed the `- Added Vertex Color Input node.`. Do I have to add anything else?

```
## 2.1.5

### Added

- Added *Include End Point* option to the *Float Range* node.
- Added collection option to *Execution Triggers*.
- Added *Sound Falloff* node.
- Added *Subprograms* to the search menu.
- Added missing nodes to the node menu
- Added *Collection Operations* node.
- Added *Rotations* output to the *Particles Data* node.
- Added Get Vertex Color Layer node.
- Added Insert Vertex Color Layer node.
```

OmarEmaraDev:
No, look good.

CHANGELOG.md
```diff
@@ -37,6 +36,8 @@
 - Added missing nodes to the node menu
 - Added *Collection Operations* node.
 - Added *Rotations* output to the *Particles Data* node.
+- Added Get Vertex Color Layer node.
+- Added Insert Vertex Color Layer node.
```
OmarEmaraDev: Why did you add them to 2.1.5? They should go to the Unreleased section.

3DSinghVFX: Sorry :) Now they are in the Unreleased section under the Added category.

OmarEmaraDev:
Shouldn't the Insert Vertex Color Layer node have options for Loop, Vertex, and Polygon like the Set Vertex Color node?

OmarEmaraDev:
My bad, just realized the options are there.

3DSinghVFX:
Thank you for helping me to understand the different aspects of AN :)

PR #1265, opened 2019-12-26
Find Linked Edges node, Modes for Line Mesh node and Updated the Node Menu

3DSinghVFX:
1. This branch has a new node that finds the linked edges to a vertex. The find function is in the Cython for this node,
Screenshot from 2019-12-26 16-38-34

2. The updated Line Mesh node has two modes: Start-End or Points with the cyclic option,
Screenshot from 2019-12-26 16-38-05

3. The node_menu.py file is updated.


OmarEmaraDev:
I will start the review on Saturday.

3DSinghVFX:
Okay, It is fine :)

OmarEmaraDev:
A pull request should contain only a single change, you have three here. We shall discuss each change in its respective pull request.

3DSinghVFX:
Okay, I'll make a separate pull request for each change.

PR #1268, opened 2019-12-29
Updated Node-Menu

3DSinghVFX:
This branch has updated node_menu.py file. It has all the nodes in the "Add" node-menu.


OmarEmaraDev:
We don't want to increase the length of the add menu too much. Perhaps we should put the collections menu in the Object menu, in place of the Collection Info node.

Also, the *Interface Menu* doesn't really make sense here. The two nodes inside are different types of interfaces. Perhaps they should go to the subprograms menu.

3DSinghVFX:
Okay. I have updated the "Add" node-menu.

animation_nodes/ui/node_menu.py
```diff
@@ -201,6 +201,7 @@ def draw(self, context):
         layout.separator()
         insertNode(layout, "an_TextBlockReaderNode", "Block Reader")
         insertNode(layout, "an_TextBlockWriterNode", "Block Writer")
+        insertNode(layout, "an_FilterBlendDataListByNameNode", "Filter Text Block List", {"dataType" : repr("Text")})
```
OmarEmaraDev: Shouldn't this be `Text Block` instead of `Text`?

3DSinghVFX: I also thought that but in the filter node, the data type is written as "Text Block".

OmarEmaraDev: I mean it should be `repr("Text Block")`. `Text` is not supported by the node as far as I know.

3DSinghVFX: Sorry, I misunderstood. Yes, it is `Text Block`. However, it is also working with `Text`.

3DSinghVFX: It is fixed.

OmarEmaraDev: Good. But I don't think it works with `Text`, str doesn't have a name attribute.

3DSinghVFX: The node is not working properly. But it does not show any error,
Screenshot from 2019-12-29 17-51-12

PR #1270, opened 2019-12-29
Line Mesh node with new modes.

3DSinghVFX:
The new Line Mesh node has two modes, **Start-End** and **Points**. The newly added **Points** mode allows creating the mesh from point list and it also has option for cyclic,
71473889-fee7c500-27fe-11ea-8a2c-cc25603c779e


OmarEmaraDev:
It seems weird to call this a "Line", which it isn't. What do you think?

3DSinghVFX:
Yes. But a line can have two or more than two points. Should we call this as **Curve Mesh** node?

OmarEmaraDev:
While it may have two or more points, it can't have irregularities, it has to be a line segment. Not sure about the naming, let me think about that.

OmarEmaraDev:
Perhaps we should just rename the node to *Lines Mesh* node, that way, it works both ways. This is also how OpenGL and other frameworks call it. What do you think?

3DSinghVFX:
Yes, this is good.

3DSinghVFX:
But from this name, it looks like the node will generate multiple lines, is it not?

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,52 @@
...
+            return "execute_PointsLine"
+
+    def execute_StartEndLine(self, start, end, steps):
         steps = max(steps, 2)
-        return getLineMesh(start, end, steps)
\ No newline at end of file
+        return getLineMesh(start, end, steps)
+
+    def execute_PointsLine(self, points, cyclic):
+        if len(points) < 2: return Mesh()
```
OmarEmaraDev: We should probably throw an exception here.

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,52 @@
...
+        pointCount = len(points)
+        edgeIndices = edges(pointCount)
+        if cyclic:
+            edgeIndices.append((pointCount - 1, 0))
+            return Mesh(points, edgeIndices, skipValidation = True)
+        else:    
+            return Mesh(points, edgeIndices, skipValidation = True)
```
OmarEmaraDev: I think we should create a function for it in `line.pyx` instead of doing things manually here.

OmarEmaraDev:
Well, it is technically creating multiple lines. :smiley:

3DSinghVFX:
> Well, it is technically creating multiple lines. smiley

You are right, but multiple lines seems to me they are not connected to each other.

animation_nodes/nodes/mesh/generation/line.py
3DSinghVFX: Removed.

animation_nodes/nodes/mesh/generation/line.py
3DSinghVFX: Okay. Omar, what type of identifier for boolean in Cython which is cyclic in this case.

3DSinghVFX: I'm using this `from cpython cimport bool`.

3DSinghVFX: Here is the function for Lines mesh,
```
def getLinesMesh(Vector3DList points, bool cyclic):
    cdef long pointCount = len(points)
    edgeIndices = edges(pointCount)
    if cyclic:
        edgeIndices.append((pointCount - 1, 0))
        return Mesh(points, edgeIndices, skipValidation = True)
    else:    
        return Mesh(points, edgeIndices, skipValidation = True)
```

OmarEmaraDev: Execute `cython -a line.pyx`. A `line.html` file will be generated. This is an annotated version of the code, yellow lines means a kind of interaction with Python, we try to minimize that as much as possible. So, the less yellow you have, the better. In this code, the third line will be yellow, because edgeIndices is not typed. So will the fifth line, because you are calling append.

3DSinghVFX: I have updated the code,
```
def getLinesMesh(Vector3DList points, bool cyclic):
    cdef long pointCount = len(points)
    cdef EdgeIndicesList edgeIndices = edges(pointCount)
    if cyclic:
        edgeIndices.append((pointCount - 1, 0))
        return Mesh(points, edgeIndices, skipValidation = True)
    else:    
        return Mesh(points, edgeIndices, skipValidation = True)
```

OmarEmaraDev: I would also try to avoid calling append.

3DSinghVFX: Sorry, but I do know how I can add an extra edge-indices in EdgeIndicesList.

OmarEmaraDev: Pass `cyclic` to the `edges` function and generate the edges accordingly.

3DSinghVFX: Thanks. Here is the code,
```
# Edges
############################################

def edges(long steps, bool cyclic):
    cdef long edgeAmount
    if cyclic:
        edgeAmount = steps
    else:
        edgeAmount = steps - 1
    cdef EdgeIndicesList edges = EdgeIndicesList(length = edgeAmount)
    cdef long i 
    for i in range(edgeAmount):
        edges.data[i].v1 = i
        edges.data[i].v2 = i + 1
    if cyclic:
        edges.data[edgeAmount - 1].v2 = 0
    return edges


# Points
############################################

def getLinesMesh(Vector3DList points, bool cyclic):
    cdef long pointCount = len(points)
    cdef EdgeIndicesList edgeIndices = edges(pointCount, cyclic)
    return Mesh(points, edgeIndices, skipValidation = True)
```

OmarEmaraDev: Use `bint` for boolean types.

3DSinghVFX: Okay.

3DSinghVFX:
Omar, Should I also change the `line.py` to `lines.py` or just rename the node?

OmarEmaraDev:
@3DSinghVFX You know, maybe we should leave it as is. Sorry. Just leave everything as `Line`.

3DSinghVFX:
Okay. Changing the name will break the old node-trees.

animation_nodes/algorithms/mesh_generation/line.pyx
```diff
@@ -33,10 +41,17 @@ def vertices(start, end, long steps):
...
-def edges(long steps):
-    edges = EdgeIndicesList(length = steps - 1)
-    cdef long i
-    for i in range(steps - 1):
+def edges(long steps, bint cyclic):
+    cdef long edgeAmount
+    if cyclic:
+        edgeAmount = steps
+    else:
+        edgeAmount = steps - 1
```
OmarEmaraDev: I think this can be an inline-if.

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,46 @@
...
+            return "execute_PointsLine"
+
+    def execute_StartEndLine(self, start, end, steps):
         steps = max(steps, 2)
-        return getLineMesh(start, end, steps)
\ No newline at end of file
+        return getLineMesh(start, end, steps)
+
+    def execute_PointsLine(self, points, cyclic):
+        if len(points) == 0: return Mesh()
```
OmarEmaraDev: Use an exception here.

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,46 @@
...
+        return getLinesMesh(points, cyclic)
+
```
OmarEmaraDev: New line at end of file.

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,46 @@
 import bpy
+from bpy.props import *
+from .... data_structures import Mesh
 from .... base_types import AnimationNode
-from .... algorithms.mesh_generation.line import getLineMesh
+from .... algorithms.mesh_generation.line import getLineMesh, getLinesMesh
+
+lineModeItems = [
+    ("STARTEND", "Start-End", "Line from start-end", "NONE", 0),
```
OmarEmaraDev: Make this `START_END`.

animation_nodes/nodes/mesh/generation/line.py
3DSinghVFX: This is the code,
```
if len(points) == 0:
            self.raiseErrorMessage("Points list is empty.")
```

animation_nodes/nodes/mesh/generation/line.py
3DSinghVFX: Modified.

animation_nodes/nodes/mesh/generation/line.py
OmarEmaraDev: The point list is only valid if there are more than 1 points, right?

3DSinghVFX: Yes.

animation_nodes/nodes/mesh/generation/line.py
3DSinghVFX: Added.

animation_nodes/algorithms/mesh_generation/line.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/line.pyx
```diff
@@ -33,10 +41,15 @@ def vertices(start, end, long steps):
 # Edges
 ############################################
 
-def edges(long steps):
-    edges = EdgeIndicesList(length = steps - 1)
-    cdef long i
-    for i in range(steps - 1):
+def edges(long steps, bint cyclic):
+    steps = steps if cyclic and steps > 2 else steps - 1
```
OmarEmaraDev: Why are you checking for `steps > 2`? This is what you might call a low-level function, it expects valid input and error handling is done at a higher level. Also, it is better if you create a new variable just like you did before instead of writing the the input directly.

animation_nodes/nodes/mesh/generation/line.py
OmarEmaraDev: It seems you misunderstood me. Text files should always have a new line as the very last characters of the file. Editors typically do this automatically. You can setup VS Code to do that by adding the following to your config:
```
    "files.trimTrailingWhitespace": true,
    "files.insertFinalNewline": true,
    "files.trimFinalNewlines": true,
```

Also see https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

animation_nodes/algorithms/mesh_generation/line.pyx
3DSinghVFX: If cyclic is True and Points are two then it will generate two similar edges which will later create a problem. That is why I have added a check for `steps > 2`.

OmarEmaraDev: In this case, you can't have two points only when cyclic is enabled. We can raise an exception then.

3DSinghVFX:
Omar, I have enabled the setting you suggested in the VS Code. Here is the code in VS Code, which shows a line in the end after saving,
Screenshot from 2019-12-30 01-37-17

and Sublime editor also shows a line for the same code.

However, the same code in _gedit_ which does not show a line in the end,
Screenshot from 2019-12-30 01-37-42

Which one is correct?

OmarEmaraDev:
gedit probably just doesn't show it. I think you are good.
By the way, Github show this in the code if you want to make sure:

3DSinghVFX:
> gedit probably just doesn't show it. I think you are good.
> By the way, Github show this in the code if you want to make sure:
> 
> 

Thanks. Next time I'll keep eye on this.

animation_nodes/algorithms/mesh_generation/line.pyx
3DSinghVFX: Okay. I'll add an exception in the line.py file.

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,50 @@
...
-        return getLineMesh(start, end, steps)
\ No newline at end of file
+        return getLineMesh(start, end, steps)
+
+    def execute_PointsLine(self, points, cyclic):
+        pointAmount = len(points)
+        if pointAmount < 2:
+            self.raiseErrorMessage("Points list should have atleast two points.")
+        if pointAmount < 3 and cyclic:
+            self.raiseErrorMessage("For cyclic, Points list should have atleast three points.")
```
OmarEmaraDev: `atleast` -> `at least`. The same for the above message.

animation_nodes/algorithms/mesh_generation/line.pyx
```diff
@@ -33,10 +41,15 @@ def vertices(start, end, long steps):
 # Edges
 ############################################
 
-def edges(long steps):
-    edges = EdgeIndicesList(length = steps - 1)
+def edges(long steps, bint cyclic):
+    cdef long edgeAmount
+    edgeAmount = steps if cyclic else steps - 1
```
OmarEmaraDev: You can put the initialization in the declaration directly.

animation_nodes/nodes/mesh/generation/line.py
```diff
@@ -1,18 +1,50 @@
 import bpy
+from bpy.props import *
+from .... data_structures import Mesh
```
OmarEmaraDev: Is `Mesh` really needed?

3DSinghVFX: Not. I have removed it.

animation_nodes/algorithms/mesh_generation/line.pyx
3DSinghVFX: Modified.

animation_nodes/nodes/mesh/generation/line.py
3DSinghVFX: Changed.

OmarEmaraDev:
@3DSinghVFX Last thing is the CHANGELOG file.

3DSinghVFX:
Okay.

3DSinghVFX:
Unreleased > Changed > 
- Added modes in the Line Mesh node.

Correct?

OmarEmaraDev:
Unreleased -> Added -> Added Points mode to the Line Mesh node.

3DSinghVFX:
Thanks Omar for everything :)

PR #1275, opened 2019-12-30
Find Linked Edges node

3DSinghVFX:
This branch has a new **Find Linked Edges** node. This node finds the linked edges and vertex indices to a vertex. The find function of this node is in the Cython.

It takes the edge indices list and vertex index. It output the verties(indices), edges(indices) and linked edge indices,
71473837-cb0c9f80-27fe-11ea-8831-fc4bf16052f9


JacquesLucke:
Personally, I think this node should not go in as is. The issue is that it has bad time complexity. Meaning, every time the node is executed, it internally iterates over all edges. This could become a real bottleneck if the node is used in a loop.

A possible solution could be caching. You could create some kind of lookup table that maps every vertex to its incident edges. This table only has to be build once for a static mesh, subsequent lookups using this node should be fast. The lookup table should only be created when necessary. When the topology of the mesh changes, it has to be deleted.

3DSinghVFX:
You are absolutely right.

Thanks for the suggestions. I'll make a lookup table.

OmarEmaraDev:
There is something wrong with the pull request, you have changes from #1270. Can you fix that?

3DSinghVFX:
> There is something wrong with the pull request, you have changes from #1270. Can you fix that?

Sorry, Omar. I did not understand what is wrong?

OmarEmaraDev:
Look at the *Files changed* tab. How exactly do you create your pull requests?

3DSinghVFX:
Ohhh. How can I fix this?

OmarEmaraDev:
I actually don't know how this can happen, so I don't know. But your pull requests seems generally strange due to the huge commit list.

First, make sure your master is identical to upstream master, always. Then branch from **master** creating a new branch, then create a pull request from this branch.

3DSinghVFX:
Okay. Should I create a new pull request or update the master then create a branch with the same name and commit the changes.

OmarEmaraDev:
It seems this behavior is described [here](https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch). I squashed your commits from #1270 but you merge them, hence the difference. I think you just need to rebase master and push again.

3DSinghVFX:
Okay.

3DSinghVFX:
Omar, it looks like the rebase master did not fix the problem.

OmarEmaraDev:
You haven't rebased, check the third answer in the questions I sent.

3DSinghVFX:
It shows me this error,
Screenshot from 2019-12-31 00-00-16

OmarEmaraDev:
There is obviously a conflict that needs to be fixed. But I still don't know why this happens, I don't recall touching the Material Output node recently.

Just close this pull request and open another one.

3DSinghVFX:
Okay.

PR #1277, opened 2019-12-30
Find Linked Edges node

3DSinghVFX:
This branch has a new **Find Linked Edges** node. This node finds the linked edges and vertex indices to a vertex. The find function of this node is in the Cython.

It takes the edge indices list and vertex index. It output the verties(indices), edges(indices) and linked edge indices,
71473837-cb0c9f80-27fe-11ea-8831-fc4bf16052f9

Reference of old request (closed): https://github.com/JacquesLucke/animation_nodes/pull/1275

### DIFF
--- a/animation_nodes/data_structures/meshes/mesh_data.pyx
+++ b/animation_nodes/data_structures/meshes/mesh_data.pyx
@@ -4,7 +4,7 @@ import functools
 from collections import OrderedDict
 from . validate import checkMeshData, calculateLoopEdges
 from .. lists.base_lists cimport (
-    UIntegerList, EdgeIndices, EdgeIndicesList, Vector3DList, Vector2DList, ColorList, LongList)
+    UIntegerList, EdgeIndices, EdgeIndicesList, Vector3DList, Vector2DList, ColorList, IntegerList)
 from ... math cimport (
     Vector3, crossVec3, subVec3, addVec3_Inplace, isExactlyZeroVec3, normalizeVec3,
     Matrix4, toMatrix4
@@ -98,7 +98,7 @@ cdef class Mesh:
 
     @derivedMeshDataCacheHelper("Linked Edges")
     def getVerticesLinkedEdges(self):
-        return findVerticesLinkedEdges(self.vertices, self.edges)
+        return findNeighboursEdges(self.vertices, self.edges)
 
     def setLoopEdges(self, UIntegerList loopEdges):
         if len(loopEdges) == len(self.polygons.indices):
@@ -146,7 +146,22 @@ cdef class Mesh:
         return self.vertexColorLayers.get(colorLayerName, None)
 
     def getVertexLinkedEdges(self, long vertexIndex):
-        return self.getVerticesLinkedEdges()[vertexIndex]
+        cdef IntegerList neighboursAmounts, neighboursStarts, neighbours, neighbourEdges
+        neighboursAmounts, neighboursStarts, neighbours, neighbourEdges = self.getVerticesLinkedEdges()
+
+        cdef i, start, amount
+        amount = neighboursAmounts[vertexIndex]
+        start = neighboursStarts[vertexIndex]
+
+        cdef IntegerList vertexNeighbours = IntegerList(length = amount)
+        cdef IntegerList vertexNeighbourEdges = IntegerList(length = amount)
+
+        cdef j = 0, end = start + amount
+        for i in range(start, end):
+            vertexNeighbours.data[j] = neighbours.data[i]
+            vertexNeighbourEdges.data[j] = neighbourEdges.data[i]
+            j += 1
+        return vertexNeighbours, vertexNeighbourEdges
 
     def copy(self):
         mesh = Mesh(self.vertices.copy(), self.edges.copy(), self.polygons.copy())
@@ -157,12 +172,10 @@ cdef class Mesh:
     def transform(self, transformation):
         self.vertices.transform(transformation)
         self.verticesTransformed()
-        self.topologyChanged()
 
     def move(self, translation):
         self.vertices.move(translation)
         self.verticesMoved()
-        self.topologyChanged()
 
     def __repr__(self):
         return textwrap.dedent(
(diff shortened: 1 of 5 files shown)


3DSinghVFX:
Thanks, Omar.

I have modified the code of `mesh_data.pyx`. To find neighbours for vertices, I have used the code from the spline and modified it for neighbour-edges.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,11 +140,29 @@ cdef class Mesh:
...
         return self.vertexColorLayers.get(colorLayerName, None)
-        
+
+    def getVertexLinkedEdges(self, long vertexIndex):
+        cdef IntegerList neighboursAmounts, neighboursStarts, neighbours, neighbourEdges
+        neighboursAmounts, neighboursStarts, neighbours, neighbourEdges = self.getVerticesLinkedEdges()
+
+        cdef int start, amount
+        amount = neighboursAmounts[vertexIndex]
+        start = neighboursStarts[vertexIndex]
```
OmarEmaraDev: Don't forget `.data`. `neighboursAmounts.data[vertexIndex]`

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -314,3 +340,38 @@ def calculateCrossProducts(Vector3DList a, Vector3DList b):
 
     return result
 
+def findNeighboursEdges(Vector3DList vertices, EdgeIndicesList edges):
```
OmarEmaraDev: Pass the length of `vertices` directly.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Okay. Should I also changed the name of the related functions in the `mesh_data.pyx` ?

OmarEmaraDev: Yes, I think this would be a good idea.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Okay. The function's name `findNeighboursEdges` or `findNeighbours`?

OmarEmaraDev: For consistency, make it `getLinkedVertices`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -92,6 +96,10 @@ cdef class Mesh:
     def getPolygonBitangents(self):
         return calculatePolygonBitangents(self.getPolygonTangents(), self.getPolygonNormals())
 
+    @derivedMeshDataCacheHelper("Linked Vertices")
+    def getLinkedVerticesCache(self):
```
OmarEmaraDev: You can use the same naming convention used for other attributes.
`getLinkedVertices` -> `calculateLinkedVertices`.
`getLinkedVerticesCache` -> `getLinkedVertices`.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Renamed.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,11 +140,29 @@ cdef class Mesh:
...
+        start = neighboursStarts.data[vertexIndex]
+
+        cdef IntegerList vertexNeighbours = IntegerList(length = amount)
+        cdef IntegerList vertexNeighbourEdges = IntegerList(length = amount)
+
+        cdef int i, index
+        for i in range(amount):
+            index = i + start
+            vertexNeighbours.data[i] = neighbours.data[index]
+            vertexNeighbourEdges.data[i] = neighbourEdges.data[index]
```
OmarEmaraDev: I wounder if we can use `memcpy` here.
Import it `from libc.string cimport memcpy`.
Then check its man page in section 3 to see how to use it.

3DSinghVFX: Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Omar, I have tried that and it is working. 

How can I slice the `InterList` e.g., `neighbours.data[start : end]`?

OmarEmaraDev: Why would you need to slice the list?

3DSinghVFX: The `neighbours` is a list of neighbours for all vertices. To get neighbours for a vertex we need to slice this list. It is happening in the loop from `start - start + amount`: https://github.com/JacquesLucke/animation_nodes/blob/9b1b9cce710d263cf95dc5fc67953926f1241dfb/animation_nodes/data_structures/meshes/mesh_data.pyx#L160-L163

OmarEmaraDev: Yes, that's why I am proposing to use `memcpy` to copy from the `neighbours` lists to the output lists. What is your current code?

JacquesLucke: I think slicing should work as well. Internally it will do the same as memcpy I think, just with a little bit more overhead. Use `neighbours[start:end]` if I remember correctly.

3DSinghVFX: I thought we are not going to use for loop. I run this code just for testing it for a list
```
cdef IntegerList vertexNeighboursTest = IntegerList(length = amount)
memcpy(vertexNeighboursTest.data, vertexNeighbours.data, sizeof(vertexNeighbours))
```

OmarEmaraDev: I looked into the `getValuesInSlice` code, it does a bit more than we need. While it has little overhead, explicit memcpy is just two extra lines of code. So I think it is acceptable.

3DSinghVFX: But in a loop it gives a error for this code,
```
for i in range(amount):
    index = i + start
    memcpy(vertexNeighbours.data[i], neighbours.data[index], sizeof(neighbours.data[index]))
```
Screenshot from 2019-12-31 17-25-54

OmarEmaraDev: I meant something like this:
```cython
memcpy(vertexNeighbours.data, neighbours.data + start, amount * sizeof(long));
memcpy(vertexNeighbourEdges.data, neighbourEdges.data + start, amount * sizeof(long));
```

3DSinghVFX: Okay

3DSinghVFX: Omar, we run the animation, the Output is flickering. Is it due to _[To avoid overflows, the size of the arrays pointed to by both the destination and source parameters, shall be at least num bytes, and should not overlap (for overlapping memory blocks, memmove is a safer approach).](http://www.cplusplus.com/reference/cstring/memcpy/)_

OmarEmaraDev: Ah, probably because we are using integer lists while we have `sizeof(long)`. It should be noted that Integer lists in AN are actually Long lists behind the scenes, so you should return a Long list instead. In conclusion, use LongList instead.

Since we already encountered an unexpected behavior, we should probably listen to @JacquesLucke, use slicing for safety and readability. :smile:

3DSinghVFX: I have made the changes. It is working fine with the slicing.

OmarEmaraDev:
You are still using integer lists though, LongLists should be used.

I will continue the review later as I have an exam tomorrow.

3DSinghVFX:
> You are still using integer lists though, LongLists should be used.
> 
> I will continue the review later as I have an exam tomorrow.

Okay. I'll change that.

No problem.

animation_nodes/nodes/mesh/get_linked_vertices.py
```diff
@@ -0,0 +1,23 @@
...
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
+        self.newInput("Integer", "Vextex Index", "vertexIndex")
+
+        self.newOutput("Integer List", "Vertices", "vertexIndices")
+        self.newOutput("Integer List", "Edges", "edgeIndices")
+
+    def execute(self, mesh, vertexIndex):
+        if mesh is None:
+            return IntegerList(), IntegerList()
```
OmarEmaraDev: Those should be `LongList`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,11 +140,21 @@ cdef class Mesh:
...
+
+    def getVertexLinkedVertices(self, long vertexIndex):
+        cdef LongList neighboursAmounts, neighboursStarts, neighbours, neighbourEdges
+        neighboursAmounts, neighboursStarts, neighbours, neighbourEdges = self.getLinkedVertices()
+
+        cdef int start, end
+        start = neighboursStarts.data[vertexIndex]
+        end = start + neighboursAmounts.data[vertexIndex]
+
+        return neighbours[start : end], neighbourEdges[start : end]
```
OmarEmaraDev: Style: `[start : end]` -> `[start:end]`.

3DSinghVFX: Okay.

animation_nodes/nodes/mesh/get_linked_vertices.py
3DSinghVFX: Okay. Actually, the name of this node is changed during the discussion, that is the reason I update the CHANGELOG.md file, in the end, I mean when everything is finalized.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Changed.

animation_nodes/nodes/mesh/get_linked_vertices.py
3DSinghVFX: Changed. Updated the CHANGELOG.md file.

3DSinghVFX:
Thanks :)

PR #1278, opened 2020-01-02
Set Bevel Vertex Weight and Set Bevel Edge Weight nodes.

3DSinghVFX:
This branch has **Set Bevel Vertex Weight** and **Set Bevel Edge Weight** nodes. These nodes allow to set the bevel weight or weights for vertices and edges. These nodes take object and weight (float) /weights (float list) as inputs,
Screenshot from 2020-01-02 13-15-11

### DIFF
--- a/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
+++ b/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
@@ -20,7 +20,7 @@ class BevelEdgeWeight(bpy.types.Node, AnimationNode):
         if object is None: return None
 
         if object.type != "MESH":
-            self.raiseErrorMessage("Object is no mesh object.")
+            self.raiseErrorMessage("Object is not a mesh object.")
 
         if object.mode != "OBJECT":
             self.raiseErrorMessage("Object is not in object mode.")
--- a/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
+++ b/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
@@ -12,28 +12,16 @@ class BevelEdgeWeight(bpy.types.Node, AnimationNode):
     def create(self):
         self.newInput("Object", "Object", "object", defaultDrawType = "PROPERTY_ONLY")
         self.newInput(VectorizedSocket("Float", "useWeightList",
-            ("Weight", "weight"), ("Weights", "weights")))
+            ("Weight", "weights"), ("Weights", "weights")))
 
         self.newOutput("Object", "Object", "object")
 
-    def getExecutionFunctionName(self):
-        if self.useWeightList:
-            return "execute_WeightList"
-        else:
-            return "execute_SingleWeight"
+    def execute(self, object, weights):
+        if object is None: return None
 
-    def execute_SingleWeight(self, object, weight):
-        if object is None or object.type != "MESH" or object.mode != "OBJECT": return
-        if object.data.use_customdata_edge_bevel is False:
-            object.data.use_customdata_edge_bevel = True
+        if object.type != "MESH" or object.mode != "OBJECT":
+            self.raiseErrorMessage("Object is not in object mode or is no mesh object.")
 
-        weights = VirtualDoubleList.create(weight, 0).materialize(len(object.data.edges))
-        object.data.edges.foreach_set('bevel_weight', weights)
-        object.data.update()
-        return object
-
-    def execute_WeightList(self, object, weights):
-        if object is None or object.type != "MESH" or object.mode != "OBJECT": return
         if object.data.use_customdata_edge_bevel is False:
             object.data.use_customdata_edge_bevel = True
 
--- a/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
+++ b/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
@@ -22,7 +22,7 @@ class BevelEdgeWeight(bpy.types.Node, AnimationNode):
         if object.type != "MESH" or object.mode != "OBJECT":
             self.raiseErrorMessage("Object is not in object mode or is no mesh object.")
 
-        if object.data.use_customdata_edge_bevel is False:
+        if not object.data.use_customdata_edge_bevel:
             object.data.use_customdata_edge_bevel = True
 
         weights = VirtualDoubleList.create(weights, 0).materialize(len(object.data.edges))
--- a/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
+++ b/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
@@ -12,7 +12,7 @@ class BevelEdgeWeight(bpy.types.Node, AnimationNode):
     def create(self):
         self.newInput("Object", "Object", "object", defaultDrawType = "PROPERTY_ONLY")
         self.newInput(VectorizedSocket("Float", "useWeightList",
-            ("Weight", "weights"), ("Weights", "weights")))
+            ("Weight", "weight"), ("Weights", "weights")))
 
         self.newOutput("Object", "Object", "object")
 
--- a/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
+++ b/animation_nodes/nodes/mesh/set_bevel_edge_weight.py
@@ -1,0 +1,43 @@
+import bpy
+from ... data_structures import VirtualDoubleList
+from ... base_types import AnimationNode, VectorizedSocket
+
+class BevelEdgeWeight(bpy.types.Node, AnimationNode):
+    bl_idname = "an_SetBevelEdgeWeight"
+    bl_label = "Set Bevel Edge Weight"
+    errorHandlingType = "EXCEPTION"
+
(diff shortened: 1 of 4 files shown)


OmarEmaraDev:
I will add a review on Saturday; as I need to study for a following exam.

3DSinghVFX:
Sure :)

OmarEmaraDev:
The single and list functions are exactly the same, they can be joined to a single execution function. Notice that Virtual Lists can be created from both lists and single values.

I also want you to test the difference in performance between using virtual lists and the standard `fromValue` method for the single value case.

3DSinghVFX:
Omar, here are performance results:
`weights = VirtualDoubleList.create(weight, 0).materialize(len(object.data.vertices))`:
Case-1: Calls = 231, Min-Time = 2.5170ms 
Case-2: Calls = 100, Min-Time = 9.6709ms 
and
`weights = VirtualDoubleList.create(weight, 0)`:
Case-2: Calls = 101, Min-Time = 9.7271ms 

`weights = DoubleList.fromValue(weight, length = len(object.data.vertices))`:
 Case-1: Calls = 227, Min-Time = 2.4957ms 
Case-2: Calls = 101, Min-Time = 9.7262ms

OmarEmaraDev:
Ok, then I guess it is ok to use virtual lists here.

3DSinghVFX:
Okay.

animation_nodes/nodes/mesh/set_bevel_edge_weight.py
```diff
@@ -0,0 +1,31 @@
...
+    bl_idname = "an_SetBevelEdgeWeight"
+    bl_label = "Set Bevel Edge Weight"
+    errorHandlingType = "EXCEPTION"
+
+    useWeightList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput("Object", "Object", "object", defaultDrawType = "PROPERTY_ONLY")
+        self.newInput(VectorizedSocket("Float", "useWeightList",
+            ("Weight", "weights"), ("Weights", "weights")))
```
OmarEmaraDev: You don't have to use the same variable name, this doesn't really matter when using an execution function as opposed to execution code.

animation_nodes/ui/node_menu.py
```diff
@@ -390,6 +390,8 @@ def draw(self, context):
         insertNode(layout, "an_GetVertexColorLayerNode", "Get Vertex Color Layer")
         insertNode(layout, "an_InsertVertexColorLayerNode", "Insert Vertex Color Layer")
         insertNode(layout, "an_SetVertexColorNode", "Set Vertex Color")
+        insertNode(layout, "an_SetBevelVertexWeight", "Set Bevel Vertex Weight")
+        insertNode(layout, "an_SetBevelEdgeWeight", "Set Bevel Edge Weight")
```
OmarEmaraDev: Perhaps we should combine all of the data setters and getters into their own sub-menu.

animation_nodes/nodes/mesh/set_bevel_edge_weight.py
3DSinghVFX: Okay.
Omar, can you please tell me the nodes from which I can learn more about the execution function and execution code.

animation_nodes/ui/node_menu.py
3DSinghVFX: Okay.

3DSinghVFX: Okay.

animation_nodes/nodes/mesh/set_bevel_edge_weight.py
3DSinghVFX: https://github.com/3DSinghVFX/animation_nodes/blob/6f9437cab30c0158ef825116c13884352a6647c9/animation_nodes/nodes/material/cycles_material_output.py#L70
Another question is, the `required` we can only use when we use `yield` in a function?

OmarEmaraDev: Animation Nodes converts a node tree into python code then executes it. Each node define the code that it emits/contributes in one of two ways:
- Execution function. In this case, the emitted code is a single function call to the execution function. This is the method you have been using so far. There isn't anything special about it.
- Execution code. In this case, the emitted code is what ever returned by the `getExecutionCode` function. We use this method in two cases:
   - We want to dynamically control the code that is emitted. For instance, if we have a node with three outputs, we don't need to compute the outputs that are not connected to anything. So we selectively emit the code that computes only the outputs that are linked. This is achieved using the `required` input, which tells us which outputs are required to be computed.
   - The execution code is very small. For instance, if your node adds two numbers, then you can just emit that code and avoid the extra overhead of a function call.

The `getExecutionCode` is a "generator". A generator behaves like an iterator. Every time you `yield` something, it is like you returned a new line of code. Eventually, all of those lines are combined, so it is not really special, just a convenient way to return multiple lines of code.

Execution code introduces some complications. To import a module, you have to implement the `getUsedModules` function so that AN imports it when preparing the execution unit. Also, if you want to use an AN function from the algorithms package—for instance—you have to use the full qualifying name (`AN.algorithms.[...]`).

animation_nodes/ui/node_menu.py
```diff
@@ -408,6 +404,20 @@ def draw(self, context):
         layout.separator()
         insertNode(layout, "an_MeshObjectOutputNode", "Object Output")
 
+class MeshGettersSettersMenu(bpy.types.Menu):
```
OmarEmaraDev: I would rather we can this Mesh Data.

3DSinghVFX: Okay.

animation_nodes/nodes/mesh/set_bevel_edge_weight.py
3DSinghVFX: Omar, thank you so much for this information.

animation_nodes/ui/node_menu.py
3DSinghVFX: It is renamed.

animation_nodes/nodes/mesh/set_bevel_edge_weight.py
```diff
@@ -0,0 +1,31 @@
...
+
+        self.newOutput("Object", "Object", "object")
+
+    def execute(self, object, weights):
+        if object is None: return None
+
+        if object.type != "MESH" or object.mode != "OBJECT":
+            self.raiseErrorMessage("Object is not in object mode or is no mesh object.")
+
+        if object.data.use_customdata_edge_bevel is False:
```
OmarEmaraDev: Is there a good reason why you are checking if it is false first?

Also, the if statement should be `if not object.data.use_customdata_edge_bevel`, if it is actually needed.

3DSinghVFX: If `object.data.use_customdata_edge_bevel` is already true then we need no to set it again. 

However, we can also use this code only `object.data.use_customdata_edge_bevel = True`.

OmarEmaraDev: If it does not trigger any recalculations, then I prefer if you do it that way.

3DSinghVFX: With the if-condtion is faster, it took Min-Time `0.0258ms` less as compared to without the if-condition.

OmarEmaraDev: This seems negligible, but sure, keep the if.

animation_nodes/nodes/mesh/set_bevel_vertex_weight.py
```diff
@@ -0,0 +1,31 @@
...
+        self.newInput(VectorizedSocket("Float", "useWeightList",
+            ("Weight", "weight"), ("Weights", "weights")))
+
+        self.newOutput("Object", "Object", "object")
+
+    def execute(self, object, weights):
+        if object is None: return None
+
+        if object.type != "MESH" or object.mode != "OBJECT":
+            self.raiseErrorMessage("Object is not in object mode or is no mesh object.")
```
OmarEmaraDev: Separate those conditions into their own exceptions.

3DSinghVFX: Changed.

OmarEmaraDev: @3DSinghVFX  `Object is no mesh object.` -> `Object is not a mesh object.`

3DSinghVFX: Omar, Mesh Object Output node also has `Object is not in object mode or is no mesh object` sentence. Should I also changed that too?

OmarEmaraDev: No. We shouldn't add different things in the same pull request.

3DSinghVFX: Okay.

PR #1282, opened 2020-01-07
Fixed the error messages of the Mesh Object Output node.

3DSinghVFX:
Fixed the error messages of the Mesh Object Output node. Now, the _Object is not in object mode or is no mesh object_ message is written into their own exceptions, and all messages are ended with a period (.).

### DIFF
--- a/animation_nodes/nodes/mesh/mesh_object_output.py
+++ b/animation_nodes/nodes/mesh/mesh_object_output.py
@@ -24,11 +24,11 @@ class MeshObjectOutputNode(bpy.types.Node, AnimationNode):
         description = "", update = propertyChanged)
 
     validateMeshVerbose: BoolProperty(name = "Validate Mesh Verbose", default = False,
-        description = "Print results from validation in the console.", update = propertyChanged)
+        description = "Print results from validation in the console", update = propertyChanged)
 
     ensureAnimationData: BoolProperty(name = "Ensure Animation Data", default = True,
         description = ("Make sure that the mesh has animation data so that "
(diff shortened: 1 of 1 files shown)


3DSinghVFX:
> Property descriptions shouldn't end with a period. Blender adds those automatically and will give warnings if they exist.

Thanks for the info. Now, It is corrected.

animation_nodes/nodes/mesh/mesh_object_output.py
```diff
@@ -7,9 +7,9 @@
 from ... events import propertyChanged, executionCodeChanged
 
 meshDataTypeItems = [
-    ("MESH_DATA", "Mesh", "Mesh object that contains only vertex locations, edge indices and polygon indices", "", 0),
-    ("BMESH", "BMesh", "BMesh object", "", 1),
-    ("VERTICES", "Vertices", "A list of vertex locations; The length of this list has to be equal to the amount of vertices the mesh already has", "", 2) ]
+    ("MESH_DATA", "Mesh", "Mesh object that contains only vertex locations, edge indices and polygon indices.", "", 0),
+    ("BMESH", "BMesh", "BMesh object.", "", 1),
+    ("VERTICES", "Vertices", "A list of vertex locations; The length of this list has to be equal to the amount of vertices the mesh already has.", "", 2) ]
```
OmarEmaraDev: The same apply for enum properties.

3DSinghVFX: Okay.

PR #1283, opened 2020-01-09
The UV Map Data Input and UV Map Data Output nodes.

3DSinghVFX:
This branch has two new nodes for UV map, the **UV Map Data Input** and **UV Map Data Output** nodes,
Screenshot from 2020-01-09 18-57-20

The **UV Map Data Input** gives the coordinates of the vertices of the UV Map (correspond to the UV Map Index) of an object.

The **UV Map Data Output** sets the coordinates of the vertices of the UV Map (correspond to the UV Map Index) of an object. This node can add UV Map or UV Maps if there is no to an object.


OmarEmaraDev:
We already have support for UV Maps in the mesh data structure and the extended Blender mesh object, those facilities should be utilized instead. The workflow should be very similar to how we implemented vertex colors.

The problem lie in the fact that we don't have supports for 2D vectors beside the base structures. So we need to think about this first. This is probably why @JacquesLucke didn't implement those nodes initially.

3DSinghVFX:
Okay. I'll make the changes accordingly.

Definitely. Omar, if we can implement these four nodes then we can use other AN nodes (3D Vector, Numbers and Matrices nodes) to do other operations (transform vectors, offset vectors, etc) for 2D vectors:
1. Convert (float, float) or two float lists into 2D vectors.
2. Convert 2D vectors into (float, float) or two float lists.
3. Convert 2D vector(s) into 3D vector(s).
4. Convert 3D vector(s) into 2D vector(s).

However, if we want to implement all the nodes for 2D vectors as we have for 3D vectors then we have to discuss many things e.g., the transformation of 2D vectors, compose/decompose of the matrix for 2D vectors, evaluate falloff for 2D vectors, offset 2D vectors, etc.

3DSinghVFX:
Omar, I have modified the UV nodes. Now, they have similar structure and name as we have for the Vertex Color nodes.

OmarEmaraDev:
I will consult Jacques on the approach we should take and get back to you.

3DSinghVFX:
> I will consult Jacques on the approach we should take and get back to you.

Okay :)

OmarEmaraDev:
I have added Vector 2D and Vector 2D List sockets. You should update the patch to use those sockets. For now, implicit conversion between 2D and 3D makes the nodes usable. We will progressively add processing nodes for 2D vectors as needed.

3DSinghVFX:
Okay, I'll update the nodes.

3DSinghVFX:
Omar, I have a small query. Can I add the conversion vector nodes (2D<-->3D) in this pull request?

OmarEmaraDev:
If you mean the implicit conversion, then the code is already available.

3DSinghVFX:
Yes, it is there. If I connect the 2D-vector to 3D-vector (or vice-versa) then it does convert them accordingly, there is no problem. My concern is, in future if we removed the " implicit conversion" then it shall break the node-tree or may be I'm wrong.

OmarEmaraDev:
We won't remove the implicit conversion, we implemented it as a feature not a temporary workaround. So I wouldn't worry about that.

3DSinghVFX:
Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -130,6 +130,9 @@ cdef class Mesh:
     def getUVMapNames(self):
         return list(self.uvMaps.keys())
 
+    def getUVMapCos(self, str uvMapName):
```
OmarEmaraDev: What does `Cos` denote here?

animation_nodes/nodes/mesh/get_uv_map.py
```diff
@@ -0,0 +1,24 @@
...
+class GetUVMapNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetUVMapNode"
+    bl_label = "Get UV Map"
+    errorHandlingType = "EXCEPTION"
+
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
+        self.newInput("Text", "Name", "uvMapName")
+
+        self.newOutput("Vector 2D List", "Vectors2D", "vectors2D")
```
OmarEmaraDev: I would just call this `Vectors`.

animation_nodes/nodes/mesh/set_uv_map.py
```diff
@@ -0,0 +1,53 @@
+import bpy
+import numpy as np
```
OmarEmaraDev: Do we need numpy here?

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: UV coordinates.

animation_nodes/nodes/mesh/set_uv_map.py
3DSinghVFX: No, I'll remove that.

animation_nodes/data_structures/meshes/mesh_data.pyx
OmarEmaraDev: I see. It is not immediately clear to me, so I would consider changing it.

3DSinghVFX: Okay. Should I rename it as "UVMapVectors"?

OmarEmaraDev: Ok. Or maybe, `getUVMapPositions`.

3DSinghVFX: Okay. In the next commit, I'll change that.

OmarEmaraDev:
I commited the virtual list, you may use it now.

3DSinghVFX:
> I commited the virtual list, you may use it now.

Okay.

animation_nodes/nodes/mesh/get_uv_map.py
```diff
@@ -0,0 +1,24 @@
...
+        self.newInput("Mesh", "Mesh", "mesh")
+        self.newInput("Text", "Name", "uvMapName")
+
+        self.newOutput("Vector 2D List", "Vectors", "vectors")
+
+    def execute(self, mesh, uvMapName):
+        if uvMapName == "":
+            self.raiseErrorMessage("UV map name can't be empty.")
+
+        uvMapCos = mesh.getUVMapPositions(uvMapName)
```
OmarEmaraDev: `uvMapCos` -> `uvMapPositions`, here and else where.

3DSinghVFX: Okay.

animation_nodes/nodes/mesh/insert_uv_map.py
```diff
@@ -0,0 +1,25 @@
...
+
+class InsertUVMapNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_InsertUVMapNode"
+    bl_label = "Insert UV Map"
+    errorHandlingType = "EXCEPTION"
+
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh", dataIsModified = True)
+        self.newInput("Text", "Name", "uvMapName", value = "AN-UV Map")
+        self.newInput("Vector 2D List", "Vectors", "vectors")
```
OmarEmaraDev: This should be a vectorized socket.

animation_nodes/nodes/mesh/insert_uv_map.py
```diff
@@ -0,0 +1,25 @@
...
+
+        self.newOutput("Mesh", "Mesh", "mesh")
+
+    def execute(self, mesh, uvMapName, vectors):
+        if uvMapName == "":
+            self.raiseErrorMessage("UV map name can't be empty.")
+        elif uvMapName in mesh.getVertexColorLayerNames():
+            self.raiseErrorMessage(f"Mesh has already a uv map with the name '{uvMapName}'.")
```
OmarEmaraDev: `has already` -> `already has`.

animation_nodes/nodes/mesh/insert_uv_map.py
```diff
@@ -0,0 +1,25 @@
...
+
+        vectors = VirtualVector2DList.create(vectors, [0, 0]).materialize(len(mesh.polygons.indices))
```
OmarEmaraDev: `[0, 0]` -> `Vector((0, 0))`.

animation_nodes/nodes/mesh/insert_uv_map.py
3DSinghVFX: But we have to input the vector list, not a vector.

animation_nodes/nodes/mesh/insert_uv_map.py
3DSinghVFX: Omar, this gives me an error,
Screenshot from 2020-01-17 15-24-20

animation_nodes/nodes/mesh/insert_uv_map.py
OmarEmaraDev: Just like the vertex color nodes, since we use virtual lists, a single input will materialize to a list.

animation_nodes/nodes/mesh/insert_uv_map.py
OmarEmaraDev: Are you sure you imported `Vector`?

3DSinghVFX: It gets solved.

animation_nodes/nodes/mesh/insert_uv_map.py
3DSinghVFX: Okay.

animation_nodes/nodes/mesh/insert_uv_map.py
3DSinghVFX: Changed.

animation_nodes/nodes/mesh/get_uv_map.py
```diff
@@ -0,0 +1,24 @@
...
+class GetUVMapNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetUVMapNode"
+    bl_label = "Get UV Map"
+    errorHandlingType = "EXCEPTION"
+
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
+        self.newInput("Text", "Name", "uvMapName")
+
+        self.newOutput("Vector 2D List", "Vectors", "vectors")
```
OmarEmaraDev: Change `Vectors` to `Positions` then we can merge.

3DSinghVFX: Okay.

3DSinghVFX: Should I also change this in 'insert_uv_map.py' and 'set_uv_map.py'?

OmarEmaraDev: Yes.

OmarEmaraDev:
Please update the CHANGELOG.

3DSinghVFX:
> Please update the CHANGELOG.

Updated.

PR #1291, opened 2020-01-19
Fixed the Zero Division Error.

3DSinghVFX:
Fixed the "ZeroDivisionError" showed by animation nodes in Blender2.83,
Screenshot from 2020-01-19 21-57-27

I have found that `@cython.cdivision(True)` is not written which gave the error,
https://github.com/JacquesLucke/animation_nodes/blob/2d7ddbc18a0fcf9056c191fea12a208e4afa0f36/animation_nodes/math/vector.pyx#L120-L124

### DIFF
--- a/animation_nodes/math/vector.pyx
+++ b/animation_nodes/math/vector.pyx
@@ -120,14 +120,11 @@ cdef float dotVec3(Vector3* a, Vector3* b):
 
 @cython.cdivision(True)
 cdef float angleVec3(Vector3 *a, Vector3 *b):
+    cdef float denominator = lengthVec3(a) * lengthVec3(b)
+    if denominator == 0: return 0
+
     cdef float dot = dotVec3(a, b)
-    cdef float val, lengthVecA, lengthVecB
-    lengthVecA = lengthVec3(a)
-    lengthVecB = lengthVec3(b)
-    if lengthVecA == 0 or lengthVecB == 0:
-        val = 0
-    else:
-        val = dot / (lengthVecA * lengthVecB)
+    cdef float val = dot / denominator
     if val > 1: val = 1
     elif val < -1: val = -1
     return acos(val)
--- a/animation_nodes/math/vector.pyx
+++ b/animation_nodes/math/vector.pyx
@@ -118,6 +118,7 @@ cdef float distanceSquaredVec3(Vector3* a, Vector3* b):
 cdef float dotVec3(Vector3* a, Vector3* b):
     return a.x * b.x + a.y * b.y + a.z * b.z
 
+@cython.cdivision(True)
 cdef float angleVec3(Vector3 *a, Vector3 *b):
     cdef float dot = dotVec3(a, b)
     cdef float val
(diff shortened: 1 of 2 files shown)


OmarEmaraDev:
`@cython.cdivision(True)` removes the zero division checks that Cython adds. We add this directive to make the division faster and avoid the overhead of zero division checks.

So you didn't really avoid the zero division here. Keep the directive but add an explicit check for zero division.

OmarEmaraDev:
Which call to `angleVec3` resulted in the zero division?

3DSinghVFX:
> Which call to `angleVec3` resulted in the zero division?

Omar, I have tried with the same blend (and other files) but not able to regenerate the error.

3DSinghVFX:
> Which call to `angleVec3` resulted in the zero division?

Mesh From Spline node give the error when I trim the spline,
Screenshot from 2020-01-20 18-04-04

OmarEmaraDev:
Ok.

3DSinghVFX:
When I commented this line `matrices = getMatricesAlongSpline(spline, amount, self.parameterType)` in the Mesh From Spline node then error disappeared.

3DSinghVFX:
> `@cython.cdivision(True)` removes the zero division checks that Cython adds. We add this directive to make the division faster and avoid the overhead of zero division checks.

Okay.

OmarEmaraDev:
Will you continue working on this patch?

3DSinghVFX:
Omar, I have tried but I did not find anything related to `angleVec3` in the `matrices = getMatricesAlongSpline(spline, amount, self.parameterType)`.

OmarEmaraDev:
`getMatricesAlongSpline` -> `ensureNormals` -> `calculateNormalsForTangents` -> `calcNextNormal` -> `angleVec3`.

In this patch, you just need to add an explicit check for zero.

3DSinghVFX:
Okay, Thanks :)

OmarEmaraDev:
I actually meant in the `angleVec3` function.

3DSinghVFX:
Okay, I'll change that.

animation_nodes/math/vector.pyx
```diff
@@ -118,10 +118,16 @@ cdef float distanceSquaredVec3(Vector3* a, Vector3* b):
...
     return a.x * b.x + a.y * b.y + a.z * b.z
 
+@cython.cdivision(True)
 cdef float angleVec3(Vector3 *a, Vector3 *b):
     cdef float dot = dotVec3(a, b)
-    cdef float val
-    val = dot / (lengthVec3(a) * lengthVec3(b))
+    cdef float val, lengthVecA, lengthVecB
+    lengthVecA = lengthVec3(a)
+    lengthVecB = lengthVec3(b)
```
OmarEmaraDev: I think you should just assign `lengthVec3(a) * lengthVec3(b)` to a single variable, then check for zero, returning zero directly.

3DSinghVFX: Okay.

3DSinghVFX: Changed.

PR #1294, opened 2020-01-26
Refactor Auto Execution mechanism.

OmarEmaraDev:
# Description

This patch solves two issues:
- Crashes upon rendering or exporting scenes.
- Inconsistency between scenes and rendered/exported scenes.

Crashes were solved by only allowing executions to happen inside `frame_changed_post` application handlers. Any other mean of execution is forbidden until the render is done.

The inconsistency was solved by passing the `DAG_EVAL_RENDER` dependency graph we get from the `frame_changed_post` handler to the execution units, then using evaluated IDs when appropriate. The dependency graph is stored in the `events.evaluatedDepsgraph` global variable, it is invalidated after execution by setting it to `None`.

View layers are no longer updated after execution, instead, the update is deferred until it is actually needed. That's because view layers can't be updated while the dependency graph is being evaluated, this caused some of the renders and exports to fail.

The *Frame Changed* option now allow execution even if the frame didn't actually change. This guarantee correct renders and exports. The scene could have been in any state while being at the same frame.

# Results

So far, all exports and renders I tried had no crashes or inconsistency. Though I am not sure if it works in all cases, so I guess we will just have to wait for use feedback.

# Concerns And To Dos

The execution restrictions were only applied on renders, not exports. Because there is no way to reliably tell if we are exporting a scene or not. For that reason, it is recommended that users turn off all other execution options except for *Frame Changed* option until this is resolved.

Some nodes may still need to be updated to use evaluated IDs. I updated some of them, but not sure which of the other nodes needs to be updated as well.


animation_nodes/execution/auto_execution.py
```diff
@@ -15,10 +13,6 @@ def executeNodeTrees(nodeTrees):
         nodeTree.autoExecute()
 
 def afterExecution():
-    for scene in bpy.data.scenes:
-        for viewLayer in scene.view_layers:
-            viewLayer.update()
```
JacquesLucke: Do objects that are parented to objects animated by AN still work?

OmarEmaraDev: Yes, we read from evaluated IDs and write to original IDs.
I will test the parenting.

OmarEmaraDev: Yes, parents works as expected.

OmarEmaraDev:
@JacquesLucke I talked to Sybren about adding callbacks for exporters and explained the situation. He said I should make a proposal and it will be considered. I just wrote a test implementation and it seem to work. So, hopefully, our concerns with regards to exporters should be resolved.

JacquesLucke:
Great! I can't see any obvious errors in this patch. I think you probably tested this quite a bit better than I could currently. Let's get more users to test it by merging it in master.

PR #1297, opened 2020-01-29
Added the grease pencil nodes.

3DSinghVFX:
This branch has the essential grease pencil nodes.

The grease pencil object store data in the following order,
Input order: `Object > Layer(s) > Frame(s) > Stroke(s) > Point(s)`
Output order: `Point(s) > Stroke(s) > Frame(s) > Layer(s) > Object`

1. **Layer**: It has information about the _frames_, opacity, blend mode (e.g., overlay, add, etc.), pass index, thickness, etc. A layer can have one frame or multiple frames.
2. **Frame**: Frame contains info about frame-number (time) and stroke(s). The important thing is each Frame has unique frame-number or we can say frame-number is the fingerprint of Frame.
3. **Stroke**: Stroke contains info about the Point(s) that make the stroke, and line width, material index, display mode, cyclic, etc.
4. **Point(s)**: A Point(is like vertex) in 3D space that has its coordinate, strength, pressure, UV-rotation.

Nodes to get grease pencil data:
1. The **GPencil Object Input** node get the layer by name (or all layers) from the object,
Screenshot from 2020-01-29 14-08-50
2. Now, we have the layer. The **GPencil Layer Input** node gives frames, frame-numbers and layer name, blend mode, opacity, pass index. Because a layer can have many frames, so to get an individual frame, this node has four options: Active Frame, Frame Index(Indices), Frame Number (time), and All Frames, (Default)
Screenshot from 2020-01-29 15-00-26
(Extended):
Screenshot from 2020-01-29 15-01-53
3. The **GPencil Frame Input** node gives all strokes or stroke(s) based on the index(indices) and also the frame-number,
Screenshot from 2020-01-29 14-19-33
4. Because a stroke is a collection of Points (have coordinates, strength, pressure, etc.), so to make node-tree compact, the **GPencil Stroke Input** node directly gives points(coordinates), strengths, pressures, UV-rotations, line width, cyclic, start cap(Flat or Round), end cap (Flat or Round), material index, 
Screenshot from 2020-01-29 14-25-58

5. Here is complete node-tree to get grease-pencil data,
Screenshot from 2020-01-29 14-28-23
 
Nodes to set grease pencil data:
1. The **GPencil Stroke Output** node generates stroke from a vector list with default values for other inputs(strength, pressure, etc.). However, this node also allows setting strength, pressure, UV-rotation per Point,
Screenshot from 2020-01-29 14-35-48
2. With the **GPencil Frame Output** node, we can put all the strokes onto a single Frame or separate Frames. As I said frame must have unique frame-nimber. So, in case of strokes onto separate Frames, this node checks whether each Frame has a unique frame-number or not,
Screenshot from 2020-01-29 14-41-47
3. Next is **GPencil Layer Output** node, this node can take input as Frame or Frame List, layer name, blend mode, opacity, pass index. It is possible to input Frame List, so this node also make sure that each Frame has unique frame-number,
Screenshot from 2020-01-29 14-50-53
4. The **GPencil Object Output** node can take a layer or layer list. This node generates gpencil object with a layer, frame, empty stroke, and material (which is essential to view the stroke),
Screenshot from 2020-01-29 14-54-33
5. Here is complete node-tree to set grease-pencil data,
Screenshot from 2020-01-29 14-58-56

Blend File:
[GPencilNodes.blend.zip](https://github.com/3DSinghVFX/animation_nodes/files/4127207/GPencilNodes.blend.zip)


OmarEmaraDev:
I will finish the thing at hand and start the review.

3DSinghVFX:
Okay :)

OmarEmaraDev:
Is there a reason why you created the data structures as Cython structures?

3DSinghVFX:
> Is there a reason why you created the data structures as Cython structures?

1. To implement the method for vertex weight.
2. To implement methods for vertex color but it is not in master yet.

However, for the current GPencil nodes, it does not need to be in Cython.

OmarEmaraDev:
What will those methods do?

3DSinghVFX:
To set weights and colors for vertices.

OmarEmaraDev:
Ok. We will get back to this later.

animation_nodes/data_structures/__init__.py
```diff
@@ -3,6 +3,9 @@ def importDataStructures():
     from . struct import ANStruct
 
     from . meshes.mesh_data import Mesh
+    from . gpencils.gpencil_data import Layer
+    from . gpencils.gpencil_data import Stroke
+    from . gpencils.gpencil_data import Frame
```
OmarEmaraDev: I think the names should be prefixed with `GP`, as the names are a bit ambiguous right now.

animation_nodes/data_structures/gpencils/gpencil_data.pyx
```diff
@@ -0,0 +1,98 @@
...
+        return textwrap.dedent(
+            f"""AN Stroke Object:
+            Points: {len(self.vertices)}
+            Material Index: {self.material_index}
+            Display Mode: {self.display_mode}""")
+
+    def copy(self):
+        return Stroke(self.vertices, self.strength, self.pressure, self.uv_rotation,
+        self.line_width, self.draw_cyclic, self.start_cap_mode, self.end_cap_mode,
+        self.material_index, self.display_mode)
```
OmarEmaraDev: Separate those classes to their own files.

3DSinghVFX:
Okay. Another point is I used NumPy to check a unique frame-number (in GPencil Frame Output and GPencil Layer Output nodes) which can be replaced by a method.

animation_nodes/data_structures/__init__.py
3DSinghVFX: Okay. Should I also change the name of the socket files?

animation_nodes/data_structures/gpencils/gpencil_data.pyx
3DSinghVFX: Okay

animation_nodes/data_structures/__init__.py
OmarEmaraDev: They are in their own directory, so that won't be necessary.

3DSinghVFX: Okay.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
```diff
@@ -0,0 +1,42 @@
...
+        return textwrap.dedent(
+            f"""AN Stroke Object:
+            Points: {len(self.vertices)}
+            Material Index: {self.material_index}
+            Display Mode: {self.display_mode}""")
+
+    def copy(self):
+        return GPStroke(self.vertices, self.strength, self.pressure, self.uv_rotation,
+        self.line_width, self.draw_cyclic, self.start_cap_mode, self.end_cap_mode,
+        self.material_index, self.display_mode)
```
OmarEmaraDev: Notice that this is a shallow copy. You need to copy the individual components as well. `GPStroke(self.vertices.copy(), ...`. The same goes for other structures.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
```diff
@@ -0,0 +1,42 @@
...
+        self.vertices = vertices
+        self.strength = strength
+        self.pressure = pressure
+        self.uv_rotation = uv_rotation
+        self.line_width = line_width
+        self.draw_cyclic = draw_cyclic
+        self.start_cap_mode = start_cap_mode
+        self.end_cap_mode = end_cap_mode
+        self.material_index = material_index
+        self.display_mode = display_mode
```
OmarEmaraDev: Use camel case. `uv_rotation` -> `uvRotation`. The same for other attributes.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
```diff
@@ -0,0 +1,42 @@
...
+        if draw_cyclic is None: draw_cyclic = False
+        if start_cap_mode is None: start_cap_mode = "ROUND"
+        if end_cap_mode is None: end_cap_mode = "ROUND"
+        if material_index is None: material_index = 0
+        if display_mode is None: display_mode = "SCREEN"
+
+        self.vertices = vertices
+        self.strength = strength
+        self.pressure = pressure
+        self.uv_rotation = uv_rotation
```
OmarEmaraDev: Use plural nouns. `stength` -> `strengths`. The same for all list attributes.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
3DSinghVFX: Okay.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
3DSinghVFX: Okay.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
3DSinghVFX: Okay.

animation_nodes/data_structures/gpencils/gpencil_stroke_data.pyx
3DSinghVFX: Is it possible to add `copy()` to the `str or int` variables, e.g.,
https://github.com/JacquesLucke/animation_nodes/blob/715d665fe5df53800d8a4255967bef2b60fc3cf8/animation_nodes/data_structures/gpencils/gpencil_layer_data.pyx#L37-L38

OmarEmaraDev: You don't need to copy those, just referenced structures like lists.

3DSinghVFX: Okay.

OmarEmaraDev:
I would go with:
- GP Layer Info.
- GP Frame Info.
- GP Stroke Info.
- GP Layer From Frames.
- GP Frame From Strokes.
- GP Strokes From Points.

If you agree, update the patch with those names.

3DSinghVFX:
We can change the `GPencil > GP` as a prefix for nodes.

I used the `Input` and `Output` suffix because that we have in Mesh nodes as well. I think this convesion is fine,
- GP {Layer, Frame, Stroke} Info.
- GP {Layer, Frame, Stroke} From {Frames, Strokes, Points}.

OmarEmaraDev:
Looks good.

animation_nodes/data_structures/gpencils/gpencil_frame_data.pyx
```diff
@@ -0,0 +1,23 @@
...
+
+    def __repr__(self):
+        return textwrap.dedent(
+        f"""AN Frame Object:
+        Strokes: {len(self.strokes)}
+        Frame Index: {self.frameIndex}
+        Frame Number: {self.frameNumber}""")
+
+    def copy(self):
+        return GPFrame(self.strokes.copy(), self.frameIndex, self.frameNumber)
```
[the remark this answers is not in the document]
3DSinghVFX: Do you mean something like this,
```
def copy(self):
        newStrokes = []
        for stroke in self.strokes:
            newStrokes.append(stroke.copy())
        return GPFrame(newStrokes, self.frameIndex, self.frameNumber)
```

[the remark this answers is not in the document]
OmarEmaraDev: Or using a list comprehension:
```
GPFrame([stroke.copy() for stroke in self.strokes], self.frameIndex, self.frameNumber)
```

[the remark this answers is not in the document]
3DSinghVFX: Okay.

3DSinghVFX:
Should I also change the name convention in the code of the nodes, e.g., 
https://github.com/JacquesLucke/animation_nodes/blob/a0ac04c3158c5cf32033f89c7db80cbba53766fe/animation_nodes/nodes/gpencil/gpencil_frame_input.py#L23

to sth like this `self.newInput("GPFrame", "GP Frame", "gpFrame", dataIsModified = True)`?

OmarEmaraDev:
No, that won't be necessary. It is understood from the context.

3DSinghVFX:
Okay :)

animation_nodes/data_structures/gpencils/gp_layer_data.py
```diff
@@ -0,0 +1,39 @@
+# cython: profile=True
```
OmarEmaraDev: No need for those lines.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
```diff
@@ -0,0 +1,49 @@
...
+    def execute_StrokesFrameNumber(self, strokes, frameNumber):
+        if not self.useStrokeList:
+            strokes = [strokes]
+
+        return GPFrame(strokes, 0, frameNumber)
+
+    def execute_StrokesFrameNumbers(self, strokes, frameNumbers):
+        frames = []
+        if not self.useStrokeList:
+            return frames
```
OmarEmaraDev: Note that you can add multiple vectorized socket properties per socket like we did before. This should remove the case where `self.useNumberList` is true and `self.useStrokeList` is not.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if len(strokes) > len(frameNumbers):
```
OmarEmaraDev: What if it was less than?

animation_nodes/nodes/gpencil/gp_frame_info.py
```diff
@@ -0,0 +1,68 @@
...
+    bl_label = "GP Frame Info"
+    errorHandlingType = "EXCEPTION"
+
+    strokeType: EnumProperty(name = "Stroke Type", default = "ALL",
+        items = strokeTypeItems, update = AnimationNode.refresh)
+
+    useIntegerList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput("GPFrame", "Frame", "frame", dataIsModified = True)
```
OmarEmaraDev: Is `dataIsModified` really needed here?

animation_nodes/nodes/gpencil/gp_frame_info.py
```diff
@@ -0,0 +1,68 @@
...
+        outStrokes = []
+        if len(frame.strokes) == 0:
+            return outStrokes, frame.frameNumber
+
+        for index in strokeIndices:
+            outStrokes.append(self.getStroke(strokes, index))
+        return outStrokes, frame.frameNumber
+
+    def getStroke(self, strokes, strokeIndex):
+        try: return strokes[strokeIndex]
```
OmarEmaraDev: Don't use exceptions if you can write defensive code. In this case, you can easily check if the index indeed exist.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,141 @@
...
+        elif self.frameType == "ALL":
+            return "execute_AllFrames"
+
+    def execute_ActiveFrame(self, layer, scene):
+        frames = layer.frames
+        if len(layer.frames) == 0:
+            return GPFrame(), 0, 0, layer.layerName, layer.blendMode, layer.opacity, layer.passIndex
+        frameNumbers = layer.frameNumbers
+        activeIndex = self.getActiveFrameIndex(scene.frame_current, frameNumbers)
+        return frames[activeIndex], activeIndex, frameNumbers[activeIndex], layer.layerName,\
```
OmarEmaraDev: You can put those in parentheses to avoid the `\` at the end.

animation_nodes/nodes/gpencil/gp_object_input.py
```diff
@@ -0,0 +1,111 @@
...
+            return "executeSingle"
+        elif self.layerType == "ALL":
+            return "executeList"
+
+    def executeSingle(self, object, useWorldSpace, layerName):
+        if object is None:
+            return GPLayer()
+
+        if useWorldSpace:
+            worldMatrix = object.matrix_world
```
OmarEmaraDev: You have to get the matrix from an evaluated object. See 081819d for how to do that.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,114 @@
...
+            ("Layer", "layer"), ("Layers", "layers")))
+        self.newOutput("Object", "Object", "object")
+
+    def getExecutionFunctionName(self):
+        if self.useLayerList:
+            return "execute_LayerList"
+        else:
+            return "execute_Layer"
+
+    def execute_Layer(self, object, layer):
```
OmarEmaraDev: You seem to have a lot of code duplication between `execute_Layer` and `execute_LayerList`. Can you refactor that?

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
```diff
@@ -0,0 +1,58 @@
...
+        self.newInput("Boolean", "Start Cap", "startCapMode", value = False)
+        self.newInput("Boolean", "End Cap", "endCapMode", value = False)
+        self.newInput("Integer", "Material Index", "materialIndex", value = 0)
+        self.newOutput("GPStroke", "Stroke", "stroke")
+
+        visibleInputs = ("Points", "Strength", "Strengths", "Pressure", "Pressures", "Line Width", "Material Index")
+        for socket in self.inputs:
+            socket.hide = socket.name not in visibleInputs
+
+    def execute(self, vertices, strengths, pressures, uvRotations, lineWidth, drawCyclic, startCapMode, endCapMode, materialIndex):
```
OmarEmaraDev: Try to wrap long lines.

animation_nodes/nodes/gpencil/gp_stroke_info.py
```diff
@@ -0,0 +1,40 @@
...
+        self.newOutput("Float List", "UV-Rotations", "uvRotations")
+        self.newOutput("Float", "Line Width", "lineWidth")
+        self.newOutput("Boolean", "Cyclic", "drawCyclic")
+        self.newOutput("Boolean", "Start Cap", "startCapMode")
+        self.newOutput("Boolean", "End Cap", "endCapMode")
+        self.newOutput("Integer", "Material Index", "materialIndex")
+
+        visibleOutputs = ("Points", "Strengths", "Pressures", "Line Width", "Material Index")
+        for socket in self.outputs:
+            socket.hide = socket.name not in visibleOutputs
```
OmarEmaraDev: I would just hide theme in the `newOutput` call.

animation_nodes/data_structures/gpencils/gp_layer_data.py
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
3DSinghVFX: This is the one to one case i.e., each stroke has its own frame. However, if it is less than then we can assign the rest of the stroke on to the last frame.

animation_nodes/nodes/gpencil/gp_frame_info.py
3DSinghVFX: It is not needed. I'll remove.

animation_nodes/nodes/gpencil/gp_frame_info.py
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: Okay.

3DSinghVFX:
> Excuse the slow review, I am still trying to understand GP.
> 
>     * Why do we store the frame numbers in the GP layer? We already store the frame number inside the GP frame object.
> 
>     * Why do we store the frame index in the GP frame? Does this provide any real information? Does the list need to be sorted or something?
> 
>     * Why are _Start Cap_ and _End Cap_ booleans? I would expect them to be enums represented by strings as we have in the rest of AN.

It is fine, I can understand that.

*To check each frame has a unique frame-number we need frame-number from all frames and also to get Active-Frame. So, to avoid loop to get frame-number, I store the frame-number in the gp-layer. 

*I also thought to sort them but frames may not have frame-index according to frame-numbers (e.g., Frames[0] has frame-number 20 and Frames[1] has frame-number 1). But we do not need to sort them. However, In the case of Layer Info (Inputs) node, frame index servers as the serial number of the frame which sometimes handy to know which frame we are getting (especially in the case of Frame Number and Active Frame of the Layer Info node).

*I thought it easy to set a stroke is _Round_ or _Flat_ than typing the text.

animation_nodes/nodes/gpencil/gp_frame_info.py
3DSinghVFX: Yes, it is needed for operation on the stroke level.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
3DSinghVFX: Omar, I have changed the code and it worked,
```
def create(self):
        self.newInput(VectorizedSocket("GPStroke", "useStrokeList",
            ("Stroke", "stroke"), ("Strokes", "strokes")), dataIsModified = True)
        self.newInput(VectorizedSocket("Float", ["useStrokeList", "useNumberList"],
            ("Frame Number", "frameNumber"), ("Frame Numbers", "frameNumbers")))
        self.newOutput(VectorizedSocket("GPFrame", "useNumberList",
            ("Frame", "frame"), ("Frames", "frames")))
```
However, if I suddenly cut the input (stokes and frame-numbers) then output does not fall back to _Frame_ . However, it remains _Frames_ which cause problem in _GP Layer From Frames_ node give an error,
Screenshot from 2020-02-01 00-31-31
Screenshot from 2020-02-01 00-30-56

3DSinghVFX: If I use this code then the output fall back to _Frame_  and the above error does not occur,
```
def create(self):
        self.newInput(VectorizedSocket("GPStroke", "useStrokeList",
            ("Stroke", "stroke"), ("Strokes", "strokes")), dataIsModified = True)
        self.newInput(VectorizedSocket("Float", ["useStrokeList", "useNumberList"],
            ("Frame Number", "frameNumber"), ("Frame Numbers", "frameNumbers")))
        if self.useNumberList:
            self.newOutput("GPFrame List", "Frames", "frames")
        else:
            self.newOutput("GPFrame", "Frame", "frame")
```

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
OmarEmaraDev: If I understand the function correctly. If the number of strokes is less than the number of frames. The rest of the frames will be ignored. Which doesn't seem predictable enough. So you should either use virtual lists here or raise an error.

3DSinghVFX: Okay, here I'll raise an error.

OmarEmaraDev:
- Well, you will loop either way. I don't think this will affect performance very much. You may try it. So I think the frame numbers list should be removed from the layer.
- But you said it yourself. It doesn't make sense to keep an index of something that is not ordered. What if some frames were removed? Now this index is useless.
- If you want to make it a boolean. Change the name to something that would be more expressive, like `Is Start Cap Round`. But the problem is, more modes may be added in future, so we will eventually have to change that to an enum represented by a string. Plus, a simple compare and switch nodes can be used. So it is not that inconvenient.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
OmarEmaraDev: I can't replicate the issue. But I will take a look at the vectorization code, see if something is up.

3DSinghVFX:
>     * ut you said it yourself. It doesn't make sense to keep an index of something that is not ordered. What if some frames were removed? Now this index is useless.

I agree with all your points, I'll make the changes.

3DSinghVFX:
Omar, When I'm typing the Layer Name (Text input) which add the unwanted layers,
Screenshot from 2020-02-01 14-33-58


Is there a way that text input only updates when I finished the typing?

animation_nodes/nodes/gpencil/gp_object_input.py
3DSinghVFX: It is done.

animation_nodes/nodes/gpencil/gp_object_output.py
3DSinghVFX: It is changed.

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
3DSinghVFX: Done.

animation_nodes/nodes/gpencil/gp_stroke_info.py
3DSinghVFX: Changed.

OmarEmaraDev:
It updates due to the `{"TEXTEDIT_UPDATE"}` option in the string property in the text socket. I am not sure if we can overwrite that already. I will take a look.

3DSinghVFX:
Okay.

animation_nodes/nodes/gpencil/gp_frame_info.py
```diff
@@ -0,0 +1,69 @@
...
+                return "execute_StrokeIndices"
+            else:
+                return "execute_StrokeIndex"
+
+    def execute_AllStrokes(self, frame):
+        return frame.strokes, frame.frameNumber
+
+    def execute_StrokeIndex(self, frame, strokeIndex):
+        strokes = frame.strokes
+        if len(frame.strokes) == 0:
```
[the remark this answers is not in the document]
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_frame_info.py
```diff
@@ -0,0 +1,69 @@
...
+        return self.getStroke(strokes, strokeIndex), frame.frameNumber
+
+    def execute_StrokeIndices(self, frame, strokeIndices):
+        strokes = frame.strokes
+        outStrokes = []
+        if len(frame.strokes) == 0:
+            return outStrokes, frame.frameNumber
+
+        for index in strokeIndices:
+            outStrokes.append(self.getStroke(strokes, index))
```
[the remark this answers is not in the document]
3DSinghVFX: Done.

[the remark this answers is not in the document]
3DSinghVFX: Done.

animation_nodes/nodes/gpencil/gp_object_input.py
```diff
@@ -0,0 +1,111 @@
...
+            gpencilLayers.append(GPLayer(layer.info, self.getFrames(layer, worldMatrix), layer.blend_mode,
+                                         layer.opacity, layer.pass_index))
+        return gpencilLayers
+
+    def getLayer(self, object, layerName):
+        if object.type != "GPENCIL":
+            self.raiseErrorMessage("Object is not a grease pencil object.")
+        if object.mode != "OBJECT":
+            self.raiseErrorMessage("Object is not in object mode.")
+        try: return object.data.layers[layerName]
```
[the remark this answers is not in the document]
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_object_input.py
```diff
@@ -0,0 +1,111 @@
...
+        frames = []
+        for frame in layerFrames:
+            frames.append(GPFrame(self.getStrokes(frame.strokes, matrixWorld), frame.frame_number))
+        return frames
+
+    def getStrokes(self, strokes, matrixWorld):
+        newStrokes = []
+        for stroke in strokes:
+            newStrokes.append(self.getStroke(stroke, matrixWorld))
+        return newStrokes
```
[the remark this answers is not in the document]
3DSinghVFX: Replaced here and also for frames.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,87 @@
...
+        gpencilStroke.material_index = stroke.materialIndex
+        gpencilStroke.display_mode = stroke.displayMode
+        gpencilStroke.draw_cyclic = stroke.drawCyclic
+        gpencilStroke.start_cap_mode = stroke.startCapMode
+        gpencilStroke.end_cap_mode = stroke.endCapMode
+        return gpencilStroke
+
+    def getLayer(self, gpencil, layer):
+        layerName = layer.layerName
+        try: gpencilLayer = gpencil.layers[layerName]
```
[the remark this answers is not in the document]
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def execute(self, frames, layerName, blendMode, opacity, passIndex):
+        if not self.useFrameList:
+            frames = [frames]
+
+        frameNumbers = LongList(length = len(frames))
+        for i, frame in enumerate(frames):
+            frameNumbers[i] = frame.frameNumber
+        if len(np.unique(frameNumbers)) != len(frameNumbers):
+            self.raiseErrorMessage("Some Frame Numbers are repeated.")
```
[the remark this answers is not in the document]
3DSinghVFX: Done.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
```diff
@@ -0,0 +1,49 @@
...
+            self.raiseErrorMessage("Invalid Frame Numbers list.")
+
+        if len(np.unique(frameNumbers)) != len(frameNumbers):
+            self.raiseErrorMessage("Some Frame Numbers are repeated.")
+
+        frames = []
+        for i, stroke in enumerate(strokes):
+            frames.append(GPFrame([stroke], frameNumbers[i]))
+        return frames
```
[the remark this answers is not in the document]
3DSinghVFX: Omar can please explain this, I do not find the way to do that. I'm trying this 'GPFrame(zip(strokes, frameNumbers))'?

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
```diff
@@ -0,0 +1,49 @@
...
+            strokes = [strokes]
+
+        return GPFrame(strokes, frameNumber)
+
+    def execute_StrokesFrameNumbers(self, strokes, frameNumbers):
+        if len(strokes) < len(frameNumbers):
+            self.raiseErrorMessage("Invalid Stroke list.")
+
+        if len(strokes) > len(frameNumbers):
+            self.raiseErrorMessage("Invalid Frame Numbers list.")
```
OmarEmaraDev: I would join both errors and raise `"Strokes and Frame Numbers have different lengths."`

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
```diff
@@ -0,0 +1,49 @@
...
+
+        if len(np.unique(frameNumbers)) != len(frameNumbers):
```
OmarEmaraDev: `np.unique(frameNumbers.asNumpyArray())`

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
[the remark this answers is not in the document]
OmarEmaraDev: `zip` will combine multiple lists into tuples element-wise. So for instance `zip([0, 1, 2], [3, 4, 5])` will be:

```
[(0, 3),
 (1, 4),
 (2, 5)]
``` 

While looping, if you write multiple looping variables, each variable will be assigned one of the elements of the tuples. So your expression can something like this:

`[GPFrame([stroke], number) for stroke, number in zip(strokes, frameNumbers))]`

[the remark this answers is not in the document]
OmarEmaraDev: Is there a reason why frameNumbers is a float list instead of an integer list?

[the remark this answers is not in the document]
3DSinghVFX: Blender has an option for Subframes. So, if someone have Frame at the subframe then we need fameNumber(s) as Float (list).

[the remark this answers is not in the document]
OmarEmaraDev: Blender have subframes, but does it allow us to insert subframes?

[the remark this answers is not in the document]
OmarEmaraDev: Ok. I can see a mention of float frames in the API. You are good.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
3DSinghVFX: Replaced.

animation_nodes/nodes/gpencil/gp_frame_info.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def execute_StrokeIndex(self, frame, strokeIndex):
+        strokes = frame.strokes
+        if len(strokes) == 0:
+            return GPStroke(), frame.frameNumber
+        return self.getStroke(strokes, strokeIndex), frame.frameNumber
+
+    def execute_StrokeIndices(self, frame, strokeIndices):
+        strokes = frame.strokes
+        outStrokes = []
```
OmarEmaraDev: This is not needed. Just return `[]` directly.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
```diff
@@ -0,0 +1,34 @@
...
+        self.newInput("Float", "Opacity", "opacity", value = 1, hide = True)
+        self.newInput("Integer", "Pass Index", "passIndex", value = 0, hide = True)
+        self.newOutput("GPLayer", "Layer", "layer")
+
+    def execute(self, frames, layerName, blendMode, opacity, passIndex):
+        if not self.useFrameList:
+            frames = [frames]
+
+        frameNumbers = [frame.frameNumber for frame in frames]
+        if len(np.unique(frameNumbers)) != len(frameNumbers):
```
OmarEmaraDev: Use `len(frameNumbers) != len(set(frameNumbers))`

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,150 @@
...
+        elif self.frameType == "ALL":
+            return "execute_AllFrames"
+
+    def execute_ActiveFrame(self, layer, scene):
+        frames = layer.frames
+        if len(layer.frames) == 0:
+            return GPFrame(), 0, layer.layerName, layer.blendMode, layer.opacity, layer.passIndex
+
+        frameNumbers = self.getFrameNumbers(frames)
+        activeIndex = self.getActiveFrameIndex(scene.frame_current, frameNumbers)
```
OmarEmaraDev: You should have a single `getActiveFrame` method. It returns the active frame directly. The output frame number can then be computed from the frame itself.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,150 @@
...
+            if self.useFrameList:
+                return "execute_FrameNumbers"
+            else:
+                return "execute_FrameNumber"
+        elif self.frameType == "ALL":
+            return "execute_AllFrames"
+
+    def execute_ActiveFrame(self, layer, scene):
+        frames = layer.frames
+        if len(layer.frames) == 0:
```
OmarEmaraDev: Use `frames` here directly. The same goes for the other methods.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,150 @@
...
+
+    useFrameList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput("GPLayer", "Layer", "layer", dataIsModified = True)
+
+        if self.frameType == "ACTIVE":
+            self.newInput("Scene", "Scene", "scene", hide = True)
+            self.newOutput("GPFrame", "Frame", "frame")
+            self.newOutput("Float", "Frame Number", "frameNumber")
```
OmarEmaraDev: I would say the Frame Number doesn't belong here. If the user wants to get it, it can be computed using the Frame Info node.

3DSinghVFX: Should I remove Frame Number from all the cases?

OmarEmaraDev: Yes.

3DSinghVFX: Okay

animation_nodes/nodes/gpencil/gp_frame_info.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
3DSinghVFX: Done.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: Replaced.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: Removed.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def execute_AllFrames(self, layer):
+        frames = layer.frames
+        return frames, layer.layerName, layer.blendMode, layer.opacity, layer.passIndex
+
+    def getActiveFrame(self, currentFrame, frames):
+        frameNumbers = self.getFrameNumbers(frames)
+        if currentFrame <= frameNumbers.getMinValue(): return frames[0]
+        allFrames = list(filter(lambda x: x <= currentFrame, frameNumbers))
+        return frames[frameNumbers.index(allFrames[-1])]
```
OmarEmaraDev: I think this is simpler and faster.
```python
min(frame for frame in frames if frame.frameNumber >= currentFrame, key = lambda f : f.frameNumber)
```

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def getFrameNumbers(self, frames):
+        frameNumbers = LongList(length = len(frames))
+        for i, frame in enumerate(frames):
+            frameNumbers[i] = frame.frameNumber
+        return frameNumbers
```
OmarEmaraDev: This function can be inlined. And a list comprehension can be used.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def getFrameNumberIndex(self, frameNumbers, frameNumber):
+        if frameNumber not in set(frameNumbers):
```
OmarEmaraDev: No need to construct a set here.

animation_nodes/nodes/gpencil/gp_object_input.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def executeList(self, object, useWorldSpace):
+        if object is None:
+            return []
+
+        evaluatedObject = getEvaluatedID(object)
+        if useWorldSpace:
+            worldMatrix = evaluatedObject.matrix_world
+        else:
+            worldMatrix = None
```
OmarEmaraDev: Just pass the evaluated object and do the check there.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: Omar this expression gave me an error. Can you please explain briefly what is going on in this line?

OmarEmaraDev: Sorry, my bad. We want to find the frame with the largest frame number that is less than or equal the current frame. So we first filter the frames whose frame number is less than or equal the current frame, we do that using the generator:

```python
(frame for frame in frames if frame.frameNumber <= currentFrame)
```

To get the frame which has maximum frame number, we just use max with the `frameNumber` as key. So the expression is: 

```python
max((frame for frame in frames if frame.frameNumber <= currentFrame), key = lambda f : f.frameNumber)
```

3DSinghVFX: Omar, the above code works fine when `currentFrame >= frames[0].frameNumber`, otherwise it gives error because sequese(in max function) becomes empty. Here is workaround,
```
def getActiveFrame(self, currentFrame, frames):
        frameNumbers = self.getFrameNumbers(frames)
        if currentFrame <= min(frameNumbers): return frames[0]
        return max((frame for frame in frames if frame.frameNumber <= currentFrame),
                    key = lambda frame : frame.frameNumber)
```
or
```
def getActiveFrame(self, currentFrame, frames):
        try: return max((frame for frame in frames if frame.frameNumber <= currentFrame),
                    key = lambda frame : frame.frameNumber)
        except: frames[0]
```

OmarEmaraDev: The max function can take a default value, I think.
https://docs.python.org/3/library/functions.html#max

3DSinghVFX: Thanks, it worked.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
```diff
@@ -0,0 +1,33 @@
...
+from ... data_structures import GPLayer
+from ... base_types import AnimationNode, VectorizedSocket
+
+class GPLayerFromFramesNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GPLayerFromFramesNode"
+    bl_label = "GP Layer From Frames"
+    errorHandlingType = "EXCEPTION"
+
+    useFrameList: VectorizedSocket.newProperty()
+    useGPFrameList: VectorizedSocket.newProperty()
```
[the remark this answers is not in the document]
3DSinghVFX: Removed.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,136 @@
...
+        outFrames = [self.getFrame(frames, index) for index in frameIndices]
+        return outFrames, layer.layerName, layer.blendMode, layer.opacity, layer.passIndex
+
+    def execute_FrameNumber(self, layer, frameNumber):
+        frames = layer.frames
+        if len(layer.frames) == 0:
+            return GPFrame(), layer.layerName, layer.blendMode, layer.opacity, layer.passIndex
+
+        frameNumbers = self.getFrameNumbers(frames)
+        index = self.getFrameNumberIndex(frameNumbers, frameNumber)
```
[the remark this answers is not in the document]
3DSinghVFX: Added.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: This function is no longer needed.

animation_nodes/nodes/gpencil/gp_layer_info.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_object_input.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,89 @@
...
+        gpencil.layers.active.frames.update()
+        return object
+
+    def setLayerData(self, gpencil, layer):
+        gpencilLayer = self.getLayer(gpencil, layer)
+        gpencilLayer.clear()
+
+        for frame in layer.frames:
+            gpFrame = gpencilLayer.frames.new(frame.frameNumber, active = True)
+            gpFrame.clear()
```
OmarEmaraDev: Do we need to clear things when they were just created?

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,89 @@
...
+            for stroke in frame.strokes:
+                gpStroke = self.setStrokeProperties(gpFrame.strokes.new(), stroke)
```
OmarEmaraDev: I think you should create the stroke in a separate statement. And don't return anything in the `setStrokeProperties` method.

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
```diff
@@ -0,0 +1,56 @@
...
+        self.newInput("Text", "End Cap Mode", "endCapMode", value = 'ROUND', hide = True)
+        self.newInput("Integer", "Material Index", "materialIndex", value = 0)
+        self.newInput("Text", "Display Mode", "displayMode", value = 'SCREEN', hide = True)
+        self.newOutput("GPStroke", "Stroke", "stroke")
+
+    def execute(self, vertices, strengths, pressures, uvRotations, lineWidth, drawCyclic,
+                startCapMode, endCapMode, materialIndex, displayMode):
+
+        amount = len(vertices)
+        stroke = GPStroke()
```
OmarEmaraDev: Move the everything before this statement and pass them to the constructor.

animation_nodes/nodes/gpencil/gp_layer_info.py
```diff
@@ -0,0 +1,127 @@
...
+    def execute_AllFrames(self, layer):
+        frames = layer.frames
+        return frames, layer.layerName, layer.blendMode, layer.opacity, layer.passIndex
+
+    def getActiveFrame(self, currentFrame, frames):
+        return max((frame for frame in frames if frame.frameNumber <= currentFrame),
+                    key = lambda frame : frame.frameNumber, default = frames[0])
+
+    def getFrameFromNumber(self, frames, frameNumber):
+        filteredFrame = list(filter(lambda f: f.frameNumber == frameNumber, frames))
```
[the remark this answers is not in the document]
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,89 @@
...
+        gpencil.layers.active.frames.update()
+        return object
+
+    def execute_LayerList(self, object, layers):
+        if object is None: None
+        gpencil = self.getObjectData(object)
+
+        if len(layers) == 0: return object
+        for layer in layers:
+            if layer is None: return
```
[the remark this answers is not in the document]
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_object_output.py
3DSinghVFX: Not for frames. It is removed.

animation_nodes/nodes/gpencil/gp_object_output.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
3DSinghVFX: Replaced.

3DSinghVFX:
> Also, can use an object as the property of the Layer and Layer List sockets? See the Spline socket to see how this works.

Omar, I think it is possible. However, the grease pencil objects generally have multiple layers, so the GP Object Input node is enough.

The Evaluate Spline node sample the first spline of the object. If it is necessary to implement _object as the property_ for Layer or Layer List sockets then GP Layer Info should allow selecting (by name) the layer if we input a grease pencil object. I'll try and report back to you.

OmarEmaraDev:
Lets do this later then. I want to merge this first.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if len(layers) == 0: return object
+        for layer in layers:
+            self.setLayerData(gpencil, layer)
+        gpencil.layers.active.frames.update()
+        return object
+
+    def setLayerData(self, gpencil, layer):
+        gpencilLayer = self.getLayer(gpencil, layer)
+        gpencilLayer.clear()
```
OmarEmaraDev: Clear inside the `getLayer` function.

I wounder if we should clear the whole object data. So the layers of the object will be only those we input.

animation_nodes/sockets/object.py
```diff
@@ -59,6 +59,17 @@ def createObject(self):
             data = bpy.data.curves.new("Curve", "CURVE")
             data.dimensions = "3D"
             data.fill_mode = "FULL"
+        if type == "GPENCIL":
+            data = bpy.data.grease_pencils.new("ANGPencil")
+            scene = bpy.context.scene
+            scene.grease_pencil = data
```
OmarEmaraDev: Why do we set annotations to this data block?

animation_nodes/nodes/gpencil/gp_object_output.py
3DSinghVFX: I have tried that, But If we clear the whole data then it is not possible to add manual layers and draw on them. That is the reason I only clear the layers which are added in the AN.

OmarEmaraDev: Why is this the case?

animation_nodes/sockets/object.py
3DSinghVFX: Sorry, we do not need set annotations.

animation_nodes/nodes/gpencil/gp_object_output.py
3DSinghVFX: If we clear the data of the object it clears everything that means all layers of the object.

3DSinghVFX:
> Update the CHANGELOG.

Omar, do these nodes come in _Unreleased > Added_ category?

OmarEmaraDev:
> > Update the CHANGELOG.
> 
> Omar, do these nodes come in _Unreleased > Added_ category?

Yes.

animation_nodes/nodes/gpencil/gp_object_output.py
OmarEmaraDev: Ok, but that's what I am after. I would the expect the object to have only the layers that I gave it. It is like the Spline Object Output node. The exiting splines are removed and only the ones I gave it are there.

3DSinghVFX: This will remove the artist's control over the GP objects from AN. The reason I have reworked the whole GP nodes to add many frames and layers through the AN as well as a manual which will make these nodes more powerful. Here is a simple example, a tree trunk is drawn manually but leaves movement is controlled by AN. 

We can add a button in the GP Output node to clear all the layers of the GP object if someone need to clear them.

OmarEmaraDev: On the other hand, one will not be able to remove existing layers. To preserve existing layers, one might read the layers from the object, append one's layers, then write them.

Maybe we can add an option in the advanced node settings to control this?

3DSinghVFX: Here is my plan:
1. By default, the GP Output node clears all layers.
2. If someone wants to add a custom layer(s) then it will enable the _Append_ option in the advanced node settings. Then, GP Output node does not clear all the layers, only the layers from the AN.

What do you think?

OmarEmaraDev: Looks good. :+1:

3DSinghVFX: Okay, I'll make the changes.

3DSinghVFX: Added the option for appending custom layers.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,97 @@
...
+        gpStroke.line_width = stroke.lineWidth
+        gpStroke.material_index = stroke.materialIndex
+        gpStroke.display_mode = stroke.displayMode
+        gpStroke.draw_cyclic = stroke.drawCyclic
+        gpStroke.start_cap_mode = stroke.startCapMode
+        gpStroke.end_cap_mode = stroke.endCapMode
+
+    def getLayer(self, gpencil, layer):
+        layerName = layer.layerName
+        if layerName in gpencil.layers:
```
OmarEmaraDev: You can check for `appendCustomLayers` here. Since if it is not set, there won't be any layers.

animation_nodes/nodes/gpencil/gp_object_output.py
```diff
@@ -0,0 +1,97 @@
+import bpy
+from bpy.props import *
+from ... base_types import AnimationNode, VectorizedSocket
+
+class GPObjectOutputNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GPObjectOutputNode"
+    bl_label = "GP Object Output"
+    errorHandlingType = "EXCEPTION"
+
+    appendCustomLayers: BoolProperty(name = "Append Custom Layers", default = False,
```
OmarEmaraDev: I would just call this `Append Layers`.

animation_nodes/sockets/object.py
```diff
@@ -59,6 +59,16 @@ def createObject(self):
             data = bpy.data.curves.new("Curve", "CURVE")
             data.dimensions = "3D"
             data.fill_mode = "FULL"
+        if type == "GPENCIL":
+            data = bpy.data.grease_pencils.new("ANGPencil")
+            layer = data.layers.new("AN-Layer", set_active=True)
```
OmarEmaraDev: Spaces around `=`.

3DSinghVFX: Ohh, sorry.

animation_nodes/sockets/object.py
```diff
@@ -59,6 +59,16 @@ def createObject(self):
...
+        if type == "GPENCIL":
+            data = bpy.data.grease_pencils.new("ANGPencil")
+            layer = data.layers.new("AN-Layer", set_active = True)
+            frame = layer.frames.new(0)
+            stroke = frame.strokes.new()
+            stroke.display_mode = "3DSPACE"
+            stroke.material_index = 0
+            material = bpy.data.materials.new('AN-Material')
+            bpy.data.materials.create_gpencil_data(material)
+            data.materials.append(material)
```
OmarEmaraDev: I think all of this should be removed. We should leave everything at default.

3DSinghVFX: I'll make the changes.

3DSinghVFX: It is done.

3DSinghVFX:
Thank you helping me :)

PR #1300, opened 2020-01-30
Added the Change Spline Direction node.

3DSinghVFX:
This branch has **Change Spline Direction** node that can change the spline(s) direction,
Screenshot from 2020-01-31 01-57-55


OmarEmaraDev:
What are the possible applications for this?

You can use a vectorization code effect here. See any node that have this line `codeEffects = [VectorizedSocket.CodeEffect]`

3DSinghVFX:
> What are the possible applications for this?

To reverse the direction of the spline, we need to reverse the Point List (and Radii, Tilt) with Loop Node, e.g., https://github.com/3DSinghVFX/animation_nodes/issues/84

> You can use a vectorization code effect here. See any node that have this line `codeEffects = [VectorizedSocket.CodeEffect]`

Okay.

Should I add a boolean input for this node as we have in the Cyclic Spline Node?

OmarEmaraDev:
I am just concerned that users will prefer to use this node instead of inverting the evaluation parameter, which is much, much more efficient. But sure, it is not really an issue.

No. The Cyclic Spline node sets an attribute in the splines. This node operates on it.

3DSinghVFX:
> I am just concerned that users will prefer to use this node instead of inverting the evaluation parameter, which is much, much more efficient.

You mean something like this,
Screenshot from 2020-02-02 12-07-19

Should I change the code of the node?

OmarEmaraDev:
No. I mean that it is an XY problem. You think you want to reverse the spline to get the desired effect. But what you actually want to do is reverse the parameters of the trimming.

For instance, those two networks trim the spline as time pass, but one is the reverse of the other. 



Makes sense?

3DSinghVFX:
Yes.

3DSinghVFX:
Updated the Change Spline Direction node's code, now it uses `codeEffects = [VectorizedSocket.CodeEffect]`.

animation_nodes/nodes/spline/change_spline_direction.py
```diff
@@ -0,0 +1,37 @@
+import bpy
+from ... data_structures import PolySpline, BezierSpline
+from ... base_types import AnimationNode, VectorizedSocket
+
+class ChangeSplineDirectionNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ChangeSplineDirectionNode"
+    bl_label = "Change Spline Direction"
+
+    codeEffects = [VectorizedSocket.CodeEffect]
```
OmarEmaraDev: Move this one line upward.

3DSinghVFX: It is Moved.

PR #1307, opened 2020-02-04
Added the grease pencil nodes to change layer and stroke properties.

3DSinghVFX:
I have added the Grease Pencil nodes to change layer and stroke properties. These nodes are work similar to nodes for spline, e.g., to make cyclic spline, reset the spline radius, replicate spline, etc.

- A Layer has Blend Mode, Opacity and Pass Index properties which can be changed with the following nodes,
Screenshot from 2020-02-04 18-03-39
These nodes can take a Layer (or Layer list), Opacity (or Opacities), and Pass Index (or Pass Indices).
 
- A Stroke has Cyclic, Line Width, Start Cap Mode, Start Cap Mode, Material Index, Display Mode properties which can control with the following nodes,
Screenshot from 2020-02-04 18-28-27
These nodes are can take a Stroke (Stroke List), Cyclic-Boolean (Cyclic List), Line Width (Line Width List), Start Cap Mode-Text (Start Cap Mode List), End Cap Mode-Text (End Cap Mode List), Material Index (Material Indices).
- A Stroke or Strokes can be transformed from one place to another in 3D space or can be replicated like a spline.  Following nodes can transform and replicate a Stroke or Strokes,
Screenshot from 2020-02-04 18-29-00
_Replicate Stroke node's code is adopted from Replicate Spline node_.


OmarEmaraDev:
Sorry for the delay @3DSinghVFX. I am still trying to figure out what is going on in #1308.

3DSinghVFX:
> Sorry for the delay @3DSinghVFX. I am still trying to figure out what is going on in #3DSinghVFX 

It is fine. That issue was more important to be resolved.


> The Blend and Display Mode nodes should have a string input instead.

Okay, I'll make the changes.

OmarEmaraDev:
Note that you can create a virtual list from a python object using the `VirtualPyList` structure. This could be used with strings, for instance.

3DSinghVFX:
Ohh, that is nice. I'll make the changes.

3DSinghVFX:
>     1. The Replicate node and the Transform node...
Omar, I think it is more like the Splines because we are at the Stroke level. Like in splines, we have a separate node to transformation and replication of splines. At the Layer level, we can _Transform Layer_ node as we have for Mesh.

>     2. All nodes should accept two virtual vectorized...
I agree with you. I think we do this only to Set Stroke Width and Set Stroke Material Index nodes.

>     3. The nodes should follow the naming convention _Set {Stroke, Frame} Property_...
Should I keep the _GP_ as a prefix for the name of these nodes?

OmarEmaraDev:
1. I see. Looks like you are right. However, the transform node should follow my second point as well.
2. Why only those nodes? I think we should do this to all nodes if we possible, which seem to be the case here. Use VirtualPyList when necessary.
3. Yes, keep the GP prefix.

3DSinghVFX:
Okay, I'll make the changes.

3DSinghVFX:
Omar, I have renamed the nodes and made the changes according to your _second point (2. All nodes should accept two virtual vectorized lists...)_ for all the _Set nodes_. I have added the new _GP Transform Layer_ node.

- Should we also add an option to offset frame-number and custom layer-name in the _GP Transform Layer_ node?
- Should we also add a node to offset the frame-number at the layer level?

OmarEmaraDev:
Don't add more nodes in this pull request. We can discuss those separately.

OmarEmaraDev:
This is not exactly what I had in mind.

We should have two functions only, one for the single case and one for the case where any of the inputs is a list. In the latter function, we create two virtual lists, use the `getMaxRealLength` class method to get the length of the output list, and finally loop based on this length. Look for all calls to the `getMaxRealLength` function and you will get what I mean.

3DSinghVFX:
> Don't add more nodes in this pull request. We can discuss those separately.

Should I remove the _GP Transform Layer_ node for now?

> We should have two functions...

Do you mean something like this,
https://github.com/JacquesLucke/animation_nodes/blob/90116f9e1fcd05ed32b17af8ccd85e4b7c7fc4e5/animation_nodes/nodes/mesh/transform_mesh.py#L118-L129

- I think we should rename the _GP Set Stroke Cyclic_ node as _GP Make Stroke Cyclic_ node

- What do you think about this pull request https://github.com/3DSinghVFX/animation_nodes/pull/105
Can we merge this pull request or the next time?

OmarEmaraDev:
You shouldn't have added more to this pull request, but no need to remove anything.

> I think we should rename the GP Set Stroke Cyclic node as GP Make Stroke Cyclic node.

Yes, `Make Stroke Cyclic` seems fine. However, I don't want GP as a prefix of the node name, but rather, a prefix of the type. It should read naturally. Like this:

- GP Make Stroke Cyclic -> Make GP Stroke Cyclic
- GP Replicate Stroke -> Replicate GP Stroke
- GP Set Layer Blend Mode -> Set GP Layer Blend Mode
- ...

Also, `GP Stroke Transform` -> `Transform GP Stroke`. And so on.

> What do you think about this pull request 3DSinghVFX#105

If you mean merging upstream, then the pull request should be submitted upstream.

3DSinghVFX:
> If you mean merging upstream, then the pull request should be submitted upstream.

Omar, I did not understand. Should I tell Sedrew (https://github.com/3DSinghVFX/animation_nodes/pull/105) to submit the pull request to the Animation Nodes?

OmarEmaraDev:
Yes, Sedrew should submit the pull request here. And you can review it here.

3DSinghVFX:
Okay.

3DSinghVFX:
I have changed the names.

Omar, should we add frame-number offset (per layer) and custom layer-name option in _Transfom GP Layer_ node?

OmarEmaraDev:
Can you elaborate? What should those options do?

OmarEmaraDev:
If it is not simple and you want to do it. Feel free to move the *Transform Layer* node into a separate pull request.

3DSinghVFX:
The _Transfom GP Layer_ node allows replicating of a layer basis on the input vectors/matrices. So, during the replication, we can offset the frame-number with _Frame-Number Offset_ option i.e., we can offset the timing of the animation of character (a plant) for a crowd (forest). The second option allows naming the of the instanced layers for the better organization of the layers.

OmarEmaraDev:
I don't think we should add this in this pull request, we will discuss this later.

3DSinghVFX:
Okay, that is fine.

animation_nodes/nodes/gpencil/transform_gp_stroke.py
```diff
@@ -0,0 +1,45 @@
...
+        amount = VirtualPyList.getMaxRealLength(_strokes, _matrices)
+
+        outStrokes = []
+        for i in range(amount):
+            strokeNew = _strokes[i].copy()
+            self.strokeTransfom(strokeNew, _matrices[i])
+            outStrokes.append(strokeNew)
+        return outStrokes
+
+    def strokeTransfom(self, stroke, matrix):
```
OmarEmaraDev: Rename this to `transformStroke`.

3DSinghVFX: Okay.

3DSinghVFX:
> Update the CHANGELOG.

Should I also mention about the fixes of the _GP Object Input_ and _GP Object Output_ nodes or not required?

OmarEmaraDev:
Don't mention them. The CHANGELOG mention changes since the last subversion. Those bugs were introduced in the same subversion.

3DSinghVFX:
Okay.

animation_nodes/nodes/gpencil/transform_gp_stroke.py
3DSinghVFX: Changed.

3DSinghVFX:
Omar, There is some issue with the _Set_ nodes when they are in the _Loop_ node,
Screenshot from 2020-02-09 20-14-50

3DSinghVFX:
Omar, `amount` is zero in the above example, I do not why?
https://github.com/JacquesLucke/animation_nodes/blob/fe280b91ab9857ab7826065269dfc69cc7b888b4/animation_nodes/nodes/gpencil/set_gp_stroke_line_width.py#L33

OmarEmaraDev:
Can you print `strokes` and see if it is actually zero?

3DSinghVFX:
_strokes_ in the node code is not zero.

OmarEmaraDev:
I am not sure. I will look into it.

sedrew:
> The Blend and Display Mode nodes should have a string input instead.
Can I object? Of course, string adds functionality to the node. But the drop-down list was more convenient, the user is more used to choosing from the list. But this is my personal opinion.

3DSinghVFX:
The `_strokes` and `_lineWidth` has non-zero length but `amount` is zero. However, it is working fine in the case of _Set GP Layer_ nodes.

OmarEmaraDev:
@sedrew Say you want to change the Display Type based on some condition, what do you do this case? Convenience comes at cost of flexibility and efficiency.

3DSinghVFX:
> > The Blend and Display Mode nodes should have a string input instead.
> > Can I object? Of course, string adds functionality to the node. But the drop-down list was more convenient, the user is more used to choosing from the list. But this is my personal opinion.

If those nodes are in the Loop Node or in the Group Node then it is difficult to change them with the _drop-down_.

sedrew:
I understood. Thanks for the clarification.

OmarEmaraDev:
@3DSinghVFX I can't reproduce the issue. Can you reconfirm with a clean checkout? Can you provide a blend file?

3DSinghVFX:
Let me check again with the fresh build.

3DSinghVFX:
@OmarEmaraDev Issue is still there, I have tested with Blender 2.82 and 2.83,
Screenshot from 2020-02-09 23-19-00

OmarEmaraDev:
Can you attach the blend file?

3DSinghVFX:
> Can you attach the blend file?

Here is the file,
[GPencil1.zip](https://github.com/JacquesLucke/animation_nodes/files/4177296/GPencil1.zip)

3DSinghVFX:
Omar, I have noticed that _Set GP Stroke Line Width_ node works when there are two or more strokes as input.

3DSinghVFX:
In case of stroke list (one or more strokes) if I input a _line width list_ then it is also working.

OmarEmaraDev:
Ok, it was a bug in VirtualPyList. Fixed in 6d953a6. Does it work now?

3DSinghVFX:
> Ok, it was a bug in VirtualPyList. Fixed in [6d953a6](https://github.com/JacquesLucke/animation_nodes/commit/6d953a6fe138de21575c7a72fb3f68202930fdd6). Does it work now?

Thanks, now it is working fine.

OmarEmaraDev:
Any reason why the Transform Layer node is implemented differently? Why can't we use the same approach we used in the Transform Stroke node?

3DSinghVFX:
> Any reason why the Transform Layer node is implemented differently? Why can't we use the same approach we used in the Transform Stroke node?

I followed the _Transform Mesh_ node. I think we can use the same approach.

OmarEmaraDev:
I think the same approach should be taken. I will probably merge after you update that. So if you want to do some final testing, do that.

3DSinghVFX:
Okay, I'll update the _Transfrom GP Layer_ node.

3DSinghVFX:
Omar, I have done several tests, so far all the gp nodes are working fine.

JacquesLucke:
Sorry for the late reply. Does it make sense to have a `Set GP Layer Attribute` and `Set GP Stroke Attribute` instead? It could have inputs similar to the `Mesh Object Output` node (so they can be enabled and disabled).

OmarEmaraDev:
Seems like a good idea. @3DSinghVFX It would be good to handle this before your materials patch.

3DSinghVFX:
Okay. I'll create the pull request soon.

3DSinghVFX:
I have a small query, Are these nodes going to operate on the grease pencil object or layer (stroke) object?
Here is the layout,
Screenshot from 2020-02-10 22-24-53

JacquesLucke:
That is not what I meant. The node should have multiple inputs for the different attributes.

OmarEmaraDev:
I don't think this is what Jacques meant. They will be similar to the *Mesh Object Output* node, but they will only return the strokes/layers, they won't delegate the changes to Blender.

3DSinghVFX:
Sorry, I did not understand completely. Do you mean the _Set GP Layer Attribute_ node to set _stroke(s), frame (or frame-number), layer-name, etc_ together and return a Layer?

OmarEmaraDev:
Yes. Or possibly output multiple layers.

3DSinghVFX:
Okay. But multiple output layers on what basis, I mean basis on the layer-name list or sth else?

What about the _Set GP Stroke Attribute_ node, to set _points, strengths (all stroke attributes) + frame-number, layer-name_ and return a Layer?

OmarEmaraDev:
Now, we have:
- Added Set GP Layer Blend Mode node.
- Added Set GP Layer Opacity node.
- Added Set GP Layer Pass Index node.

We want to combine all of those into a single *Set GP Layer Attribute* node. The node will have inputs for *Blend Mode*, *Opacity*, and *Pass Index*. Each input will utilize the *Is Used* property much like the output nodes we have. The output will be the input layer(s) after setting the appropriate attribute. The same goes for Strokes. The inputs can be virtual like we did before.

3DSinghVFX:
Thanks for the clarification. Here is the layout, I make the inputs _Is Used_ property,
Screenshot from 2020-02-10 23-25-13

3DSinghVFX:
With the _Is Used_ property for inputs,
Screenshot from 2020-02-10 23-35-12

OmarEmaraDev:
Yes, looks alright. Attributes should be vectorized as well though.

3DSinghVFX:
Okay.

PR #1310, opened 2020-02-08
Fixed the Distribute Matrices node for the Spiral option.

3DSinghVFX:
The _Distribute Matrices_ node with the _Spiral_ break the node-tree when _Amount = 1_ and gives this error,
Screenshot from 2020-02-09 02-48-57
Now, it is fixed in this branch.


3DSinghVFX:
I have updated the `CHANGELOG.md` file.

PR #1314, opened 2020-02-11
Added the GP Material Output and GP Object Material Output nodes.

3DSinghVFX:
This branch has _GP Material Output_ and _GP Object Material Output_ nodes.
- GP Material Output: This node allows set/change the properties of the grease pencil material.
- GP Object Material Output: This node allows to set material to the grease pencil object.

Screenshot from 2020-02-12 00-28-51


animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,51 @@
...
+    ("DOTS", "Dots", "", "NONE", 1),
+    ("BOX", "Boxes", "", "NONE", 2)
+]
+
+class GPMaterialOutputNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GPMaterialOutputNode"
+    bl_label = "GP Material Output"
+
+    modeType: EnumProperty(name = "Mode Type", default = "LINE",
+        items = modeTypeItems, update = AnimationNode.refresh)
```
OmarEmaraDev: Is a refresh really needed here?

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,51 @@
...
+
+    strokeBool: BoolProperty(name="Stroke",
+                        description = "Stroke Mode for material",
+                        default=True, update=propertyChanged)
+    fillBool: BoolProperty(name="Fill",
+                        description = "Fill Mode for material",
+                        default=False, update=propertyChanged)
```
OmarEmaraDev: This should be inputs, not properties. Also, note the spaces around the equal sign.

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def draw(self, layout):
+        layout.prop(self, "strokeBool")
+        layout.prop(self, "modeType", text = "")
+        layout.row().separator()
+        layout.prop(self, "fillBool")
+
+    def execute(self, material, color, fillColor, passIndex):
+        if material is None: return None
+        bpy.data.materials.create_gpencil_data(material)
```
OmarEmaraDev: Shouldn't we check if the data already exist first? Using the `is_grease_pencil` method maybe.

animation_nodes/ui/node_menu.py
```diff
@@ -517,6 +517,9 @@ def draw(self, context):
         insertNode(layout, "an_GPFrameFromStrokesNode", "Frame From Strokes")
         insertNode(layout, "an_GPLayerFromFramesNode", "Layer From Frames")
         insertNode(layout, "an_GPObjectOutputNode", "Object Output")
+        layout.separator()
+        insertNode(layout, "an_GPMaterialOutputNode", "Material Output")
```
OmarEmaraDev: Also add this to the Materials panel as well.

animation_nodes/nodes/gpencil/gp_object_material_output.py
```diff
@@ -0,0 +1,47 @@
+import bpy
+from bpy.props import *
+from ... events import propertyChanged
+from ... base_types import AnimationNode, VectorizedSocket
+
+class GPObjectMaterialOutputNode(bpy.types.Node, AnimationNode):
```
OmarEmaraDev: I think this node is not needed. One can use the *Object Material Output* node instead.

3DSinghVFX: Okay. Omar, as we have added the option for _Custom Layers_ in _GP Output Object_ node, should we add an option for _Custom Material_ in _Object Material Output_ node? If someone is painting a custom layer, it will also add their custom materials as well.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: I replaced it with _propertyChanged_.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: Changed.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: Done.

animation_nodes/ui/node_menu.py
3DSinghVFX: Added.

animation_nodes/nodes/gpencil/gp_object_material_output.py
OmarEmaraDev: Ok, you may add an Append Material option in the advanced node settings.

3DSinghVFX: Okay. I'll make the changes.

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,48 @@
...
+    ("LINE", "Line", "Draw for strokes using a continuous line", "NONE", 0),
+    ("DOTS", "Dots", "Draw for strokes using separated dots", "NONE", 1),
+    ("BOX", "Boxes", "Draw for strokes using separated rectangle boxes", "NONE", 2)
+]
+
+class GPMaterialOutputNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GPMaterialOutputNode"
+    bl_label = "GP Material Output"
+
+    modeType: EnumProperty(name = "Mode Type", default = "LINE",
```
OmarEmaraDev: This should be an input as well. Same discussion we had before.

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,48 @@
...
+        if material is None: return None
+        if not material.is_grease_pencil:
+            bpy.data.materials.create_gpencil_data(material)
+
+        gpMaterial = material.grease_pencil
+        gpMaterial.show_stroke = strokeBool
+        gpMaterial.mode = self.modeType
+        gpMaterial.show_fill = fillBool
+
+        if strokeBool and  gpMaterial.stroke_style == "SOLID":
```
OmarEmaraDev: I don't see why we need those if statements. We should set the colors regardless.

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,48 @@
...
+        self.newInput("Boolean", "Fill", "fillBool", value = False)
+        self.newInput("Color", "Fill Color", "fillColor")
+        self.newInput("Integer", "Pass Index", "passIndex", value = 0, hide = True)
+
+        self.newOutput("Material", "Material", "material")
+
+    def draw(self, layout):
+        layout.prop(self, "modeType", text = "")
+
+    def execute(self, material, strokeBool, strokeColor, fillBool, fillColor, passIndex):
```
OmarEmaraDev: Booleans should be called `isSomething`, `useSomething`, and so on. Not `somethingBool`. In this case, we can just use `showFill` and `showStroke`.

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,48 @@
+import bpy
```
OmarEmaraDev: I think this node should use the `isUsed` socket mechanism.

animation_nodes/nodes/material/object_material_output.py
```diff
@@ -1,10 +1,15 @@
 import bpy
+from bpy.props import *
 from ... base_types import AnimationNode, VectorizedSocket
 
 class ObjectMaterialOutputNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ObjectMaterialOutputNode"
     bl_label = "Object Material Output"
 
+    appendMaterials: BoolProperty(name = "Append Materials", default = False,
+        description = "This option allow to add custom materials",
```
OmarEmaraDev: `Append input material(s) to the object's materials instead of overwriting them`

animation_nodes/nodes/material/object_material_output.py
```diff
@@ -23,13 +32,40 @@ def getExecutionFunctionName(self):
...
+
+        objectMaterials = object.data.materials
+        if not self.appendMaterials: objectMaterials.clear()
+
+        if object.type == "GPENCIL":
+            if material is not None:
+                if not material.is_grease_pencil:
+                    bpy.data.materials.create_gpencil_data(material)
+                self.setObjectMaterial(objectMaterials, material)
+            return object
```
OmarEmaraDev: Don't make a distinction here. `create_gpencil_data` shouldn't be called, it is only supposed to set materials, not edit them.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: Done.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: Omar, It is needed, when `stroke_style = 'TEXTURE'` then there no option for `color`.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: Done.

animation_nodes/nodes/material/object_material_output.py
3DSinghVFX: Omar, there is a difference between the _material_ for the _Mesh or Curve_ type object and material for the _GPencil_ object. We can't assign normal material (means material of the _Mesh_ or _Curve_) to GPencil object, we have to first convert that material into the gpencil-material.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: Okay. I'll make the changes.

animation_nodes/nodes/material/object_material_output.py
3DSinghVFX: Sorry, I didn't understand this point?

OmarEmaraDev: I mean change the description to this.

3DSinghVFX: Okay.

3DSinghVFX:
Omar, Can we hide the _Stroke Color_ socket with the _Stroke (strokeBool)_ input?

https://github.com/JacquesLucke/animation_nodes/blob/701173a81913f6138bde35739003007451090848/animation_nodes/nodes/gpencil/gp_material_output.py#L21-L22

animation_nodes/nodes/material/object_material_output.py
OmarEmaraDev: We can assign normal materials to GP objects. Yes, it won't be very useful, but that's what the user asked for. Ensuring that the material has GP data is the user's responsibility.

animation_nodes/nodes/gpencil/gp_material_output.py
OmarEmaraDev: The color data is still there, and it should be set regardless. In fact, when the *Use As Stencil Mask* or the *Mix With Color* options are enabled in the `TEXTURE` style, the color will be utilized.

OmarEmaraDev:
> Omar, Can we hide the Stroke Color socket with the Stroke (strokeBool) input?

 I don't think we should do that.

3DSinghVFX:
Okay.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: You're right. Thanks.

animation_nodes/nodes/material/object_material_output.py
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/gp_material_output.py
```diff
@@ -0,0 +1,53 @@
...
+        if any([isShowStroke, isStrokeDrawMode, isStrokeColor, isShowFill, isFillColor, isPassIndex]):
+            yield                      "    gpMaterial = material.grease_pencil"
+            if isShowStroke:     yield "    gpMaterial.show_stroke = showStroke"
+            if isStrokeDrawMode: yield "    self.setStrokeDrawMode(gpMaterial, strokeDrawMode)"
+            if isStrokeColor:    yield "    gpMaterial.color = strokeColor"
+            if isShowFill:       yield "    gpMaterial.show_fill = showFill"
+            if isFillColor:      yield "    gpMaterial.fill_color = fillColor"
+            if isPassIndex:      yield "    gpMaterial.pass_index = passIndex"
+        else:
+            yield "material"
```
OmarEmaraDev: Not sure what you are trying to do here, but this line currently does nothing.

animation_nodes/nodes/material/object_material_output.py
```diff
@@ -1,10 +1,15 @@
...
+from bpy.props import *
 from ... base_types import AnimationNode, VectorizedSocket
 
 class ObjectMaterialOutputNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_ObjectMaterialOutputNode"
     bl_label = "Object Material Output"
 
+    appendMaterials: BoolProperty(name = "Append Materials", default = False,
+        description = "Append input material(s) to the object's materials instead of overwriting them",
+        update = AnimationNode.refresh)
```
OmarEmaraDev: Refresh is not needed here.

animation_nodes/nodes/material/object_material_output.py
```diff
@@ -23,13 +32,25 @@ def getExecutionFunctionName(self):
...
+
+        objectMaterials = object.data.materials
+        if not self.appendMaterials: objectMaterials.clear()
+
         for material in materials:
-            object.data.materials.append(material)
+            self.setObjectMaterial(objectMaterials, material)
         return object
+
+    def setObjectMaterial(self, objectMaterials, material):
```
OmarEmaraDev: Rename this to `appendMaterial`.

animation_nodes/nodes/gpencil/gp_material_output.py
3DSinghVFX: I'm trying to return material when any input-sockets are not used. Removed.

animation_nodes/nodes/material/object_material_output.py
3DSinghVFX: Done.

animation_nodes/nodes/material/object_material_output.py
3DSinghVFX: Okay. Removed.

animation_nodes/nodes/gpencil/gp_material_output.py
OmarEmaraDev: Ah, I see. This is not a function, the code is emitted directly to the execution code of the execution unit. In this case, we **assign** to the outputs, not return them. In this case, however, we want to return the input material, the output has the same name as the input, so we don't have to do anything.

3DSinghVFX: Okay, thanks.

PR #1315, opened 2020-02-12
Added GP Mask Layer

sedrew:
image


OmarEmaraDev:
Thanks for the patch @sedrew! It would be nice if you can provide a minimal example of use so that we can understand how this works better.

@3DSinghVFX Will you review this?

sedrew:
This allows to enable or disable the layer mask. Takes a boolean value or boolean list. Present in other GP nodes where there is an input or output socket a layer. Hidden by default.

ezgif com-gif-)
image

3DSinghVFX:
> I think this should be called `isMaskLayer`. Also, please update the CHANGELOG file.

Omar, in _Layer_ panel of the blender, it is called as _Mask Layer_.

OmarEmaraDev:
Initially, I thought *Mask Layer* was a reference to another Layer that masks this layer, not that this layer is a mask layer. That's why I want to call it `isMaskLayer`.

3DSinghVFX:
Okay.

Here is the explanation from the Blender Doc: _When the mask is enabled, all the layers below the current one function as a single mask. This mask hides any pixel in the current layer outside the mask created by the underlying layers._

OmarEmaraDev:
I see. In this case, I think the name is fine.

animation_nodes/nodes/gpencil/set_gp_layer_attributes.py
```diff
@@ -40,19 +43,22 @@ def getExecutionCode(self, required):
         isBlendMode = s[2].isUsed
         isOpacity = s[3].isUsed
         isPassIndex = s[4].isUsed
-        if any([self.useLayerList, self.useNameList, self.useBlendModeList, self.useOpacityList, self.usePassIndexList]):
-            if any([isName, isBlendMode, isOpacity, isPassIndex]):
+        isMaskLayer = s[5].isUsed
+        if any([self.useLayerList, self.useNameList, self.useBlendModeList, self.useOpacityList, self.usePassIndexList, self.useMaskLayerList]):
```
OmarEmaraDev: Split this into two lines.

sedrew:
I made a change to CHANGELOG.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 
 - Optimized the *Object Instancer* node.
 - Make all numeric types comparable.
+- Added Mask Layer to the Set GP Layer Attributes node, GP Object Input node, GP Layer Info node, GP Layer From Frames node, GP Object Output node
```
OmarEmaraDev: Actually, we are in the same subversion, so we shouldn't update the CHNAGELOG. My apologies, this should be removed.

3DSinghVFX:
Thanks @sedrew :)

PR #1316, opened 2020-02-12
Added the Replicate GP Layer and Offset GP Layer Frame Number nodes.

3DSinghVFX:
This branch has _Replicate GP Layer_ and _Offset GP Layer Frame Number_ nodes.
- **Replicate GP Layer** node: Allow to replicate the Layer(s) basis on the input matrices(vectors). This node also all to offset the frame-number per Layer(s) during the replication.

- **Offset GP Layer Frame Number** node: Allow to offset the frame-number per Layer. This is very handy to make multiple characters/plants/graphics with offset animation.

Screenshot from 2020-02-13 02-36-00


animation_nodes/nodes/gpencil/offset_gp_layer_frame_number.py
```diff
@@ -0,0 +1,43 @@
...
+    bl_label = "Offset GP Layer Frame Number"
+
+    useLayerList: VectorizedSocket.newProperty()
+    useFrameNumberList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput(VectorizedSocket("GPLayer", "useLayerList",
+            ("Layer", "layer"), ("Layers", "layers")), dataIsModified = True)
+        self.newInput(VectorizedSocket("Float", "useFrameNumberList",
+            ("Offset Frame Number", "offset"), (" Offset Frame Numbers", "offsets")), value = 0)
```
OmarEmaraDev: Just call this *Offset*.

animation_nodes/nodes/gpencil/offset_gp_layer_frame_number.py
```diff
@@ -0,0 +1,43 @@
...
+            ("Layer", "outLayer"), ("Layers", "outLayers")))
+
+    def getExecutionFunctionName(self):
+        if self.useLayerList or self.useFrameNumberList:
+            return "execute_LayerList_FrameNumberList"
+        else:
+            return "execute_Layer_FrameNumber"
+
+    def execute_Layer_FrameNumber(self, layer, offset):
+        layerNew = layer.copy()
```
OmarEmaraDev: No need to copy here.

animation_nodes/nodes/gpencil/offset_gp_layer_frame_number.py
3DSinghVFX: Okay. I have a concern with the name of the node. Do you have any idea?

animation_nodes/nodes/gpencil/offset_gp_layer_frame_number.py
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/offset_gp_layer_frame_number.py
OmarEmaraDev: I am not sure. But I think *Offset GP Layer Frames* is ok.

3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/replicate_gp_layer.py
```diff
@@ -0,0 +1,77 @@
...
+        if isinstance(layers, GPLayer):
+            layers = [layers]
+
+        _offsets = VirtualDoubleList.create(offsets, 0)
+
+        outLayers = []
+        for i, vector in enumerate(vectors):
+            for layer in layers:
+                newLayer = layer.copy()
+                self.transformLayerByMatrix(newLayer, _offsets[i], vector)
```
JacquesLucke: Should this be `transformLayerByVector`? Or just pass `Matrix.Translation(vector)` from here.

3DSinghVFX: Yes, I'll change it to `transformLayerByVector`.

OmarEmaraDev: Or use `move` instead of `transform`.

3DSinghVFX: I have added the `move` for vectors.

OmarEmaraDev:
I wounder if we should "increment" the name of the layers after replication, so that no name collision happens. What do you think?

3DSinghVFX:
Yes, I implemented first then I removed it. 

Should we also add an option for a suffix for the duplicated layers?

OmarEmaraDev:
Ok. Do that.

3DSinghVFX:
Omar is this right order for new layer-name `newLayer.layerName = layer.layerName + "-" + nameSuffix + "-" + str(i)`?

OmarEmaraDev:
Oh, I thought you meant adding an option to disable or enable the suffix number. I don't think we should add a suffix, just the number. I will have to go do something, so I will get back to you in couple of hours.

3DSinghVFX:
Okay. I'll make the changes.
No problem.

animation_nodes/nodes/gpencil/replicate_gp_layer.py
```diff
@@ -0,0 +1,80 @@
...
+
+    transformationType: EnumProperty(name = "Transformation Type", default = "Matrix List",
+        items = transformationTypeItems, update = AnimationNode.refresh)
+
+    def create(self):
+        self.newInput(VectorizedSocket("GPLayer", "useLayerList",
+            ("Layer", "layer", dict(defaultDrawType = "PROPERTY_ONLY")),
+            ("Layers", "layers")))
+        self.newInput(self.transformationType, "Transformations", "transformations")
+        self.newInput("Float List", "Offset Frame Numbers", "offsets", value = 0, hide = True)
```
OmarEmaraDev: I don't think we should have offsets here. Users can utilize the Offset node for that.

animation_nodes/nodes/gpencil/replicate_gp_layer.py
```diff
@@ -0,0 +1,80 @@
...
+        self.newInput("Boolean", "Name Suffix", "nameSuffix", value = False, hide = True)
```
OmarEmaraDev: I think we should force the suffix. So no need for this input.

3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/replicate_gp_layer.py
3DSinghVFX: It is easy to offset _frame-number_ with the Offset node when we are replicating a single layer. However, if we are replicating two or more layers then it is difficult to offset _frame-number_ with Offset node. Here is the comparison,
**Offset with the Replicate Layer**,
Screenshot from 2020-02-14 01-16-41

**Offset with the Offset node**,
Screenshot from 2020-02-14 01-16-19

Full scene,
Screenshot from 2020-02-14 01-17-07

OmarEmaraDev: I see. Ok. Rename the input to just *Offset*. Should it be a vectorizaed virtual list?

3DSinghVFX: Okay. I think yes.

3DSinghVFX: Omar, should we name it as _Offset Frame(s)_?

OmarEmaraDev: Ok.

PR #1319, opened 2020-02-15
Added the Texture Input node.

3DSinghVFX:
This branch has the **Texture Input** node which evaluates the texture data (color(s)) basis on the input vectors,
Screenshot from 2020-02-15 18-23-28

Allows to set vertex colors basis on the texture,
Screenshot from 2020-02-15 18-22-18

Allows to set vertex weight basis on the texture,
Screenshot from 2020-02-15 18-22-00


OmarEmaraDev:
Can you provide some numbers for the performance of this node?

I am guessing the Cython code provide no advantages over python. So I would test a pure python approach as well.

3DSinghVFX:
Omar, with Cython function it is `~2.67` times faster,
**With Cython code:**
Screenshot from 2020-02-15 20-46-21

**With Python code:**
Screenshot from 2020-02-15 20-45-45

OmarEmaraDev:
Ok, good. I will consult Jacques about this and get back to you.

OmarEmaraDev:
Unfortunately, we think that this feature shouldn't be supported in Animation Nodes until Blender provides a more efficient method of texture evaluation, or more specifically for images, a more efficient access to image pixel data.

So lets get back to this once an advancement happen in this front. Maybe we will be the ones we make that happen. :wink:

3DSinghVFX:
😊😀 Okay. I'm going to close this for now.

PR #1320, opened 2020-02-15
Gpencil fix bug replicate layer/stroke

sedrew:
Is this a bug? @3DSinghVFX

image

fix2

fix


sedrew:
Name input is not displayed stroke and layer

3DSinghVFX:
@sedrew  Thanks :)

PR #1323, opened 2020-02-16
Added Set Polygon Material Index node

3DSinghVFX:
I have added the **Set Polygon Material Index** node which allows to set material index per polygon,
Screenshot from 2020-02-16 17-47-56

Example,
Screenshot from 2020-02-16 17-53-41

### DIFF
--- a/animation_nodes/nodes/mesh/set_polygon_material_index.py
+++ b/animation_nodes/nodes/mesh/set_polygon_material_index.py
@@ -1,140 +1,41 @@
 import bpy
-from bpy.props import *
-from ... events import propertyChanged
 from ... data_structures import LongList, VirtualLongList
 from ... base_types import AnimationNode, VectorizedSocket
-
-modeItems = [
-    ("ALL", "All", "Set material index of every polygon", "NONE", 0),
-    ("INDEX", "Index", "Set material index of a specific polygon", "NONE", 1)
-]
 
 class SetPolygonMaterialIndexNode(bpy.types.Node, AnimationNode):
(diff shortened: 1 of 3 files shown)


OmarEmaraDev:
Is there a reason why you aren't using `foreach_set` here? I think this node should be similar to the other data set nodes, no need for all of those extra options.

3DSinghVFX:
Omar, I'm using the `foreach_set` function in this node.

You mean only the _ALL_ option always? This node is similar to _Set Vertex Weight_ node.

OmarEmaraDev:
Yes. *Set Vertex Weight* is a special case because its data is stored per element, not contiguously. So the node should be similar to the other set nodes.

3DSinghVFX:
Okay, I'll make the changes.

3DSinghVFX:
I have made the changes.

animation_nodes/nodes/mesh/set_polygon_material_index.py
```diff
@@ -0,0 +1,40 @@
...
+        if object is None: return None
+
+        if object.type != "MESH":
+            self.raiseErrorMessage("Object is not a mesh object.")
+
+        if object.mode != "OBJECT":
+            self.raiseErrorMessage("Object is not in object mode.")
+
+        if self.useMaterialIndexList:
+            if min(materialIndices) < 0:
```
OmarEmaraDev: You can use the `containsValueLowerThan(0)` method here.

3DSinghVFX: Done.

PR #1325, opened 2020-02-17
Decompose Text node.

OmarEmaraDev:
# Description

The *Decompose Text* node provides a new workflow to work with texts in Animation Nodes.

Motion graphics typically requires texts to move and deform in a per character/word basis. Previously, to achieve this, one would create the text, invoke the *Separate Text* operator, use the *Initial Transforms ID Key* as a base transformations, and finally offset those transforms as needed. This workflow is not ideal as it requires a lot of manual intervention. Moreover, it is not really procedural and does not fit the general workflow of Animation Nodes.

The *Decompose Text* node provides information to the user so as to allow a completely procedural workflow. In particular, the node decomposes the input text into characters or words and provides the transformations necessary to position those characters or words in space. The node also allows the user to set a multitude of options including character spacing, word spacing, line spacing, text alignment, text sizing, and more.

A base node tree for this workflow would be as follows:

Base Node Tree

An *Object Instancer* node to create the text objects, a *Text Object Output* node to set the text attributes, and an *Object Matrix Output* node to position the texts in space. Each of the aforementioned nodes takes their inputs from the *Decompose Text* node and possibly from its inputs.

The transforms can be easily manipulated using standard Animation Nodes workflows, such as using falloffs:

Falloff Example

The node itself allows manipulations that were even impossible to create using the older workflow. And it provides flexibility that Blender texts doesn't even allow. For instance, per-character sizing based on a drifting sine wave:

Sizing Example

This is Animation Nodes, so the possibilities are endless here.

# Implementation

The implementation is a bit tricky because information about fonts is not readily available or exposed by Blender. Essentially, we utilize the `blf` module to get the widths of glyphs and compute the transforms using simple cumulative sums. However, we want widths to exactly match Blender's default text widths, which depends on `VChar.width`, which we don't have. So, we assign an arbitrary size to the `blf` font and compute the ratio between the computed widths and the actual widths using some helper text objects. Notice that underlines are activated because they take the whole width of the glyphs. This process is costly (tens of miliseconds), but the result is cached per font, so I don't really consider this a big issue. The only limitation of the implementation, however, is that BFont is not supported, because it is embedded on the Blender binary and we don't have access to it.

### DIFF
--- a/animation_nodes/nodes/text/decompose_text.py
+++ b/animation_nodes/nodes/text/decompose_text.py
@@ -26,9 +26,6 @@ class DecomposeTextNode(bpy.types.Node, AnimationNode):
     useLineSpacingList : VectorizedSocket.newProperty()
     useAlignmentList : VectorizedSocket.newProperty()
 
-    includeWhiteSpaces : BoolProperty(name = "Include White Spaces", default = True,
-        update = propertyChanged)
-
     decomposeType: EnumProperty(name = "Decompose Type", default = "CHARACTER",
         items = decomposeTypeItems, update = AnimationNode.refresh)
 
@@ -52,6 +49,8 @@ class DecomposeTextNode(bpy.types.Node, AnimationNode):
         self.newInput(VectorizedSocket("Text", "useAlignmentList",
             ("Alignment", "alignment", dict(value = "LEFT", hide = True)),
             ("Alignments", "alignments")))
+        self.newInput("Boolean", "Include Whitespaces", "includeWhitespaces",
+            value = True, hide = True)
 
         self.newOutput("Matrix List", "Transforms", "transforms")
         if self.decomposeType == "CHARACTER":
@@ -63,10 +62,8 @@ class DecomposeTextNode(bpy.types.Node, AnimationNode):
     def draw(self, layout):
         layout.prop(self, "decomposeType", text = "")
 
-    def drawAdvanced(self, layout):
-        layout.prop(self, "includeWhiteSpaces")
-
-    def execute(self, text, fonts, sizes, charsSpacing, wordsSpacing, linesSpacing, alignments):
+    def execute(self, text, fonts, sizes, charsSpacing, wordsSpacing, linesSpacing,
+            alignments, includeWhitespaces):
         self.validateAlignments(alignments)
         if not self.useFontList: fonts = [fonts]
         fontIDs = [self.getFontID(font) for font in fonts]
@@ -79,33 +76,33 @@ class DecomposeTextNode(bpy.types.Node, AnimationNode):
 
         if self.decomposeType == "CHARACTER":
             return self.getCharTransforms(text, fontIDs, fontRatios, sizes,
-                charsSpacing, wordsSpacing, linesSpacing, alignments)
+                charsSpacing, wordsSpacing, linesSpacing, alignments, includeWhitespaces)
         else:
             return self.getWordTransforms(text, fontIDs, fontRatios, sizes,
                 charsSpacing, wordsSpacing, linesSpacing, alignments)
 
     def getCharTransforms(self, text, fontIDs, fontRatios, sizes,
-            charsSpacing, wordsSpacing, linesSpacing, alignments):
+            charsSpacing, wordsSpacing, linesSpacing, alignments, includeWhitespaces):
         allChars = []
         charIndex = 0
-        numberOfSpaces = 0
+        spaceIndex = 0
         allTransforms = Matrix4x4List()
         for i, line in enumerate(text.splitlines()):
             chars = []
             transforms = []
             cumulativeWidth = 0
             for char in line:
-                if self.includeWhiteSpaces or char not in whitespace:
+                if includeWhitespaces or char not in whitespace:
                     scale = Matrix.Scale(sizes[charIndex], 4)
                     translation = Matrix.Translation((cumulativeWidth, -i * linesSpacing[i], 0))
                     transforms.append(translation @ scale)
                     chars.append(char)
                 
                 cumulativeWidth += getCharWidth(char, fontIDs[charIndex], fontRatios[charIndex],
-                    charsSpacing[charIndex], wordsSpacing[numberOfSpaces]) * sizes[charIndex]
+                    charsSpacing[charIndex], wordsSpacing[spaceIndex]) * sizes[charIndex]
 
                 charIndex += 1
-                if char == " ": numberOfSpaces += 1
+                if char == " ": spaceIndex += 1
 
             transforms = Matrix4x4List.fromValues(transforms)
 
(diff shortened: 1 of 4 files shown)


3DSinghVFX:
It is going to be amazing.

JacquesLucke:
> However, we want widths to exactly match Blender's default text widths, which depends on VChar.width, which we don't have.

Can you describe this more precisely? What would happen if you don't do this?

OmarEmaraDev:
> Can you describe this more precisely? What would happen if you don't do this?

The width returned by `blf` is in a drawing unit, which we might consider arbitrary relative to Blender's space, but they are proportional. Now, Blender has its own heuristic to determine the widths of the text objects, I didn't dive deep enough into the source code to determine the exact relations, but we only have access to the width returned by `blf`, so it wouldn't matter anyway. The factor of proportionality is not constant, it differs across fronts. So the only way to compute it is to actually evaluate the Blender object and measure the ratio. Makes sense?

By the way, let me link the reference code from Blender. This is where most of the numbers you see in the code come from.

https://github.com/OmarEmaraDev/blender/blob/master/source/blender/blenkernel/intern/font.c#L1023

OmarEmaraDev:
I could dive deeper into the source code and give you a more concise answer if you want.

JacquesLucke:
I just wonder what would happen if we ignore this additional factor. Does it result in worse quality?

OmarEmaraDev:
Yes, spacing might look very weird for some fonts.

animation_nodes/nodes/text/decompose_text.py
```diff
@@ -0,0 +1,107 @@
...
+            elif align == "RIGHT":
+                offset = -comulativeWidth + charSpacing / 2 - 0.5
+                transforms.transform(Matrix.Translation((offset, 0, 0)))
+
+            allChars += chars
+            allTransforms.extend(transforms)
+
+        return allTransforms, allChars, len(allChars)
+
+    def getCharWidth(self, fontID, char, fontRatio, charSpacing, wordSpacing):
```
JacquesLucke: `getCharWidth` and `getFontRatio` can be pulled out of the class, because they do not depend on `self`. The `lru_cache` might cache the same font multiple times currently, because of the `self` parameter.

animation_nodes/nodes/text/decompose_text.py
```diff
@@ -0,0 +1,107 @@
...
+        width = blf.dimensions(fontID, char)[0] * fontRatio
+        if char == " ":
+            width *= wordSpacing
+        return width + charSpacing / 2 - 0.5
+
+    @lru_cache()
+    def getFontRatio(self, font, fontID):
+        data = bpy.data.curves.new("an_helper_font_curve", type = "FONT")
+        textObject = bpy.data.objects.new("an_helper_text_object", data)
```
JacquesLucke: The `textObject` has to be removed as well.

OmarEmaraDev: `BlendDataCurves.remove` removes the objects by default.

JacquesLucke: I see. I think either you should remove the `textObject` explicitely before removing the curve, or add a comment.

OmarEmaraDev: Alright.

OmarEmaraDev:
An example for word decomposition:

Word Decomposition

enzyme69:
Hi Omar, I tested this Decompose Text node, pretty awesome! Found a little bug:

If you "convert all the text letters into Mesh", and then change the actual string text, the update does not update the first letter somewhat. I need to do a force reset and it is then updating.

I got this bug showing in my latest youtube video.

OmarEmaraDev:
@enzyme69 Seems to be an issue with the Instancer node. I will take a look.

ppaawweeuu:
Hi @OmarEmaraDev . Great job and thanks again.
I found little problem with loading fonts. First time I tried use Decomposite Text I used font already loaded into blend file. Everyting was ok til first saving blend file. After that my geometry disapeared and i saw 'Can't load font'. I had to copy font file from my Windows/Fonts folder into folder with my blend file, load font again traditional way (not using your node), and then problem was solved.
For a moment i thought that it is intentional to block using system depending files.
If I load font using '+' button in your node everything is fine.
I see that fonts in blend file have .filepath relative:
```
>>> bpy.data.fonts["ACaslonPro-Semibold"].filepath
'//..\\..\\..\\..\\..\\Windows\\Fonts\\ACaslonPro-Semibold.otf'
```
and font lodead by your node are absolute:
```
>>> bpy.data.fonts["OpenSans-Regular"].filepath
'C:\\Windows\\Fonts\\OpenSans-Regular.ttf'
```
is it bug or limitation? (once again - thanks)

OmarEmaraDev:
@ppaawweeuu Thanks for the report! Fixed in 9d39fa7.

ppaawweeuu:
Awesome!

noc2art:
Hi Omar;
I have 2 questions regarding Text Decompose, if I may:
- The individual characters produced by the node tend to have their centers on the left side of their respective bounding boxes. Is there a way of adjusting their pivot position without having to convert them to a mesh object?
- As you may know, there's an option among the 'Text Box' options named 'Scale to fit'. Is there a way to get the auto line cut-off and re-scaling on the horizontal and vertical axis to reflect to the AN decomposed text? If not, can its implementation be possible in the new future in AN?
Cheers 👍 
AJ

OmarEmaraDev:
- It is possible to support different character alignments, but we don't have support for that currently. I will add support once I have the time.
- I can't find this option. Can you share more details?

noc2art:
> * It is possible to support different character alignments, but we don't have support for that currently. I will add support once I have the time.
> * I can't find this option. Can you share more details?

That alignment support would be awesome, thank you 👍 

At the very bottom of the text properties you can find that option (all post-2.81 versions and in 2.9+) named "Overflow" under 'Text Boxes' that has 3 options to choose from. Right now, the decompose node can recognize the default, which is 'Overflow'. But can it made to recognize the 'Scale to fit' option which essentially fits the written text into the predefined box dimensions by scaling the character size and making auto line-breaks. We use this option a lot in our workflow and it would be absolutely awesome to have in AN.

image

OmarEmaraDev:
I see. I will put that into consideration.

noc2art:
> I see. I will put that into consideration.

You're awesome. Thank you 👍

Gerstmann-Bradley:
Omar, we always need to call a "font input" node to link the sockets in "decompose text" and "text object output" node.
Would you consider to add an "font" output in decompose text node, so that it spares us some time calling an addition node plus making two linkages?
As a result we just need one linkage instead of 2, and we can also change the font directly in decompose text node.

This may cause a very subtle difference, but I do think it makes node construction much easier for users.

OmarEmaraDev:
@Gerstmann-Bradley I am not sure. The input is virtual vectorized, and there would be ambiguity as to what the output should be, so I don't think it is worth it.

By the way, if you select the decompose node, press W, and select the font from the left menu, an input will be added, set, and connected automatically for you. This should make your workflow easier, I think.

Gerstmann-Bradley:
> @Gerstmann-Bradley I am not sure. The input is virtual vectorized, and there would be ambiguity as to what the output should be, so I don't think it is worth it.
> 
> By the way, if you select the decompose node, press W, and select the font from the left menu, an input will be added, set, and connected automatically for you. This should make your workflow easier, I think.

ok I see, thanks for the reply.

PR #1335, opened 2020-02-23
Added the Offset GP Stroke node.

3DSinghVFX:
I have added the _Offset GP Stroke_ node which allows to offset the _points, strengths, pressures_ and _uv-rotations_ of a stroke basis on the falloff,
Screenshot from 2020-02-23 18-54-18

Example,
Peek 2020-02-23 18-53
Screenshot from 2020-02-23 18-48-51


OmarEmaraDev:
Sorry for the delay. I am taking care of the other pull request and will get back to you as soon as possible.

3DSinghVFX:
Omar, no problem :)

animation_nodes/nodes/gpencil/offset_gp_stroke.py
```diff
@@ -0,0 +1,96 @@
...
+        subrow.prop(self, "useUVRotation", text = "UVRotation", toggle = True)
+
+    def updateSocketVisibility(self):
+        self.inputs[2].hide = not self.useLocation
+        self.inputs[3].hide = not self.useStrength
+        self.inputs[4].hide = not self.usePressure
+        self.inputs[5].hide = not self.useUVRotation
+
+    def execute(self, stroke, falloff, offsetLocations, offsetStrengths, offsetPressures, offsetUVRotations):
+
```
OmarEmaraDev: No need for this white line.

animation_nodes/nodes/gpencil/offset_gp_stroke.py
```diff
@@ -0,0 +1,96 @@
...
+        if not any((self.useLocation, self.useStrength, self.usePressure, self.useUVRotation)):
+            return stroke
+
+        vertices = stroke.vertices.copy()
```
OmarEmaraDev: Why do we copy the vertices here? For `offsetFloatList`, the vertices are only used for evaluation.

animation_nodes/nodes/number/offset_number.pyx
```diff
@@ -0,0 +1,16 @@
+from ... data_structures cimport Falloff, FalloffEvaluator, FloatList, VirtualDoubleList, Vector3DList
+
+def offsetFloatList(Vector3DList vectors, FloatList points, VirtualDoubleList offsets, FalloffEvaluator falloffEvaluator, bint invert = False):
```
OmarEmaraDev: Lets name `points` to `numbers` here.

animation_nodes/nodes/gpencil/offset_gp_stroke.py
3DSinghVFX: If we are offsetting the vertices and strength (say) simultaneously then we have to keep the copy of original vertices for the evaluation for strength, isn't it?

animation_nodes/nodes/number/offset_number.pyx
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/offset_gp_stroke.py
3DSinghVFX: Removed.

animation_nodes/nodes/gpencil/offset_gp_stroke.py
OmarEmaraDev: The `offsetFloatList` doesn't edit the input vertices, right? It is just used for evaluation. On the other hand, `offsetVector3DList` edit the input vertices, but this is the last statement, so it can just edit the vertices directly with no side effects, since we have `dataIsModified = True`. Makes sense?

3DSinghVFX: Thanks, now it is clear.

3DSinghVFX: Done.

animation_nodes/nodes/number/offset_number.pyx
```diff
@@ -0,0 +1,16 @@
+from ... data_structures cimport Falloff, FalloffEvaluator, FloatList, VirtualDoubleList, Vector3DList
```
OmarEmaraDev: I think this should be a `c_utils` file in the `/gpencil` directory. We can move it later if needed.

3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/offset_gp_stroke.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self.newOutput("GPStroke", "Stroke", "stroke")
+
+        self.updateSocketVisibility()
+
+    def draw(self, layout):
+        col = layout.column()
+
+        row = col.row()
+        subrow = row.row(align = True)
```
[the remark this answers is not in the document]
3DSinghVFX: Done.

PR #1338, opened 2020-02-27
Vectorized the Tilt Spline node.

3DSinghVFX:
I have vectorized the Tilt Spline node. Now it works similar to Set Spline Radius node. Fix the issue of the previous Tilt Spline node which changed the tilt of the input spline. 
Screenshot from 2020-02-27 19-21-49
_(Code is adopted from the Set Spline Radius node)_.

### DIFF
--- a/animation_nodes/nodes/spline/tilt_spline.py
+++ b/animation_nodes/nodes/spline/tilt_spline.py
@@ -10,22 +10,65 @@ class TiltSplineNode(bpy.types.Node, AnimationNode):
     bl_label = "Tilt Spline"
 
     useTiltList: VectorizedSocket.newProperty()
+    useSplineList: VectorizedSocket.newProperty()
+
     accumulateTilts: BoolProperty(name = "Accumulate Tilts", default = False,
         update = propertyChanged)
 
     def create(self):
-        self.newInput("Spline", "Spline", "spline", defaultDrawType = "PROPERTY_ONLY")
+        socket = self.newInput(VectorizedSocket("Spline", "useSplineList",
+            ("Spline", "spline"), ("Splines", "splines")))
+        socket.defaultDrawType = "PROPERTY_ONLY"
+        socket.dataIsModified = True
         self.newInput(VectorizedSocket("Float", "useTiltList",
             ("Tilt", "tilt"), ("Tilts", "tilts")))
-        self.newOutput("Spline", "Spline", "spline")
+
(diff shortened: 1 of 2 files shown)


animation_nodes/nodes/spline/tilt_spline.py
```diff
@@ -10,22 +10,65 @@ class TiltSplineNode(bpy.types.Node, AnimationNode):
...
-        return spline
\ No newline at end of file
+        return spline
+
+    def execute_SingleSpline_MultipleTilts(self, spline, tilts):
+        _tilts = VirtualFloatList.fromList(self.prepareRadiusList(tilts, len(spline.points)), 0)
+        tiltSplinePoints(spline, _tilts, self.accumulateTilts)
+        return spline
+
+    def prepareRadiusList(self, tilts, length):
```
OmarEmaraDev: Why is this function needed? Aren't we using a `VirtualFloatList`, which does what this function does?

animation_nodes/nodes/spline/tilt_spline.py
```diff
@@ -10,22 +10,65 @@ class TiltSplineNode(bpy.types.Node, AnimationNode):
...
+
     accumulateTilts: BoolProperty(name = "Accumulate Tilts", default = False,
         update = propertyChanged)
 
     def create(self):
-        self.newInput("Spline", "Spline", "spline", defaultDrawType = "PROPERTY_ONLY")
+        socket = self.newInput(VectorizedSocket("Spline", "useSplineList",
+            ("Spline", "spline"), ("Splines", "splines")))
+        socket.defaultDrawType = "PROPERTY_ONLY"
+        socket.dataIsModified = True
```
OmarEmaraDev: New line here.

animation_nodes/nodes/spline/tilt_spline.py
3DSinghVFX: I also thought about this, but it is in the Set Spline Radius node.

animation_nodes/nodes/spline/tilt_spline.py
3DSinghVFX: Done.

animation_nodes/nodes/spline/tilt_spline.py
3DSinghVFX: Used the `VirtualFloatList`.

PR #1342, opened 2020-02-28
Added Offset Spline Node.

3DSinghVFX:
I have added the _Offset Spline_ node which allows offsetting the spline points, left handles, right handles, radii, and tilts.
Screenshot from 2020-02-29 01-39-09

**Edit:**
Peek 2020-02-29 16-35

Node-Tree:
Screenshot from 2020-02-29 16-35-57


OmarEmaraDev:
It would be nice if you can provide some practical applications with pull requests.

One issue I see here is that the bezier handles are offset based on their location and not the location of their points, which would make offsetting the bezier curve not really useful. So those situations needs to be studied first.

3DSinghVFX:
> It would be nice if you can provide some practical applications with pull requests.

I have edited the initial comment.

> One issue I see here is that the bezier handles are offset based on their location and not the location of their points, which would make offsetting the bezier curve not really useful. So those situations needs to be studied first.

Omar, I think there are two ways we can do this:
1. Only offset the handles after the transformation of the points.
2. We add a bool option when it is enabled which offset the handles after the transformation of the points, and when it is disabled then offset the handles independent of the points.

What do you think?

OmarEmaraDev:
I think we should go with option one.

3DSinghVFX:
> I think we should go with option one.

It is done.

OmarEmaraDev:
hmmm, we are still evaluating at the handles. The falloff should be evaluated at `points`, the handles should be offset based on this evaluation. Moreover, the spline type enum should be removed, you should detect the type of the spline and conditionally offset the handles.

3DSinghVFX:
> hmmm, we are still evaluating at the handles. The falloff should be evaluated at `points`, the handles should be offset based on this evaluation.

Okay.

> Moreover, the spline type enum should be removed, you should detect the type of the spline and conditionally offset the handles.

I wanted to do that but did not find the correct syntax to detect the spline type from the input.

OmarEmaraDev:
```
if spline.type == "BEZIER":
    offset left handle
    offset right handle
```

3DSinghVFX:
Sorry, that I know. I meant was spline type in the `draw` and `updateSocketVisibility` functions.

OmarEmaraDev:
I though we were going to remove the bezier handles offset inputs?

3DSinghVFX:
My bad. My point was to keep the handle offset inputs but offset happens after the transformation of the points.

OmarEmaraDev:
I see. I think this fine control is better left at a lower level. We should offset the points and the handles by the same amount.

3DSinghVFX:
Okay, I'll make the changes. The `offsetVector3DList` function directly offset the input vectors. I can evaluate the falloff with the points using `influences = falloffEvaluator.evaluateList(spline.points)`. Is there a way that I can use the vector list math here to the operation `vectors += offsets * influences`?

OmarEmaraDev:
I think you should make a dedicated function for this. Just like you did with the offset numbers function.

3DSinghVFX:
Okay.

3DSinghVFX:
I have defined that function in the `spline/c_utils.pyx` like this,

```
def offsetVector3DList(Vector3DList vectors, VirtualVector3DList offsets, VirtualFloatList influences, bint invert = False):
    cdef Vector3 *offset
    cdef float influence
    cdef Py_ssize_t i

    for i in range(len(vectors)):
        influence = influences.get(i)

        if invert:
            influence = <float>1 - influence

        offset = offsets.get(i)
        vectors.data[i].x += offset.x * influence
        vectors.data[i].y += offset.y * influence
        vectors.data[i].z += offset.z * influence
```

OmarEmaraDev:
Ok. But lets move this to `/utils/falloffs.pyx`. Also put the offset numbers function there. For input names, remove the "Offset" from the name, I think it is implied.

3DSinghVFX:
Okay. 

Omar, in `offsetFloatList` function, we have used `VirtualDoubleList offsets`. I think we should replace it with `VirtualFloatList offsets`, isn't it?

```def offsetFloatList(Vector3DList vectors, FloatList numbers, VirtualDoubleList offsets, FalloffEvaluator falloffEvaluator, bint invert = False)```

**Comment by OmarEmaraDev** [comment]
The offsets that the user inputs are doubles, so we should keep it as such.

**Comment by 3DSinghVFX** [comment]
Okay.

**Comment by 3DSinghVFX** [comment]
Done.

File: animation_nodes/nodes/spline/offset_spline.py [hunk 1]
```diff
@@ -0,0 +1,83 @@
...
+            _offsetTilts = VirtualDoubleList.create(offsetTilts, 0)
+            offsetFloatList(spline.points, spline.tilts, _offsetTilts, falloffEvaluator)
+
+        if self.useLocation:
+            _offsetLocations = VirtualVector3DList.create(offsetLocations, (0, 0, 0))
+            influences = VirtualFloatList.create(falloffEvaluator.evaluateList(spline.points), 0)
+            offsetVector3DList(spline.points, _offsetLocations, influences)
+            if spline.type == "BEZIER":
+                offsetVector3DList(spline.leftHandles, _offsetLocations, influences)
+                offsetVector3DList(spline.rightHandles, _offsetLocations, influences)
```
OmarEmaraDev: You actually implemented something really nice here. You only evaluated the falloff a single time. But we can take this one step further and use the influences for the rest of the offsets as well. In which case, `offsetFloatList` would need to be rewritten to be similar to `offsetVector3DList`.

However, note that the influences needn't be a virtual list. And the `invert` parameter is redundant in this case as well, so it can be removed.

Also, in this case, it wound't make sense to put the functions in an `falloffs.pyx` file. I would say the structure should be:

`offsetFloats` -> `/nodes/number/c_utils.pyx`
`offset3DVectors` -> `/nodes/vector/c_utils.pyx`

You may update the *Offset GP Stroke* as well if you want.

3DSinghVFX: > In which case, offsetFloatList would need to be rewritten to be similar to offsetVector3DList.
> You may update the Offset GP Stroke as well if you want.

I was just going to ask for this.

Omar, should we change the name of `offsetFloats, offset3DVectors` these functions something else? Because it may confuse with the `offset3DVectors` function from `vector/offset_vector.pyx`.

OmarEmaraDev: The function in `offset_vector.pyx` is called `offsetVector3DList` (Or are you talking about something else?), that's why I was proposing those function names:

`offsetFloats()` -> /nodes/number/c_utils.pyx
`offset3DVectors()` -> /nodes/vector/c_utils.pyx

3DSinghVFX: Ohh sorry, I did not notice that.

3DSinghVFX: Done.

PR #1348, opened 2020-03-06
Added Mesh Points Scatter Node.

3DSinghVFX:
I have added the _Mesh Points Scatter_ node. This node allows scattering random points on any mesh.
Screenshot from 2020-03-06 22-23-22
This node does mesh sampling based on the [Barycentric Coordinate System](http://www.joesfer.com/?p=84) for triangle polygons. So, I have also implemented the Mesh Triangulation method to triangulate the polygons of mesh. 
This node also respects weight which can be Vertex Weight or Falloff weight.
Screenshot from 2020-03-06 22-23-04
Screenshot from 2020-03-06 22-21-39
Screenshot from 2020-03-06 22-33-04


OmarEmaraDev:
I will not review this yet as I still need to study alternative implementations. It should be noted that @JacquesLucke worked on an algorithm to solve this exact problem a while ago. So maybe we can steal some code from him. The relevant code is here: https://github.com/JacquesLucke/blender/blob/functions/source/blender/simulations/bparticles/emitters.cpp#L264-L283

3DSinghVFX:
> I will not review this yet as I still need to study alternative implementations.

Okay.

> It should be noted that @JacquesLucke worked on an algorithm to solve this exact problem a while ago. So maybe we can steal some code from him. The relevant code is here: https://github.com/JacquesLucke/blender/blob/functions/source/blender/simulations/bparticles/emitters.cpp#L264-L283

Omar, I have improved the `randomPairGenerator` function by using on the JacquesLucke's code for random pair generation,
https://github.com/3DSinghVFX/animation_nodes/blob/4e4aae07f1ca9eed305614cf8ab8d3639e81c968/animation_nodes/nodes/mesh/mesh_points_scatter.py#L94-L100

animation_nodes/nodes/mesh/mesh_points_scatter.py
```diff
@@ -0,0 +1,106 @@
...
+    def execute(self, mesh, seed, amount, weights):
+        vertices = mesh.vertices
+        polygons = mesh.polygons
+
+        if len(vertices) == 0 or len(polygons) == 0 or amount == 0:
+            return Vector3DList()
+
+        if self.triangulate: polygons = triangulatePolygons(polygons)
+        if polygons.polyLengths.getMaxValue() > 3:
+            self.raiseErrorMessage("Mesh should have triangle polygons.")
```
OmarEmaraDev: I don't think the mesh should be triangulated or be considered triangular. The algorithm should account for this as needed.

animation_nodes/nodes/mesh/mesh_points_scatter.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return self.randomPointsScatter(vertices, polygons, seed, amount, weights)
+
+    # Mesh Sampling basis on Barycentric Coordinate System: http://www.joesfer.com/?p=84
+    def randomPointsScatter(self, vertices, polygons, seed, amount, weights):
+        areas = trianglePolygonsArea(vertices, polygons)
+        areaMin = areas.getMinValue()
+        if areaMin == 0: self.raiseErrorMessage("Remove doubles.")
```
OmarEmaraDev: I don't think this should be a problem. The algorithm will have to change to account for it.

animation_nodes/nodes/mesh/mesh_points_scatter.py
```diff
@@ -0,0 +1,106 @@
...
+
+        if len(weights) != 0:
+            usePolyWeight = True
+            polyAmount = len(polygons)
+            polyWeights = FloatList(length = polyAmount)
+            weights = VirtualDoubleList.create(weights, 1).materialize(polyAmount)
+            for i, polygon in enumerate(polygons):
+                polyWeights[i] = (weights[polygon[0]] + weights[polygon[1]] + weights[polygon[2]]) / 3
+        else:
+            usePolyWeight = False
```
OmarEmaraDev: Weights should always be used. You already use a virtual list with a default of 1, so this should work even if `len(weights) == 0`.

animation_nodes/nodes/mesh/mesh_points_scatter.py
```diff
@@ -0,0 +1,106 @@
...
+                    probabilities.append(i)
+
+        lenProb = len(probabilities)
+        if lenProb == 0: return Vector3DList()
+        points = Vector3DList(length = amount)
+        for i in range(amount):
+            randIndex = probabilities[int(uniformRandomDoubleWithTwoSeeds(seed + i, self.nodeSeed, 0, lenProb))]
+            polygon = polygons[randIndex]
+            points[i] = self.randomPolyPoint(i, vertices, polygon, seed)
+        return points
```
OmarEmaraDev: One problem with this approach is that you will often need to go to memory to get the triangle. An alternative better algorithm will put caching into consideration and generate all points of a single triangle successively.

animation_nodes/nodes/mesh/mesh_points_scatter.py
3DSinghVFX: Okay.

3DSinghVFX:
> Still not a full review, just some points that I need to make.

Okay.

> Also, why is the core algorithm in python? Shouldn't we utilize Cython here?

Can I use the `uniformRandomDoubleWithTwoSeeds` in Cython?

animation_nodes/nodes/mesh/mesh_points_scatter.py
3DSinghVFX: Omar, we do not know prior to the total number of points on a polygon.

1. We can cache polygon (or its vertices by polygon index) then use it for the points generations.
2. Or, we split this loop into two: first, calculate how many points each polygon has then in the second loop generates those points.

What do you think?

OmarEmaraDev:
> Can we use the `uniformRandomDoubleWithTwoSeeds` in Cython?

Yes, or even better, you can use a more specialized function, check the `algorithms/random.pxd` file.

animation_nodes/nodes/mesh/mesh_points_scatter.py
OmarEmaraDev: I don't think (1) would be very useful in this case.
We can try (2). The best thing you can do is time all of this. Time all the methods you can think of and see which one is best.

3DSinghVFX: Omar, with the method(2), I have implemented the main algorithum in the Cython.

animation_nodes/nodes/mesh/mesh_points_scatter.py
3DSinghVFX: Done.

animation_nodes/nodes/mesh/mesh_points_scatter.py
3DSinghVFX: Done.

JacquesLucke:
It might also be good to implement a new random number sequence generator in cython. This could also improve the performance of nodes like Random Number and Random Vector when they generate long lists.

I think for this kind of algorithm you also need to generate many random numbers. Maybe some algorithm from https://en.m.wikipedia.org/wiki/Xorshift is good enough. We have to check if the distribution is random enough for us, but in my last tests it worked well.

OmarEmaraDev:
@JacquesLucke Sounds good. What do you think about the general approach of this patch?

@3DSinghVFX You should split the scatter function into multiple smaller function where it makes sense. Are you working on a version without pre-triangulating the mesh? Did you time your new implementation?

JacquesLucke:
@OmarEmaraDev The API should be somewhat similar to https://developer.blender.org/diffusion/B/browse/master/source/blender/blenlib/BLI_rand.h. Just a bit more object oriented. So there could e.g. be a `XorshiftGenerator` class.

I'm not sure if we can reasonably sample a mesh without converting it to triangles first. However, maybe this conversion could be implicit (without actually computing the triangles). I'm not sure.

OmarEmaraDev:
Ok. Yes, implicit triangulation is what I meant here. It seems unnecessary to pre-triangulate while we can just do the computation putting in consideration that a polygon can be multiple triangles.

JacquesLucke:
The only difficulty might occur when there are concave polygons. Not sure if we absolutely have to deal with that, yet. Those might require more complex triangulation algorithms.

OmarEmaraDev:
I see. Good point.

3DSinghVFX:
> @3DSinghVFX You should split the scatter function into multiple smaller function where it makes sense. 

Okay.

> Are you working on a version without pre-triangulating the mesh?

Sorry, what do you mean without pre-triangulating the mesh? We can combine the calculation of triangulation, area, and triangle-vertices in a single-loop.

> Did you time your new implementation?

In Python, it was `~135ms` for `25000` points. Now in Cython, it is `~29.5ms` for `25000` points.

Another thing is, the last loop in `points_scatter.pyx` the file is on _polygons_, so it takes almost the same time `~29ms` for `25000` and `100000` points,
**25000 points:**
Screenshot from 2020-03-08 15-46-02

**100000 points:**
Screenshot from 2020-03-08 15-46-25

OmarEmaraDev:
I mean something like this:
```
for polygon in polygons:
    for triangle in polygon.triangles:
        compute the area and store it.
```

3DSinghVFX:
> I mean something like this:

It is done. I have refactored the code.

3DSinghVFX:
> I am still not entirely happy about the structure of the code.

I have improved the structure of the code, please have look.

>     * I don't think we should compute data earlier than needed. For instance, we shouldn't compute `triVertices` and pass it to the sample function, instead just loop over the triangles again in `sampleRandomPoints`.

Changed.

>     * Don't add a condition for triangles vs polygons, they should share the same code.

Removed.

>     * For computing minimum and maximum, we typically assign the initial value as the first value of the array, that way we don't have to use a magic number like `1000000`. Also, consider the case where the weights are all zeros.

Changed.

>     * Try to avoid appending and extending lists if you know the length in advance.

Now, it is only used once.

>     * The final seed should be computed in python and passed to the Cython code. Maybe abstract this away in a `seedFromTwoSeeds` function.

Added.

>     * `cumulativeProbabilities` seems to be a misnomer, as the output is not cumulative nor a probability.

Renamed it as `trianglesDistribution`.

>     * Split the `sampleRandomPoints` function further, just like you had before and just like Jacques does in his code.

Omar, I have tested by splitting the random pairs in a separate function (shown below) but it slows down the node alot, 
```
def sampleRandomPoints(Vector3DList vertices, PolygonIndicesList polygons, LongList totalTriPoints,
                       Py_ssize_t distLength, Py_ssize_t seed, Py_ssize_t pointAmount):
    cdef Vector3DList points = Vector3DList(length = pointAmount)
    cdef UIntegerList polyLengths = polygons.polyLengths
    cdef Py_ssize_t polyAmount = polygons.getLength()
    cdef Py_ssize_t i, j, k, triangleIndex, index, polyIndex1, polyIndex2, polyIndex3, randSeed
    cdef Vector3 v1, v2, v3, v
    cdef double p1, p2, p3

    index = 0
    triangleIndex = 0
    for i in range(polyAmount):
        polygon = polygons[i]
        polyIndex1 = polygon[0]
        v1 = vertices.data[polyIndex1]
        for j in range(polyLengths.data[i] - 2):
            polyIndex2 = polygon[1 + j]
            polyIndex3 = polygon[2 + j]

            v2 = vertices.data[polyIndex2]
            v3 = vertices.data[polyIndex3]
            randSeed = int(randomDouble_Range(i + j + seed, 0, distLength))
            for k in range(totalTriPoints.data[triangleIndex]):
                p1 = randomDouble_Range(i + j + k + seed + randSeed, 0.0, 1.0)
                p2 = randomDouble_Range(i + j + k + 2 * seed + randSeed + 100, 0.0, 1.0)
                if p1 + p2 > 1.0:
                    p1 = 1.0 - p1
                    p2 = 1.0 - p2
                p3 = 1.0 - p1 - p2

                v.x = p1 * v1.x + p2 * v2.x + p3 * v3.x
                v.y = p1 * v1.y + p2 * v2.y + p3 * v3.y
                v.z = p1 * v1.z + p2 * v2.z + p3 * v3.z
                points.data[index] = v
                index += 1
            triangleIndex += 1
    return points

def getRandomPairs(Py_ssize_t seed):
    cdef double p1, p2
    p1 = randomDouble_Range(seed, 0.0, 1.0)
    p2 = randomDouble_Range(2 * seed, 0.0, 1.0)
    if p1 + p2 > 1.0:
        p1 = 1.0 - p1
        p2 = 1.0 - p2
    return p1, p2, 1.0 - p1 - p2
```
**With the `getRandomPairs` function:**
b283_outfn_randpoints

**Without the `getRandomPairs` function** (https://github.com/3DSinghVFX/animation_nodes/blob/27c201b733cb1c41431777b94b07dd9a6771d0fb/animation_nodes/nodes/mesh/points_scatter.pyx#L119-L125)
b283_embeded_randpoints

>     * Move the triangleArea function into somewhere common.

Moved.
 
>     * Clean the c_utils file as it now have unused functions.

Cleaned.

OmarEmaraDev:
Generate the annotated code like we did before and try to spot the areas where there is overhead. For functions that will only be called in the cython file, define them using `cdef` not `def`. Also, return a `Vector3` instead of a tuple.

JacquesLucke:
I think this should really use a different random number sequence generator that does not need this this "seed hacking".

3DSinghVFX:
> Generate the annotated code like we did before and try to spot the areas where there is overhead. For functions that will only be called in the cython file, define them using `cdef` not `def`.

I have defined the functions with `cdef`, it improved the performance.

>Also, return a `Vector3` instead of a tuple.

Omar, I have tried to return a `Vector3` but it does not improve the performance. Am I doing something wrong here?

```
cdef sampleRandomPoints(Vector3DList vertices, PolygonIndicesList polygons, LongList totalTriPoints,
                       Py_ssize_t distLength, Py_ssize_t seed, Py_ssize_t pointAmount):
    cdef Vector3DList points = Vector3DList(length = pointAmount)
    cdef UIntegerList polyLengths = polygons.polyLengths
    cdef Py_ssize_t polyAmount = polygons.getLength()
    cdef Py_ssize_t i, j, k, triangleIndex, index, polyIndex1, polyIndex2, polyIndex3, randSeed
    cdef Vector3 v1, v2, v3, v, vr
    cdef double p1, p2, p3

    index = 0
    triangleIndex = 0
    for i in range(polyAmount):
        polygon = polygons[i]
        polyIndex1 = polygon[0]
        v1 = vertices.data[polyIndex1]
        for j in range(polyLengths.data[i] - 2):
            polyIndex2 = polygon[1 + j]
            polyIndex3 = polygon[2 + j]

            v2 = vertices.data[polyIndex2]
            v3 = vertices.data[polyIndex3]
            randSeed = int(randomDouble_Range(i + j + seed, 0, distLength))
            for k in range(totalTriPoints.data[triangleIndex]):
                vr = getRandomPairs(i + j + k + seed + randSeed)
                p1 = vr.x
                p2 = vr.y
                p3 = vr.z

                v.x = p1 * v1.x + p2 * v2.x + p3 * v3.x
                v.y = p1 * v1.y + p2 * v2.y + p3 * v3.y
                v.z = p1 * v1.z + p2 * v2.z + p3 * v3.z
                points.data[index] = v
                index += 1
            triangleIndex += 1
    return points

cdef getRandomPairs(Py_ssize_t seed):
    cdef double p1, p2
    cdef Vector3 v

    p1 = randomDouble_Range(seed, 0.0, 1.0)
    p2 = randomDouble_Range(2 * seed, 0.0, 1.0)
    if p1 + p2 > 1.0:
        p1 = 1.0 - p1
        p2 = 1.0 - p2
    v.x = p1
    v.y = p2
    v.z = 1.0 - p1 - p2
    return v
```

3DSinghVFX:
> I think this should really use a different random number sequence generator that does not need this this "seed hacking".

Sorry for the late reply. Now, I use a single array of random points `randomPoints`, 
```
p1 = randomPoints
p2 = randomPoints.reverse 
```
https://github.com/JacquesLucke/animation_nodes/blob/e933a208c5e1731eab79c5f7aa58a2ee1dcb8906/animation_nodes/nodes/mesh/points_scatter.pyx#L98-L137

3DSinghVFX:
Hi Omar,
I have added the _getTrianglePolygons_ method to the Mesh Object.

3DSinghVFX:
Hi @OmarEmaraDev 
I have implemented the advanced _**Ear Clipping**_ method for mesh triangulation especially to handle concave polygons. This method gives mesh triangulation similar to _Trinagulate_ modifier with _Quad Type = Beauty_. This method remove mesh [ear based the angle](https://arxiv.org/abs/1212.6038) which gives nice triangulated mesh with minimal silver triangle polygons. This method can also hangle the invert normal polygons.

I have tested this method on these meshes with location and rotation set to wiggle animation:
EarCut6

YorsLosilla:
It will be very cool to have this on master.

3DSinghVFX:
@OmarEmaraDev Hi... :)
I have found a way to use weight as the density for points which gives the similar result as we use vertex weight for particles:
**Mesh Points Scatter node:**
Screenshot from 2020-04-18 13-43-19

**Particles with  Vertex Weight:**
Screenshot from 2020-04-18 13-51-59

YorsLosilla:
Please put this is in master soon. I can´t imagine a feature more important than this.

OmarEmaraDev:
@YorsLosilla We will, I just need to finish college and I will start reviewing again. Sorry for the delay.

YorsLosilla:
Sorry @OmarEmaraDev  didn´t wanted to bother. It´s just that I´m really excited with this node and the packed spheres. Those open a lot of oportunities to create cool stuff with AN

3DSinghVFX:
Here are results with new _points-distribution_ method (https://github.com/JacquesLucke/animation_nodes/pull/1348/commits/23700e761c55f961fb4249b739a358c88774929e): 
2_NewOld

animation_nodes/nodes/mesh/points_scatter.pyx
```diff
@@ -0,0 +1,165 @@
...
+
+    cdef XoShiRo256StarStar rng = XoShiRo256StarStar(seed)
+    cdef Py_ssize_t j
+    for i in range(offsetAmount):
+        for j in range(polyAmount):
+            index = rng.nextIntWithMax(polyAmount)
+            triPointAmount = distribution.data[index]
+            if triPointAmount == 0: continue
+            distribution.data[index] = triPointAmount + step
+            break
```
OmarEmaraDev, line 119: It is guaranteed that the computed amount would be more than the needed about, since we are taking the ceiling, right?

3DSinghVFX:
@OmarEmaraDev Should we move the `points_scatter.pyx` file to the `algorithms/mesh` folder?

OmarEmaraDev:
@3DSinghVFX Sure.

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: That is true when we do not use `weight`. If we use weight (not as density) then `currentAmount <= pointAmount` so we need these conditions.

OmarEmaraDev, line 119: Why is this loop necessary? You are essentially breaking after the first iteration anyway.

animation_nodes/nodes/mesh/points_scatter.pyx
```diff
@@ -0,0 +1,165 @@
...
+
+    # Use weight as density.
+    if useWeightForDensity:
+        expectedAmount = 0
+        for i in range(polyAmount):
+            triPointAmount = int(ceil(distribution.data[i] * max(triWeights.data[i], 0)))
+            distribution.data[i] = triPointAmount
+            expectedAmount += triPointAmount
+    return distribution, expectedAmount
```
OmarEmaraDev, line 128: Why can't we join this with the other loop above? It seems to me `useWeightForDensity` grants us an early exist from the function.

animation_nodes/nodes/mesh/points_scatter.pyx
```diff
@@ -0,0 +1,165 @@
...
+
+cdef Vector3DList sampleRandomPoints(Vector3DList vertices, PolygonIndicesList polygons, LongList distribution,
+                                     Py_ssize_t seed, Py_ssize_t polyAmount, Py_ssize_t expectedAmount):
+    cdef DoubleList randomPoints = DoubleList(length = expectedAmount)
+    cdef XoShiRo256Plus rng = XoShiRo256Plus(seed)
+    cdef Py_ssize_t i
+    for i in range(expectedAmount):
+        randomPoints.data[i] = rng.nextDouble()
```
OmarEmaraDev, line 136: Why fill an array with random values while you can just call `rng.nextDouble()` inside the loop?

animation_nodes/nodes/mesh/points_scatter.pyx
```diff
@@ -0,0 +1,165 @@
...
+cdef distributionOfPoints(Py_ssize_t polyAmount, FloatList triAreas, FloatList triWeights,
+                          bint useWeightForDensity, Py_ssize_t seed, Py_ssize_t pointAmount):
+    # Initial distribution of points.
+    cdef double triAreaTotal = 0
+    cdef Py_ssize_t i
+    for i in range(polyAmount):
+        triAreaTotal += triAreas.data[i]
+
+    if triAreaTotal == 0: triAreaTotal = 1
+    triAreaTotal = pointAmount / triAreaTotal
```
OmarEmaraDev, line 83: Don't reuse variables here. Just create a new variable and name it something appropriate. `pointsPerUnitArea` maybe?

3DSinghVFX:
@OmarEmaraDev When I use `weight` (not as density) sometimes points fluctuate: 
Peek 2020-06-19 23-21

Node-Tree: 
untitled3

animation_nodes/nodes/mesh/mesh_points_scatter.py
```diff
@@ -0,0 +1,60 @@
...
+        if polygons.polyLengths.getMaxValue() > 3:
+            if self.methodType:
+                mesh.triangulateMesh(method = "EAR")
+                polygons = mesh.polygons
+            else:
+                mesh.triangulateMesh(method = "FAN")
+                polygons = mesh.polygons
+
+        weights = VirtualDoubleList.create(weights, 1)
+        seed  = (seed * 674523 + self.nodeSeed * 3465284) % 0x7fffffff
```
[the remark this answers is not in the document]
3DSinghVFX: Done.

animation_nodes/nodes/mesh/points_scatter.pyx
```diff
@@ -0,0 +1,165 @@
...
+@cython.cdivision(True)
+cdef calculateTriangleWeightsAreas(Vector3DList vertices, PolygonIndicesList polygons, VirtualDoubleList weights):
+    cdef Py_ssize_t vertexAmount = vertices.length
+    cdef float weightMax = 1
+    cdef float weight
+    cdef Py_ssize_t i
+
+    for i in range(vertexAmount):
+        weight = weights.get(i)
+        if weight > weightMax: weightMax = weight
```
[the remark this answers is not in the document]
3DSinghVFX: I used `cdef DoubleList realWeights = weights.realList`.
This gave me an error: 
Screenshot from 2020-06-19 23-38-52

animation_nodes/nodes/mesh/points_scatter.pyx
```diff
@@ -0,0 +1,165 @@
...
+    DoubleList,
+    UIntegerList,
+    Vector3DList,
+    VirtualDoubleList,
+    PolygonIndicesList
+    )
+
+def randomPointsScatter(Vector3DList vertices, PolygonIndicesList polygons, VirtualDoubleList weights,
+                        bint useWeightForDensity, Py_ssize_t seed, Py_ssize_t pointAmount):
+    cdef FloatList triWeights, triAreas
```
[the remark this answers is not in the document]
3DSinghVFX: Done.

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: Loop is needed when we use `weights` because some polygons do not have any vertex due to zero weight.

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: The reason is, first we make sure each polygon has correct number of vertices so that sum is equal to total points. Then we impose the weight to reduce the number of points per polygon according to their weight.

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: Okay.

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: If we do not use array we have to create `2 * expectedAmount` random numbers. Because we need two random numbers to get a random point on polygon. However with an array, we need only `expectedAmount` random points and we use something like this:
```
            p1 = randomPoints.data[index]
            p2 = randomPoints.data[expectedAmount - index - 1]
            if p1 + p2 > 1.0:
                p1 = 1.0 - p1
                p2 = 1.0 - p2
            p3 = 1.0 - p1 - p2
```

OmarEmaraDev: Ok, but why reuse random numbers?

animation_nodes/nodes/mesh/points_scatter.pyx
OmarEmaraDev: Why can't you use an if statement then? I find it very weird that you would unconditional break inside a loop.

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: For performance.

animation_nodes/nodes/mesh/points_scatter.pyx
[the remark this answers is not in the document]
OmarEmaraDev: Can you check what the available attributes are?

animation_nodes/nodes/mesh/points_scatter.pyx
3DSinghVFX: We do not know prior to whether a polygon has a point or not. However, it can be done for `use WeightForDensity` case because in that each polygon has at least one point due to `ceil`.

animation_nodes/nodes/mesh/points_scatter.pyx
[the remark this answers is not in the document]
3DSinghVFX: It has `realList` attribute
```
cdef class VirtualDoubleList_List(VirtualDoubleList):
    cdef DoubleList realList
    cdef double *realData
    cdef Py_ssize_t realLength

    def __cinit__(self, DoubleList realList):
        self.realList = realList
        self.realData = realList.data
        self.realLength = realList.length
        assert self.realLength > 0
```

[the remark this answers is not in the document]
3DSinghVFX: Should we materialize the virtual list in the `mesh_points_scatter.py` file then pass to the function?

OmarEmaraDev:
I want to clarify the behavior of the node a bit. From what I understand, we have two modes of operations. Can you verify that?

## Total Amount

The number of output points is equal to the input amount. For each triangle, we compute a weight that is equal to the average input vertices weights multiplied by the area. The sum of weights is computed. For each triangle, we estimate the number of points based on the weights, sum of weights, and the input amount. The estimates are rounded upward. Finally, we randomly reduce the amounts until the sum reaches the input amount.

## Amount Per Unit Area

For each triangle, we compute a weight that is equal to the average input vertices weights multiplied by the area. For each triangle, we compute the number of points by multiplying the weight by the input about.

3DSinghVFX:
> I want to clarify the behavior of the node a bit. From what I understand, we have two modes of operations. Can you verify that?
> ## Total Amount
> 
> The number of output points is equal to the input amount. For each triangle, we compute a weight that is equal to the average input vertices weights multiplied by the area. The sum of weights is computed. For each triangle, we estimate the number of points based on the weights, sum of weights, and the input amount. The estimates are rounded upward. Finally, we randomly reduce the amounts until the sum reaches the input amount.

a) For each triangle, I compute the `points Per Unit Area = Input Points / Total Area`. Then multiply it with the average triangle weight (average vertices weights of triangle), so `Points Per Polygon = int(ceil(triangleAreas.data[i] * pointsPerUnitArea * max(triangleWeights.data[i], 0)))`. This lead to fluctuation problem that I metioned above in case of weights (not as density).

b) So, now I have changed it which is the same as you have described that resolve that problem. 

> ## Amount Per Unit Area
> 
> For each triangle, we compute a weight that is equal to the average input vertices weights multiplied by the area. For each triangle, we compute the number of points by multiplying the weight by the input about.

Yes.

**Current Method:**
## When _Use Weight As Density_ disable
triangle weight: average weight computed from triangle vertices weight e.g., vertex weight

1) We compute the `total weight =  sum the triangle weight * triangle area`
2) `Points per unit area = Input Amount / total weight`
3)  `Points per triangle =  Points per unit area * triangle area`
4) If the sum of points-per-triangle does not equal to Input Amount, we add or remove randomly points to make them equal.

## When _Use Weight As Density_ enable
1) We compute the `total weight =  sum the triangle area`
2) `Points per unit area = Input Amount / total weight`
3)  `Points per triangle =  Points per unit area * triangle area`
4) If the sum of points-per-triangle does not equal to Input Amount, we add or remove randomly to make them equal.
5)  `New Points per triangle =  Points per unit area * triangle weight`
 
### I think we can combine both of these by modifying the step 4):
1) We compute the `total weight =  sum the triangle weight * triangle area`
2) `Points per unit area = Input Amount / total weight`
3) `Points per triangle =  Points per unit area * triangle area`
4) - _Use Weight As Density_ disable: If the sum of points-per-triangle does not equal to Input Amount, we **add or remove** randomly points to make them equal.
    - _Use Weight As Density_ enable: If the sum of points-per-triangle greater is than Input Amount, we **remove** randomly points to make them equal.

OmarEmaraDev:
@3DSinghVFX Why not unify step 1? That is, for each triangle, we compute a weight that is equal to the average input vertices weights multiplied by the area. Then if density mode is enabled, compute and return the distribution, existing early. After the if, compute the case where density is disabled.

```
for each triangle:
    compute its weight

if density mode:
    for each triangle:
        compute number of points (weight * amount)
    return distribution

compute sum of weights

for each triangle:
    estimate number of points ceil((amount / sum of weights) * weight)

for each extra point:
    subtract 1 from a random element

return distribution
```

I am still not sure about the add or remove thing. I believe remove only would work. I will take another look tomorrow.

3DSinghVFX:
> @3DSinghVFX Why not unify step 1? That is, for each triangle, we compute a weight that is equal to the average input vertices weights multiplied by the area. Then if density mode is enabled, compute and return the distribution, existing early.

I have already tried that. It has two problems, 1) number of points can greater than number of input points 2) Small change in weight cause large fluctuation in weight.

> I am still not sure about the add or remove thing. I believe remove only would work. I will take another look tomorrow.

Okay.

> ## Edit
> ### I think we can combine both of these by modifying the step 4):
> 
> We also need to change steps 1) and 3). It is working.
> 
>     1. * _Use Weight As Density_ disable: We compute the `total weight =  sum the triangle weight * triangle area`
>        * _Use Weight As Density_ enable: We compute the `total weight =  sum the triangle area`
> 
>     2. `Points per unit area = Input Amount / total weight`
> 
>     3. * _Use Weight As Density_ disable: `Points per triangle =  Points per unit area * triangle area`
>        * _Use Weight As Density_ enable: `Points per triangle =  Points per unit area * triangle area * triangle weight`
> 
>     4. * _Use Weight As Density_ disable: If the sum of points-per-triangle does not equal to Input Amount, we **add or remove** randomly points to make them equal.
>        * _Use Weight As Density_ enable: If the sum of points-per-triangle greater is than Input Amount, we **remove** randomly points to make them equal.

This method does not work. 

We need step 5) for _Use Weight For Density_:
1)  - _Use Weight As Density_ disable: We compute the `total weight =  sum the triangle weight * triangle area`.
     - _Use Weight As Density_ enable: We compute the `total weight =  sum the triangle area`.
2) `Points per unit area = Input Amount / total weight`.
3)  `Points per triangle =  Points per unit area * triangle area`.
4) If the sum of points-per-triangle does not equal to Input Amount, we **add or remove** randomly points to make them equal. If use weight as density disable then return distribution.
5)  `New Points per triangle =  Points per unit area * triangle weight`.

OmarEmaraDev:
@3DSinghVFX I just tested an implementation of my proposed code and it seemed to work. Can you take a look and test it yourself?

```cython
import cython
from ... math cimport Vector3
from libc.math cimport sqrt, ceil
from ... algorithms.random_number_generators cimport XoShiRo256Plus, XoShiRo256StarStar
from ... data_structures cimport (
    LongList,
    FloatList,
    DoubleList,
    UIntegerList,
    Vector3DList,
    VirtualDoubleList,
    PolygonIndicesList
)

def randomPointsScatter(Vector3DList vertices, PolygonIndicesList polygons, VirtualDoubleList weights,
                        bint useWeightsAsDensity, long seed, long numberOfPoints):
    cdef DoubleList triangleWeights = calculateTriangleWeights(vertices, polygons, weights)

    cdef LongList distribution
    cdef long computedNumberOfPoints
    distribution = computeDistribution(triangleWeights, numberOfPoints, useWeightsAsDensity, seed,
        &computedNumberOfPoints)

    return sampleRandomPoints(vertices, polygons, distribution, computedNumberOfPoints, seed)

@cython.cdivision(True)
cdef DoubleList calculateTriangleWeights(Vector3DList vertices, PolygonIndicesList polygons, VirtualDoubleList weights):
    cdef DoubleList triangleWeights = DoubleList(length = polygons.getLength())
    cdef Py_ssize_t i, j, i1, i2, i3
    for i in range (polygons.getLength()):
        j = 3 * i
        i1 = polygons.indices.data[j]
        i2 = polygons.indices.data[j + 1]
        i3 = polygons.indices.data[j + 2]

        triangleWeights.data[i] = triangleArea(vertices.data + i1, vertices.data + i2, vertices.data + i3)
        triangleWeights.data[i] *= (weights.get(i1) + weights.get(i2) + weights.get(i3)) / 3.0
    return triangleWeights

@cython.cdivision(True)
cdef double triangleArea(Vector3 *v1, Vector3 *v2, Vector3 *v3):
    cdef Vector3 vs1, vs2, vc
    vs1.x = v1[0].x - v3[0].x
    vs1.y = v1[0].y - v3[0].y
    vs1.z = v1[0].z - v3[0].z

    vs2.x = v2[0].x - v3[0].x
    vs2.y = v2[0].y - v3[0].y
    vs2.z = v2[0].z - v3[0].z

    vc.x = vs1.y * vs2.z - vs1.z * vs2.y
    vc.y = vs1.z * vs2.x - vs1.x * vs2.z
    vc.z = vs1.x * vs2.y - vs1.y * vs2.x
    return sqrt(vc.x * vc.x + vc.y * vc.y + vc.z * vc.z) / 2.0

@cython.cdivision(True)
cdef LongList computeDistribution(DoubleList triangleWeights, long numberOfPoints, bint useWeightsAsDensity, long seed,
                                  long *outputNumberOfPoints):
    cdef long numberOfTriangles = triangleWeights.length
    cdef LongList distribution = LongList(length = numberOfTriangles)
    cdef long computedNumberOfPoints, triangleNumberOfPoints
    cdef Py_ssize_t i

    if useWeightsAsDensity:
        computedNumberOfPoints = 0
        for i in range(numberOfTriangles):
            triangleNumberOfPoints = <long>(triangleWeights.data[i] * numberOfPoints)
            distribution.data[i] = triangleNumberOfPoints
            computedNumberOfPoints += triangleNumberOfPoints
        outputNumberOfPoints[0] = computedNumberOfPoints
        return distribution

    cdef double sumOfWeights = 0.0
    for i in range(numberOfTriangles):
        sumOfWeights += triangleWeights.data[i]

    cdef double numberOfPointsPerUnitWeight = numberOfPoints / sumOfWeights
    computedNumberOfPoints = 0
    for i in range(numberOfTriangles):
        triangleNumberOfPoints = <long>ceil(numberOfPointsPerUnitWeight * triangleWeights.data[i])
        distribution.data[i] = triangleNumberOfPoints
        computedNumberOfPoints += triangleNumberOfPoints

    cdef XoShiRo256StarStar rng = XoShiRo256StarStar(seed)
    cdef Py_ssize_t randomIndex
    for i in range(computedNumberOfPoints - numberOfPoints):
        while True:
            randomIndex = rng.nextLongWithMax(numberOfTriangles)
            triangleNumberOfPoints = distribution.data[randomIndex]
            if triangleNumberOfPoints != 0:
                distribution.data[randomIndex] -= 1
                break

    outputNumberOfPoints[0] = numberOfPoints
    return distribution

cdef Vector3DList sampleRandomPoints(Vector3DList vertices, PolygonIndicesList polygons, LongList distribution,
                                     long numberOfPoints, long seed):
    cdef XoShiRo256Plus rng = XoShiRo256Plus(seed)
    cdef Vector3DList points = Vector3DList(length = numberOfPoints)
    cdef Py_ssize_t i, j, k, index
    cdef Vector3 v1, v2, v3, v
    cdef double p1, p2, p3
    index = 0
    for i in range(polygons.getLength()):
        j = i * 3
        v1 = vertices.data[polygons.indices.data[j]]
        v2 = vertices.data[polygons.indices.data[j + 1]]
        v3 = vertices.data[polygons.indices.data[j + 2]]
        for k in range(distribution.data[i]):
            p1 = rng.nextDouble()
            p2 = rng.nextDouble()
            if p1 + p2 > 1.0:
                p1 = 1.0 - p1
                p2 = 1.0 - p2
            p3 = 1.0 - p1 - p2

            v.x = p1 * v1.x + p2 * v2.x + p3 * v3.x
            v.y = p1 * v1.y + p2 * v2.y + p3 * v3.y
            v.z = p1 * v1.z + p2 * v2.z + p3 * v3.z
            points.data[index] = v
            index += 1

    return points
```

3DSinghVFX:
> @3DSinghVFX I just tested an implementation of my proposed code and it seemed to work. Can you take a look and test it yourself?

Omar, It is working fine for `useWeightsAsDensity` is disabled and no input weight. 
When I use weight then it has problem, points fluctuate:
Peek 2020-06-20 18-21

And, in the case of `useWeightsAsDensity` is enabled, it gives points more than `Amount`:
 
BetterWieghtScatter3

Those are the problems that I mentioned before.

To solve those problems:
>     1. * _Use Weight As Density_ disable: We compute the `total weight =  sum the triangle weight * triangle area`.
>        * _Use Weight As Density_ enable: We compute the `total weight =  sum the triangle area`.

We have to compute the total weight separately.
 
>     2. `Points per unit area = Input Amount / total weight`.
> 
>     3. `Points per triangle =  Points per unit area * triangle area`.
> 
>     4. Adjust the points if needed. Return in case of _Use Weight As Density_ disable.
>
>     5. `New Points per triangle =  Points per unit area * triangle weight` and return the distribution.

This method is working fine.

OmarEmaraDev:
I thought the point of the `useWeightsAsDensity` was to treat the input amount as an amount per unit weight. In this case, it is undesirable to have the the input amount equal to the output number of points.

Can you send the test file that suffers from fluctuations?

3DSinghVFX:
> I thought the point of the `useWeightsAsDensity` was to treat the input amount as an amount per unit weight. In this case, it is undesirable to have the the input amount equal to the output number of points.

Not _input amount equal to the output number of points_. We set input points, if weight (as density) is 1 for polygons then we have number of points is equal to input amount. Otherwise we have less number points based on the weight (as density) of polygons, e.g.,:
BetterWieghtScatter4
  
> Can you send the test file that suffers from fluctuations?

[BetterWieghtScatter3.zip](https://github.com/JacquesLucke/animation_nodes/files/4807935/BetterWieghtScatter3.zip)

OmarEmaraDev:
Than's because you normalize the area. The density is the number of points per area, not per normalized area.

OmarEmaraDev:
The fluctuations is probably just uninitialized memory. This happens when all weights are zero. In this case, we have two options, either return an empty list, or return a list of zero. Both solutions should be implemented by decrementing the computed number of points and returning it.

```cython
    cdef XoShiRo256StarStar rng = XoShiRo256StarStar(seed)
    cdef Py_ssize_t randomIndex
    for i in range(computedNumberOfPoints - numberOfPoints):
        while True:
            randomIndex = rng.nextLongWithMax(numberOfTriangles)
            triangleNumberOfPoints = distribution.data[randomIndex]
            if triangleNumberOfPoints != 0:
                distribution.data[randomIndex] -= 1
                computedNumberOfPoints -= 1
                break

    outputNumberOfPoints[0] = computedNumberOfPoints
    return distribution
```
In case of an empty list output, this is all that is needed. In case of zero list, you have to check if the returned computed number of points is zero, and if yes, return a zero list.

3DSinghVFX:
> Than's because you normalize the area. The density is the number of points per area, not per normalized area.

Yes. But if we do not normalize the area then we can have any number of points e.g., if scale the object the number of points will increase:
Screenshot from 2020-06-20 19-13-44

> Than's because you normalize the area. 

Also, this is the way vertex weight works in particle-system as density. I think we should normalize the area.

OmarEmaraDev:
Ok. Sounds reasonable. What will your implementation look like?

3DSinghVFX:
> Ok. Sounds reasonable. What will your implementation look like?

This is the way I implemented in the current method:
- Compute points per polygon using `polyArea/TotalArea`.
- Remove the extra-points.
- Then use triangle weight to compute the points or weighted points per polygon `points per polygon * triangle weight`.

_Also we have to normalized the weight as well for the both cases._
Do we have a method for list normalization?

3DSinghVFX:
> The fluctuations is probably just uninitialized memory. This happens when all weights are zero. In this case, we have two options, either return an empty list, or return a list of zero. Both solutions should be implemented by decrementing the computed number of points and returning it.
> 
> ```cython
>     cdef XoShiRo256StarStar rng = XoShiRo256StarStar(seed)
>     cdef Py_ssize_t randomIndex
>     for i in range(computedNumberOfPoints - numberOfPoints):
>         while True:
>             randomIndex = rng.nextLongWithMax(numberOfTriangles)
>             triangleNumberOfPoints = distribution.data[randomIndex]
>             if triangleNumberOfPoints != 0:
>                 distribution.data[randomIndex] -= 1
>                 computedNumberOfPoints -= 1
>                 break
> 
>     outputNumberOfPoints[0] = computedNumberOfPoints
>     return distribution
> ```
> 
> In case of an empty list output, this is all that is needed. In case of zero list, you have to check if the returned computed number of points is zero, and if yes, return a zero list.

Okay. We can return a list with zeros.

OmarEmaraDev:
Why do we need to normalize the weight in the standard case?

Commit your changes so that I can take a look. I just want an implementation that doesn't complicate one case just because the other case has a small deviation.

3DSinghVFX:
> Why do we need to normalize the weight in the standard case?

Otherwise, it will create more extra points that we have to remove.

> Commit your changes so that I can take a look. I just want an implementation that doesn't complicate one case just because the other case has a small deviation.

Okay.

OmarEmaraDev:
> Otherwise, it will create more extra points that we have to remove.

How will this happen? We are already using the sum of weights as a factor of proportionality, normalizing should have no effect.

3DSinghVFX:
> > Otherwise, it will create more extra points that we have to remove.
> 
> How will this happen? We are already using the sum of weights as a factor of proportionality, normalizing should have no effect.

But not in the case of Weight As Density. We will use weight after the _extra points removal_ step.

OmarEmaraDev:
@3DSinghVFX What makes you think Blender uses the Density mode you suggested as the particle emitter distributor? My tests shows otherwise. Moreover, the [source code](https://github.com/JacquesLucke/blender/blob/functions/source/blender/simulations/bparticles/emitters.cpp#L270-L272) also suggest otherwise.

As for the property that as the objects gets bigger, the more points we get, I think this is a desirable property, that's the point of a density-based distribution.

Ours and theirs.

3DSinghVFX:
> @3DSinghVFX What makes you think Blender uses the Density mode you suggested as the particle emitter distributor? My tests shows otherwise. Moreover, the [source code](https://github.com/JacquesLucke/blender/blob/functions/source/blender/simulations/bparticles/emitters.cpp#L270-L272) also suggest otherwise.
>
> As for the property that as the objects gets bigger, the more points we get, I think this is a desirable property, that's the point of a density-based distribution.
> 
> Ours and theirs.
> 
> 

Hi, 
Here are the distribution particles with weight as density on two planes, both planes have the same number of particles.
Plane's scale (1,1,1) left-side; Plane's scale (3,3,3) right-side
Screenshot from 2020-06-21 09-55-48
As far as I understand, this means points-density per polygon is not changing with the scale of an object. The _Use As Density_ option also gives this result.

> As for the property that as the objects gets bigger, the more points we get, I think this is a desirable property, that's the point of a density-based distribution.

If scale the object, the number of emission particles does not change:
Peek 2020-06-21 10-24

OmarEmaraDev:
@3DSinghVFX That's my point. Blender doesn't use a density approach here, and instead the standard approach where the number of points is defined is used.

Then why do we have a density mode that doesn't act as a density mode? How exactly is it different from the standard approach and in what applications is it useful?

3DSinghVFX:
> @3DSinghVFX That's my point. Blender doesn't use a density approach here, and instead the standard approach where the number of points is defined is used.

But the previous approach you suggested does not use defined number of points, weight as density output points more than the input number of points.

> Then why do we have a density mode that doesn't act as a density mode?

I think we should change the name.
1. Remove the _Use As Density_ option.
2. By default, weight use as Density (for _Use As Density_ is enabled).
3. Added new option _Conserve Input Amount_ or something like that (for _Use As Density_ is disabled).

> How exactly is it different from the standard approach and in what applications is it useful?

Weight allows the user to define the region on the mesh to distribute points in two ways: (1) Conserve the number of points (Amount) (2) Or a density which calculated from Amount.
It can be used to distribute trees/rocks/buildings/diamonds/etc on a mesh with the aforementioned methods.

OmarEmaraDev:
@3DSinghVFX Can you provide a practical example where the Density per normalized area is required?

3DSinghVFX:
> @3DSinghVFX Can you provide a practical example where the Density per normalized area is required?

It is an alternate option for distribution. I think we can remove that.


I have noticed that for a small number of particles, the distribution of points is not following the gradient of weights:
BetterWieghtScatter4


A large number of particles:
BetterWieghtScatter5

OmarEmaraDev:
@3DSinghVFX Perhaps visual inspection is not the best way to judge the distribution. I really don't want to drag this patch further. If you don't mind, can we merge the most basic version now and work on more advanced options later?

3DSinghVFX:
> @3DSinghVFX Perhaps visual inspection is not the best way to judge the distribution.

Here is an example with the same weights but particle-system:
BetterWieghtScatter6

> I really don't want to drag this patch further. If you don't mind, can we merge the most basic version now and work on more advanced options later?

Okay. I'll remove the _Use As Density_ option.

3DSinghVFX:
> Thanks! For the more advanced options, we will work on them after the next release.

:) Sure.

PR #1351, opened 2020-03-08
added Bmesh Reverse Faces

Bombaba:
Hi,

Since I noticed that `Invert Normal` option in `Bmesh Recalculate Normals` node no longer works, I added another node `Bmesh Reverse Faces`, which will invert all bmesh face normals.
This node is quite simple but useful for some situation, so I hope this node be merged into main branch.

Thanks!


animation_nodes/nodes/mesh/bmesh_recalculate_face_normals.py
```diff
@@ -7,18 +7,13 @@ class BMeshRecalculateFaceNormalsNode(bpy.types.Node, AnimationNode):
...
         self.newInput("BMesh", "BMesh", "bm").dataIsModified = True
         self.newOutput("BMesh", "BMesh", "bm")
 
-    def draw(self, layout):
-        layout.prop(self, "invert")
-
     def getExecutionCode(self, required):
         recalcString = "bmesh.ops.recalc_face_normals(bm, faces = bm.faces)"
-        return [recalcString] * 2 if self.invert else recalcString
+        return recalcString
```
OmarEmaraDev: Just return directly in this case.

animation_nodes/nodes/mesh/bmesh_reverse_faces.py
```diff
@@ -0,0 +1,24 @@
+import bpy, bmesh
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types import AnimationNode
+
+class BMeshReverseFacesNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_BMeshReverseFacesNode"
+    bl_label = "BMesh Reverse Faces"
```
OmarEmaraDev: Change the node name into something more familiar. `BMesh Invert Normals` maybe.

animation_nodes/nodes/mesh/bmesh_reverse_faces.py
```diff
@@ -0,0 +1,24 @@
...
+
+    flip_multires: BoolProperty(name = "Flip Multi Resolution", update = executionCodeChanged)
```
OmarEmaraDev: Use camelCase (`flipMultires`). What does this option do?

animation_nodes/nodes/mesh/bmesh_reverse_faces.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def create(self):
+        self.newInput("BMesh", "BMesh", "bm").dataIsModified = True
+        self.newOutput("BMesh", "BMesh", "bm")
+
+    def draw(self, layout):
+        layout.prop(self, "flip_multires")
+
+    def getExecutionCode(self, required):
+        recalcString = f"bmesh.ops.reverse_faces(bm, faces = bm.faces, flip_multires = {self.flip_multires})"
```
OmarEmaraDev: Changing the execution code here is unnecessary. You can reference the property directly, `flip_multires = self.flipMultires`. Also return directly here.

Bombaba:
@OmarEmaraDev 
Thanks you.
I refactored the codes according to your advice.
As for the option `flip Multires`, I just exposed the identical option of `bmesh.ops.reverse_face`.
Please check for https://docs.blender.org/api/current/bmesh.ops.html#bmesh.ops.reverse_faces.

animation_nodes/nodes/mesh/bmesh_invert_normals.py
```diff
@@ -0,0 +1,23 @@
+import bpy
+from bpy.props import *
+from ... events import executionCodeChanged
+from ... base_types import AnimationNode
+
+class BMeshInvertNormalsNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_BMeshInvertNormalsNode"
+    bl_label = "BMesh Invert Normals"
+
+    flipMultires: BoolProperty(name = "Flip Multires", update = executionCodeChanged)
```
OmarEmaraDev: The execution code no longer changes, so `executionCodeChanged` shouldn't be used here. Instead, you should use `propertyChanged`.

animation_nodes/nodes/mesh/bmesh_recalculate_face_normals.py
```diff
@@ -1,24 +1,17 @@
-import bpy, bmesh
+import bpy
 from bpy.props import *
```
OmarEmaraDev: This import is also not needed.

Bombaba:
@OmarEmaraDev 
Ok, I changed the code as you said.

OmarEmaraDev:
To clarify, I want the flip option to be a boolean socket that is hidden by default. Just for flexibility since we don't know what it is.

Bombaba:
Like this?

PR #1360, opened 2020-03-14
Fixed the Triangulate BMesh node.

3DSinghVFX:
I have fixed the _Triangulate BMesh_ node. Due to changes in the [Blender API](https://docs.blender.org/api/current/bmesh.ops.html#bmesh.ops.triangulate), node showed an error which is now fixed: 
Screenshot from 2020-03-14 22-39-17


OmarEmaraDev:
Thanks!

PR #1378, opened 2020-03-25
Added fibonacci distribution

harisreedhar:
Added fibonacci distribution in distribute matrices node
fibdemo

### DIFF
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -1,7 +1,7 @@
 import bpy
 from bpy.props import *
 from math import pi as _pi
-from libc.math cimport sin, cos
+from libc.math cimport sin, cos, sqrt
 from ... utils.limits cimport INT_MAX
 from ... base_types import AnimationNode
 from ... events import executionCodeChanged
@@ -16,7 +16,8 @@ modeItems = [
     ("GRID", "Grid", "", "NONE", 1),
     ("CIRCLE", "Circle", "", "NONE", 2),
     ("MESH", "Mesh", "", "NONE", 3),
-    ("SPIRAL", "Spiral", "", "NONE", 4)
+    ("SPIRAL", "Spiral", "", "NONE", 4),
+    ("FIBONACCI", "Fibonacci", "", "NONE", 5)
 ]
 
 distanceModeItems = [
@@ -88,6 +89,13 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
+        elif self.mode == "FIBONACCI":
+            self.newInput("Integer", "Amount", "amount", value = 100, minValue = 1)
+            self.newInput("Integer", "Center mask", "centermask",value = 0, minValue = 0)
+            self.newInput("Float", "Angle", "angle", value = 137.5)
+            self.newInput("Float", "Start size", "startsize", value = 1.0)
+            self.newInput("Float", "End size", "endsize", value = 1.0) 
+            self.newInput("Float", "Overall Scale", "scale", value = 1.0)     
 
         self.newOutput("Matrix List", "Matrices", "matrices")
         self.newOutput("Vector List", "Vectors", "vectors")
@@ -129,6 +137,8 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
                 yield "matrices = self.execute_Polygons(mesh)"
         elif self.mode == "SPIRAL":
             yield "matrices = self.execute_Spiral(amount, startRadius, endRadius, startSize, endSize, startAngle, endAngel)"
+        elif self.mode == "FIBONACCI":
+            yield "matrices = self.execute_Fibonacci(amount, centermask, angle, startsize, endsize, scale)"    
 
         if "vectors" in required:
             yield "vectors = AN.nodes.matrix.c_utils.extractMatrixTranslations(matrices)"
@@ -248,6 +258,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             rotateStep(&iCos, &iSin, stepCos, stepSin)
 
         return matrices
+
+    def execute_Fibonacci(self, Py_ssize_t amount, Py_ssize_t centermask, float angle, float startsize, float endsize, float scale):
+        cdef Py_ssize_t i 
+
+        if centermask >= amount:
+            centermask = amount
+        i = centermask 
+
+        cdef float factor = 1 / <float>((amount-centermask) - 1) if (amount-centermask) > 1 else 0
+        cdef float golden_angle = angle * (PI/180)
+        cdef float r, theta, size, f
+        cdef Vector3 vector
+        cdef Matrix4x4List matrices = Matrix4x4List(length = amount-centermask)
+
+        while i < amount:
+            f = <float>(i-centermask) * factor
+            size = f * (endsize - startsize) + startsize
+            
+            theta = i * golden_angle
+            r = sqrt(i)/sqrt(amount)*scale
+
+            vector.x = cos(theta)*r
+            vector.y = sin(theta)*r
+            vector.z = 0
+
+            setTranslationMatrix(matrices.data + i-centermask, &vector)
+            scaleMatrix3x3Part(matrices.data + i-centermask, size)
+            i+=1
+        return matrices 
 
 cdef int limitAmount(n):
     return max(min(n, INT_MAX), 0)
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -90,11 +90,11 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
         elif self.mode == "FIBONACCI":
-            self.newInput("Integer", "Amount", "amount", value = 100, minValue = 1)
-            self.newInput("Integer", "Center mask", "centermask",value = 0, minValue = 0)
+            self.newInput("Integer", "Amount", "amount", value = 100, minValue = 0)
+            self.newInput("Integer", "Center Mask", "centerMask",value = 0, minValue = 0)
             self.newInput("Float", "Angle", "angle", value = 137.5)
-            self.newInput("Float", "Start size", "startsize", value = 1.0)
-            self.newInput("Float", "End size", "endsize", value = 1.0) 
+            self.newInput("Float", "Start Size", "startSize", value = 1.0)
+            self.newInput("Float", "End Size", "endSize", value = 1.0) 
             self.newInput("Float", "Overall Scale", "scale", value = 1.0)     
 
         self.newOutput("Matrix List", "Matrices", "matrices")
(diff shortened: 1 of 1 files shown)


animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -88,6 +89,13 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
+        elif self.mode == "FIBONACCI":
+            self.newInput("Integer", "Amount", "amount", value = 100, minValue = 1)
+            self.newInput("Integer", "Center mask", "centermask",value = 0, minValue = 0)
```
3DSinghVFX: Change `Center mask -> Center Mask`, and `centermask -> centerMask`. Do the same for other variables.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
 
         return matrices
 
+    def execute_Fibonacci(self, Py_ssize_t amount, Py_ssize_t centermask, float angle, float startsize, float endsize, float scale):
+        cdef Py_ssize_t i 
+
+        if centermask >= amount:
+            centermask = amount
```
3DSinghVFX: You can write something like `centerMask = min(centerMask, amount)`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+        i = centermask 
+
+        cdef float factor = 1 / <float>((amount-centermask) - 1) if (amount-centermask) > 1 else 0
+        cdef float golden_angle = angle * (PI/180)
```
3DSinghVFX: Change the name, add spaces, `goldenAngle = angle * (PI / 180)`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+        cdef float r, theta, size, f
+        cdef Vector3 vector
+        cdef Matrix4x4List matrices = Matrix4x4List(length = amount-centermask)
+
+        while i < amount:
+            f = <float>(i-centermask) * factor
```
3DSinghVFX: Also here `(i - centerMask)`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            size = f * (endsize - startsize) + startsize
+            
+            theta = i * golden_angle
+            r = sqrt(i)/sqrt(amount)*scale
```
3DSinghVFX: Here `r = sqrt(i) / sqrt(amount) * scale`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+
+            vector.x = cos(theta)*r
+            vector.y = sin(theta)*r
```
3DSinghVFX: Add spaces, ` * `.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            vector.z = 0
+
+            setTranslationMatrix(matrices.data + i-centermask, &vector)
+            scaleMatrix3x3Part(matrices.data + i-centermask, size)
+            i+=1
```
3DSinghVFX: Add spaces, 
```
            setTranslationMatrix(matrices.data + i - centerMask, &vector)
            scaleMatrix3x3Part(matrices.data + i - centerMask, size)
            i += 1 
```

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+
+        if centermask >= amount:
+            centermask = amount
+        i = centermask 
+
+        cdef float factor = 1 / <float>((amount-centermask) - 1) if (amount-centermask) > 1 else 0
+        cdef float golden_angle = angle * (PI/180)
+        cdef float r, theta, size, f
+        cdef Vector3 vector
+        cdef Matrix4x4List matrices = Matrix4x4List(length = amount-centermask)
```
3DSinghVFX: Add spaces ` - `.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,35 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
         return matrices
 
+    def execute_Fibonacci(self, Py_ssize_t amount, Py_ssize_t centermask, float angle, float startsize, float endsize, float scale):
+        cdef Py_ssize_t i 
+
+        if centermask >= amount:
+            centermask = amount
+        i = centermask 
+
+        cdef float factor = 1 / <float>((amount-centermask) - 1) if (amount-centermask) > 1 else 0
```
3DSinghVFX: Always use space on both sides of an operation,
```
cdef float factor = 1 / <float>((amount - centermask) - 1) if (amount - centermask) > 1 else 0
```

harisreedhar: @3DSinghVFX Updated changes

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -88,6 +89,13 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
+        elif self.mode == "FIBONACCI":
+            self.newInput("Integer", "Amount", "amount", value = 100, minValue = 0)
+            self.newInput("Integer", "Center Mask", "centerMask",value = 0, minValue = 0)
```
OmarEmaraDev: Space after `,`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -88,6 +89,13 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            self.newInput("Float", "Angle", "angle", value = 137.5)
```
OmarEmaraDev: Angles should be in radians, not degrees. It would be good to add the golden angle to the Constants node to be used with this node.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,33 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
 
         return matrices
 
+    def execute_Fibonacci(self, Py_ssize_t amount, Py_ssize_t centerMask, float angle, float startSize, float endSize, float scale):
```
OmarEmaraDev: Break this long line.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,33 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+        centerMask = min(centerMask, amount)
+
+        cdef Py_ssize_t i = centerMask
+        cdef float factor = 1 / <float>((amount - centerMask) - 1) if (amount - centerMask) > 1 else 0
+        cdef float goldenAngle = angle * (PI / 180)
+        cdef float r, theta, size, f
+        cdef Vector3 vector
+        cdef Matrix4x4List matrices = Matrix4x4List(length = amount - centerMask)
+
+        while i < amount:
```
OmarEmaraDev: Use a for loop.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,33 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            f = <float>(i - centerMask) * factor
+            size = f * (endSize - startSize) + startSize
+            
+            theta = i * goldenAngle
+            r = sqrt(i) / sqrt(amount) * scale
+
+            vector.x = cos(theta) * r
+            vector.y = sin(theta) * r
+            vector.z = 0
```
OmarEmaraDev: Can we adjust the code to avoid computing sine the cosine inside the loop? The spiral code we have already does this by incrementally rotating some initial point. See the 2D rotation matrix as a reference. Let me know if you need more details.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -249,6 +259,33 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+        cdef float r, theta, size, f
+        cdef Vector3 vector
+        cdef Matrix4x4List matrices = Matrix4x4List(length = amount - centerMask)
+
+        while i < amount:
+            f = <float>(i - centerMask) * factor
+            size = f * (endSize - startSize) + startSize
+            
+            theta = i * goldenAngle
+            r = sqrt(i) / sqrt(amount) * scale
```
OmarEmaraDev: `sqrt(amount)` seems to be a constant, so we should take it outside of the loop.

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: @OmarEmaraDev  Yes need bit more explanation. I didn't understand it's working. or an example maybe helpful

OmarEmaraDev:
@harisreedhar First lets write a minimal unoptimized snippet of code to generate points of a unit circle.

```python
angle = 0
angleStep = (2 * PI) / amount
for i in range(amount):
    x = cos(angle)
    y = sin(angle)
    angle += angleStep
```

The problem here is that we compute a `sin` and a `cos` inside the loop, which is something we can avoid. How can we avoid that? We can think about it in two different ways, I will describe both.

## Sum Of Angles Identity

We know that:

```python
cos(a + b) = cos(a)cos(b) - sin(a)sin(b)
sin(a + b) = sin(a)cos(b) + sin(b)cos(a)
```

Since `angle` of the current iteration is equal to `angle + angleStep` of the previous iteration, we can redefine `x` and `y` using the equations of sum of angles. In this case, `b` is constant and so is its sine and cosine, and `a` and its sine and cosine are known from the last iteration. So we can rewrite the code as follows:

```python
iCos = 1
iSin = 0
angleStep = (2 * PI) / amount
stepCos = cos(angleStep)
stepSin = sin(angleStep)

for i in range(amount):
    x = iCos
    y = iSin
    newCos =  iCos * stepCos -  iSin * stepSin
    iSin = stepSin * iCos + stepCos * iSin
    iCos = newCos
```

Notice that we start with `x = cos(0) = 1` and `y = sin(1) = 0`.

## 2D Rotation

Another way we can think about this is that the circle is generated by rotating an initial point with a certain step angle in each iteration. When multiplying the 2D rotation matrix with a variable vector, we get:

```python
newX = x * cos(angle) - y * sin(angle)
newY = x * sin(angle) + y * cos(angle)
```

Guess what, this is exactly the same equations as the angle sum ones. So the code will be exactly the same. It just have a geometrical interpretation, as opposed to an algebraic one.

harisreedhar:
@OmarEmaraDev Thanks a lot for detailed explanation. I need some time to understand this. i'm closing this request for now. will create a new when i'm done

PR #1380, opened 2020-03-26
Fixed Grease Pencil nodes for latest Blender API changes.

3DSinghVFX:
I have fixed the Grease Pencil nodes for the latest [Blender 2.83 API](https://wiki.blender.org/wiki/Reference/Release_Notes/2.83/Grease_Pencil) changes.
Changes are:
1. Mainly related to the layer mask.
2. Small bug fixes.
3. Added _Hardeness_ attribute to GP Stroke.

### DIFF
--- a/animation_nodes/nodes/gpencil/gp_layer_from_frames.py
+++ b/animation_nodes/nodes/gpencil/gp_layer_from_frames.py
@@ -23,8 +23,8 @@ class GPLayerFromFramesNode(bpy.types.Node, AnimationNode):
         self.newInput("Float", "Tint Factor", "tintFactor", value = 0, minValue = 0, maxValue = 1, hide = True)
         self.newInput("Float", "Stroke Thickness", "lineChange", hide = True)
         self.newInput("Integer", "Pass Index", "passIndex", value = 0, minValue = 0, hide = True)
-        self.newInput(VectorizedSocket("Text", "useMaskLayerList",
-            ("Mask Layer", "maskLayerName"), ("Mask Layers", "maskLayerNames")), hide = True)
+        self.newInput(VectorizedSocket("GPLayer", "useMaskLayerList",
+            ("Mask Layer", "maskLayer"), ("Mask Layers", "maskLayers")), hide = True)
         self.newInput(VectorizedSocket("Boolean", "useInvertMaskLayerList",
             ("Invert Mask Layer", "invertMaskLayer"), ("Invert Mask Layers", "invertMaskLayers")),
             value = False, hide = True)
@@ -32,11 +32,11 @@ class GPLayerFromFramesNode(bpy.types.Node, AnimationNode):
         self.newOutput("GPLayer", "Layer", "layer")
 
     def execute(self, frames, layerName, blendMode, opacity, tintColor, tintFactor, lineChange, passIndex,
-                maskLayerNames, invertMaskLayers):
+                maskLayersIn, invertMaskLayers):
         if not self.useFrameList:
             frames = [frames]
         if not self.useMaskLayerList:
-            maskLayerNames = [maskLayerNames]
+            maskLayersIn = [maskLayersIn]
 
         frameNumbers = [frame.frameNumber for frame in frames]
         if len(frameNumbers) != len(set(frameNumbers)):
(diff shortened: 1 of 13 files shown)


3DSinghVFX:
@OmarEmaraDev I have added the _Vertex Color_ attribute to the _GP Stroke_ object.

We can add _Vertex Color_ input in these nodes which would be very handy,
Screenshot from 2020-03-29 23-51-51

What do you think?

OmarEmaraDev:
@3DSinghVFX I don't see how vertex colors could be incorporated in those nodes, aside from the offset node. Can you elaborate?

3DSinghVFX:
> @3DSinghVFX I don't see how vertex colors could be incorporated in those nodes, aside from the offset node. Can you elaborate?

If we integrate the _Vertex Color_ in _Offset GP Stroke_  node then it is not needed for remaining nodes (especially the _GP Stroke From Spline_ and _Transform GP Stroke_ nodes). However, we can integrate _Vertex Color_ in _Replicate GP Stroke_ node, it will allow the replication of stroke with different vertex colors.

OmarEmaraDev:
I don't think we should add that to the Replicate node. The Offset node would be fine I guess.

3DSinghVFX:
> I don't think we should add that to the Replicate node. The Offset node would be fine I guess.

Okay. I'll make changes.

3DSinghVFX:
@OmarEmaraDev I have added the _offset color_ method with three basic color methods _ADD, MULTIPLY, MIX_ in `color/c_utils` file, and added the _Vertex Color_ offset option in the _Offset GP Stroke_ node.

3DSinghVFX:
Okay. I have done testing on the released Blender 2.83 build, it is working fine.

To get multiple layers by their name, we have to use multiple _GP Object Input_ nodes. Should we add _Get GP Layer From List_ node that will give a layer by name from the layer-list?
Screenshot from 2020-06-04 11-36-00

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
```diff
@@ -1,36 +1,71 @@
...
+                if self.isEmptyLayer(maskLayer): continue
+                maskLayerName = maskLayer.layerName
+                if maskLayerName != "" and maskLayerName != layerName:
+                    maskLayer.invertAsMask = invertMaskLayers[i]
+                    maskLayers.append(maskLayer)
+
+        return GPLayer(layerName, frames, blendMode, opacity, useLights, tintColor, tintFactor, lineChange,
+                       passIndex, False, maskLayers)
+
+    def isEmptyLayer(self, layer):
```
OmarEmaraDev, line 63: You can move this to the GPLayer class as a class method, notice that it is also used in `set_gp_layer_attributes.py` Also, you can do something like this:
```
def isEmptyLayer(self):
    for frame in self.frames:
        if len(frame.strokes) > 0:
            return False
    return True
```

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
```diff
@@ -1,39 +1,44 @@
...
-    def execute(self, vertices, strengths, pressures, uvRotations, lineWidth, drawCyclic,
-                startCapMode, endCapMode, materialIndex, displayMode):
+    def execute(self, vertices, strengths, pressures, uvRotations, vertexColors, lineWidth,
+                hardness, drawCyclic, startCapMode, endCapMode, materialIndex, displayMode):
 
         amount = len(vertices)
         strengths = FloatList.fromValues(VirtualDoubleList.create(strengths, 1).materialize(amount))
         pressures = FloatList.fromValues(VirtualDoubleList.create(pressures, 1).materialize(amount))
         uvRotations = FloatList.fromValues(VirtualDoubleList.create(uvRotations, 0).materialize(amount))
+        vertexColors = ColorList.fromValues(VirtualColorList.create(vertexColors, Color((0, 0, 0, 0))).materialize(amount))
```
OmarEmaraDev, line 41: The output of `materialize` is already a ColorList, so no need for `ColorList.fromValues` here.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
3DSinghVFX: Okay.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
3DSinghVFX: Done. BTW why class method not as regular method?

animation_nodes/nodes/gpencil/gp_stroke_from_points.py
3DSinghVFX: Done.

3DSinghVFX:
> @3DSinghVFX Not in this patch. Lets talk about this later.

Okay.

animation_nodes/nodes/gpencil/gp_layer_from_frames.py
OmarEmaraDev: @3DSinghVFX My apologies, I meant a normal method.

3DSinghVFX: Okay, I'll change that.

Should I removed this `if len(maskLayersIn) > 0`?. I added for the performance:
https://github.com/JacquesLucke/animation_nodes/blob/5c8005435cd226c5f3783899b7137bd0f4795fcf/animation_nodes/nodes/gpencil/gp_layer_from_frames.py#L50-L58

OmarEmaraDev: @3DSinghVFX  It shouldn't make much of a difference. It is up to you.

3DSinghVFX: Okay. I'll keep that.

PR #1391, opened 2020-04-10
Fix the None material issue of the Material Output node and updated Add menu.

3DSinghVFX:
I have fixed the _None_ material issue of the Material Output node,
Screenshot from 2020-04-10 23-18-13

Added the _Vector Noise_ node in the _Add_ menu under the _Vector_ category.


OmarEmaraDev:
Can you update the CHANGELOG?

Issue #1394, opened 2020-04-14
Question: Spline Falloff with Radii option


8BitsGotBite:
Would it be possible to use the Radii of a Spline, in the Spline Falloff node or would that be technically impossible?

8BitsGotByte:
That would be fantastic. Thanks a million :)

OmarEmaraDev:
@8BitsGotBite Feel free to provide feedback in #1394.

OmarEmaraDev:
Implemented in 3920e2f.

PR #1397, opened 2020-04-16
Added Use Spline Radius option in the Spline Falloff node.

3DSinghVFX:
I have added the _Use Spline Radius_ option in the _Spline Falloff_ node which allow to set falloff width based on the spline radius (https://github.com/JacquesLucke/animation_nodes/issues/1394).

Screenshot from 2020-04-17 01-05-13

Example:
Screenshot from 2020-04-17 01-04-44

untitled3

### DIFF
--- a/animation_nodes/nodes/falloff/spline_falloff.pyx
+++ b/animation_nodes/nodes/falloff/spline_falloff.pyx
@@ -35,11 +35,12 @@ class SplineFalloffNode(bpy.types.Node, AnimationNode):
             ("Spline", "spline", socketProps),
             ("Splines", "splines", socketProps)))
 
+        self.newInput("Boolean", "Use Radius for Distance", "useSplineRadiusForDistance", value = False)
         self.newInput("Float", "Distance", "distance", value = 0)
+        self.newInput("Boolean", "Use Radius for Width", "useSplineRadiusForWidth", value = False)
         self.newInput("Float", "Width", "width", value = 1, minValue = 0)
         self.newInput("Interpolation", "Interpolation", "interpolation",
             defaultDrawType = "PROPERTY_ONLY")
-        self.newInput("Boolean", "Use Spline Radius", "useSplineRadius", value = False)
 
         self.newOutput("Falloff", "Falloff", "falloff")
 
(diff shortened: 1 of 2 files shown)


animation_nodes/nodes/falloff/spline_falloff.pyx
```diff
@@ -101,10 +108,12 @@ cdef class SplineFalloff(BaseFalloff):
         cdef float parameter = self.spline.project_LowLevel(<Vector3*>point)
         self.spline.evaluatePoint_LowLevel(parameter, &closestPoint)
         cdef float distance = distanceVec3(<Vector3*>point, &closestPoint)
+        cdef float radius = 0
+        if self.useSplineRadius: radius = self.spline.evaluateRadius(parameter)
```
OmarEmaraDev, line 112: Use `evaluateRadius_LowLevel` here.

3DSinghVFX: Okay.

3DSinghVFX:
>One thing I want to note though. Should we multiply the radius with the width instead of adding it? The advantage here would to that the width will act as factor of how much the radius affects the falloff width, as opposed to being an offset with no control. 

Yes, and _offset_ width effect can be achieved by controlling the radius of the spline. 

>Perhaps we should consult the user who suggested the feature.

Yes.

Should we add an option that would allow whether to effect distance or width or both with the spline's radius?

OmarEmaraDev:
What would this option be like? Two booleans?

3DSinghVFX:
> What would this option be like? Two booleans?

A text input or _EnumProperty_.

3DSinghVFX:
> What would this option be like? Two booleans?

I think is two boolean is fine.

OmarEmaraDev:
Lets just roll this version and wait for feedback.

3DSinghVFX:
> Lets just roll this version and wait for feedback.

Okay.

8BitsGotBite:
You guys are fast. Great work <3.

> The advantage here would to that the width will act as factor of how much the radius affects the falloff width, as opposed to being an offset with no control.

The best thing would be to have the option in the node to just switch the behavior. Is that doable?

3DSinghVFX:
> You guys are fast. Great work <3.

You're welcome :)

> The best thing would be to have the option in the node to just switch the behavior. Is that doable?

Yes, it is doable, here is the node:
Screenshot from 2020-04-17 20-49-11

8BitsGotBite:
Perfect :)

PR #1398, opened 2020-04-18
Added mesh triangulation methods, and the Triangulate Mesh node.

3DSinghVFX:
I have added the _Fan Span_ and _Ear Clipping_ methods for mesh triangulation. And added the _Triangulate Mesh_ node that allows triangulating the mesh,
In the advanced settings of the node, we can toggle between the _Fan Span_ and _Ear Clipping_ methods.
Example:

Screenshot from 2020-04-18 23-03-17

Screenshot from 2020-04-18 23-06-14

I have tested the _Ear Clipping_ method on these meshes:
79075117-5fe3cc00-7d0e-11ea-91ff-988592909fa8


OmarEmaraDev:
Can you provided the resources used to write this implementation? I am not familiar with the method myself, so if you have some good resources, I would appreciate it.

3DSinghVFX:
> Can you provided the resources used to write this implementation? I am not familiar with the method myself, so if you have some good resources, I would appreciate it.

References for _Ear Clipping_ method: https://arxiv.org/abs/1212.6038, https://www.coursera.org/lecture/computational-geometry/4-5-an-ear-cutting-algorithm-c3d7Y

These are important steps that I found during implementation:
1. Polygon should be in counter-close wise.
2. Then calculate the _inner_ angle at all vertices.
3. If a vertex says E is convex (small inner angle), and any other vertices do not lie in the triangle (E-1, E, E+1) formed by the vertex E. Then E is an _Ear_.
4. Then, store the triangle indices and remove the vertex E.
5. Repeat steps 2-4 until we left with only three vertices. 

Polygon Transform: I used _Rodrigues' rotation formula_ to rotate the polygon-vertices to XY-plane.

3DSinghVFX:
I have noticed that if we have a _weird topology_ mesh and triangulate it (using _Triangulate Mesh_ (except _Fan Span_ method which works in some cases) or _Triangulate Bmesh_ nodes (except _Fixed_ method which works in some cases)). It is possible that two triangle-polygons (or more) from different original polygons (of input mesh) can share the same vertices. Here are examples,

**Triangulate Mesh Node**:
Screenshot from 2020-04-20 15-36-29

**Triangulate BMesh Node**:
Screenshot from 2020-04-20 15-34-20

_To resolve this we have to either skip mesh validation, and the Blender does not care about the duplicate polygons (and can be removed with the Weld modifier) or add a method which removes the duplicated triangle-polygons._

**SkipValidation**
Screenshot from 2020-04-20 15-44-28

JacquesLucke:
Unfortunately, we cannot allow duplicate edges/faces in a mesh. Blender's API docs are quite clear about that: https://docs.blender.org/api/blender2.8/bmesh.html#keeping-a-correct-state
Also `mesh.validate(...)` would mark such meshes as invalid.

Can you specify "a weird topology mesh" more precisely? Maybe we can check for this case efficiently before running the algorithm and then either discard such inputs or fix them afterwards.

OmarEmaraDev:
@JacquesLucke  I see. It is weird that the BMesh triangulate method returns duplicates in this case.

The topology Kuldeep is talking about is when two polygons share two edges, some triangulation algorithms will connect those two edges for each of the two polygons resulting in two duplicate triangles as illustrated here. (Left: Original polygons. Center: After triangulating one polygon. Right: After triangulating both polygons.)

JacquesLucke:
True, that seems to be a bug in Blender...
I think for now we could just do the validity check after running the algorithm. If the computed mesh is invalid, an empty mesh is returned instead and an error message is shown.

3DSinghVFX:
> True, that seems to be a bug in Blender...
> I think for now we could just do the validity check after running the algorithm. If the computed mesh is invalid, an empty mesh is returned instead and an error message is shown.

Thanks, @JacquesLucke and @OmarEmaraDev. I'll make the changes.

3DSinghVFX:
> Unfortunately, we cannot allow duplicate edges/faces in a mesh. Blender's API docs are quite clear about that: https://docs.blender.org/api/blender2.8/bmesh.html#keeping-a-correct-state
> Also `mesh.validate(...)` would mark such meshes as invalid.
> 
> Maybe we can check for this case efficiently before running the algorithm and then either discard such inputs or fix them afterwards.

@JacquesLucke and @OmarEmaraDev From the topology point of view, I think this situation may only occur when polygon vertices are not coplanar.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+    cdef Py_ssize_t polyStart, triIndex, polyIndex
+    cdef Py_ssize_t j, k, amount, index, preIndex, earIndex, nexIndex
+
+    triIndex = 0
+    polyIndex = 0
+    for i in range(polygonAmount):
+        polyLength = _oldPolyLengths[i]
+        polyStart = _oldPolyStarts[i]
+        neighbors = LongList(length = polyLength)
+        indices = LongList(length = polyLength)
```
OmarEmaraDev, line 467: Why do we need to store `neighbors` in an array? Aren't the neighbors already known from the original polygon indices?

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+        for j in range(polyLength):
+            neighbors.data[j] = _oldIndices[polyStart + j]
+            indices.data[j] = j
+
+        polyVertices = transformVerticesOfPolygon(oldVertices, neighbors)
```
OmarEmaraDev, line 472: Should we call this something more meaningful? Maybe `projectPolygonVertices`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+
+                subVec3(&ab, &earVertex, &preVertex)
+                subVec3(&bc, &earVertex, &nexVertex)
+
+                normalizeVec3_InPlace(&ab)
+                normalizeVec3_InPlace(&bc)
+
+                angle = angleNormalizedVec3(&ab, &bc)
+                angles.data[k] = int((angle - angle % 0.001) * 180.0 / pi)
+            sortAngles = LongList.fromNumpyArray(numpy.sort(angles.asMemoryView()).astype(int))
```
OmarEmaraDev, line 507: Do we really need to recalculate everything everytime? I assume only two angles will change here. Also, using numpy here would be questionable. If you need to, use `qsort` from stdlib.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+                index = angles.index(sortAngles[k])
+                preIndex = indices.data[(amount + index - 1) % amount]
+                earIndex = indices.data[index]
+                nexIndex = indices.data[(index + 1) % amount]
+
+                v1 = polyVertices.data[preIndex]
+                v2 = polyVertices.data[earIndex]
+                v3 = polyVertices.data[nexIndex]
+                if convexVertex(v1, v2, v3) and not pointsInTriangle(polyVertices, v1, v2, v3, preIndex, earIndex, nexIndex):
+                    indices = LongList.fromNumpyArray(numpy.delete(indices.asMemoryView(), index).astype(int))
```
OmarEmaraDev, line 520: Same as I mentioned above.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+cdef polyCounterClockwise(Vector3DList vertices):
+    cdef Py_ssize_t amount = vertices.length
+    cdef float area = 0
+    cdef Vector3 v1, v2
+    cdef Py_ssize_t i
+
+    for i in range(amount):
+        v1 = vertices.data[i]
+        v2 = vertices.data[(i + 1) % amount]
+        area += (v2.x - v1.x) * (v2.y + v1.y)
```
OmarEmaraDev, line 546: Instead of using modulo here. We can pull the last case out like this:

```python
    for i in range(amount - 1):
        v1 = vertices.data[i]
        v2 = vertices.data[i + 1]
        area += (v2.x - v1.x) * (v2.y + v1.y)
    v1 = vertices.data[amount - 1]
    v2 = vertices.data[0]
    area += (v2.x - v1.x) * (v2.y + v1.y)
```

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+        avgVector.z += rotPolyVertices.data[i].z
+    scaleVec3_Inplace(&avgVector, 1.0 / polyLength)
+
+    # Bring vertices of polygon to the world center
+    cdef Vector3DList offsetPolyVertices = Vector3DList(length = polyLength)
+    for i in range(polyLength):
+        offsetPolyVertices.data[i].x =  rotPolyVertices.data[i].x - avgVector.x
+        offsetPolyVertices.data[i].y =  rotPolyVertices.data[i].y - avgVector.y
+        offsetPolyVertices.data[i].z =  rotPolyVertices.data[i].z - avgVector.z
+
```
OmarEmaraDev, line 654: Is this really necessary? Can't we just skip it?

animation_nodes/nodes/mesh/triangulate_mesh.py
```diff
@@ -0,0 +1,35 @@
...
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
+
+        self.newOutput("Mesh", "Mesh", "meshOut")
+
+    def draw(self, layout):
+        layout.prop(self, "methodType", text = "Use Advanced Method")
+
+    def execute(self, mesh):
+        meshOut = mesh.copy()
```
OmarEmaraDev, line 25: Use `dataIsModified` for the input mesh.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
         usedSlots.data[v2] += 1
 
     return neighboursAmounts, neighboursStarts, neighbours, neighbourEdges
+
+def triangulatePolygonsUsingFanSpanMethod(PolygonIndicesList polygons):
+    cdef unsigned int *_oldPolyLengths = polygons.polyLengths.data
```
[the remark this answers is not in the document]
3DSinghVFX: Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: > Do we really need to recalculate everything everytime? I assume only two angles will change here.

That can be avoided.

> Also, using numpy here would be questionable. If you need to, use `qsort` from stdlib.
 
Okay. I did not know that.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Is there a way to delete an element Or should I use loop for that?

OmarEmaraDev: The Animation Nodes list structures implement element deletion, so you can just do that. Either using the `del` keyword or the `removeValueAtIndex` method.
But if you can avoid deletions, that would be best. Perhaps by keeping track of what was removed without actually removing it.

3DSinghVFX:
> Consider moving all of the functions into a separate file, maybe in `algroithms/mesh`.

In `algroithms/mesh_generation` or create a new folder `algroithms/mesh`?

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Actually, if the input polygon has flip (has inverted normals) then we need to reverse the polygon indices. That is why I store in a `neighbors LongList`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+    cdef Py_ssize_t polygonAmount = polygons.getLength()
+    cdef Py_ssize_t i, triAmount, polyLength
+
+    triAmount = 0
+    for i in range(polygonAmount):
+        polyLength = _oldPolyLengths[i]
+        if polyLength > 3:
+            triAmount += polyLength - 2
+        else:
+            triAmount += 1
```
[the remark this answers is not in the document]
3DSinghVFX: Ahhh, I did not think about that.

animation_nodes/data_structures/meshes/mesh_data.pyx
OmarEmaraDev: Why not reverse the indices themselves? Also, what would happen if we didn't reverse?

OmarEmaraDev:
> In `algroithms/mesh_generation` or create a new folder `algroithms/mesh`?

Create a new directory.

3DSinghVFX:
> > In `algroithms/mesh_generation` or create a new folder `algroithms/mesh`?
> 
> Create a new directory.

Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: I have done testing, there is no longer need of this offset. I'll remove it.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Omar, I tried this but failed. Please help me,
```
from libc.stdlib cimport qsort

newAngles = LongList()
qsort(newAngles, newAngles.length, sizeof(int), compare)

cdef compare(const void * a, const void * b):
   return <int*>a - <int*>b
```

animation_nodes/nodes/mesh/triangulate_mesh.py
3DSinghVFX: That gives this error,
Screenshot from 2020-04-22 17-28-49

animation_nodes/data_structures/meshes/mesh_data.pyx
OmarEmaraDev: `<int*>a` will cast `a` to an integer pointer, so you still need to dereference it to get the actual value `(<int*>a)[0]`.

animation_nodes/nodes/mesh/triangulate_mesh.py
OmarEmaraDev: Can you show me the relevant code using `dataIsModified`?

3DSinghVFX: This is edited _Traingulate Mesh_ node code,
```
import bpy
from bpy.props import *
from ... base_types import AnimationNode
from ... data_structures.meshes.validate import checkMeshData

class TriangulateMeshNode(bpy.types.Node, AnimationNode):
    bl_idname = "an_TriangulateMeshNode"
    bl_label = "Triangulate Mesh"
    errorHandlingType = "EXCEPTION"

    methodType: BoolProperty(name = "Use Advanced Method", default = False,
                             description = "Use advanced method for Mesh Triangulation",
                             update = propertyChanged)

    def create(self):
        self.newInput("Mesh", "Mesh", "mesh", dataIsModified = True)

        self.newOutput("Mesh", "Mesh", "meshOut")

    def draw(self, layout):
        layout.prop(self, "methodType", text = "Use Advanced Method")

    def execute(self, mesh):
        if self.methodType:
            mesh.triangulateMesh(method = "EAR")
        else:
            mesh.triangulateMesh(method = "FAN")

        try:
            checkMeshData(mesh.vertices, mesh.edges, mesh.polygons)
            return mesh
        except Exception as e:
            self.raiseErrorMessage(str(e))
```

3DSinghVFX: This is the function in the `mesh_data.pyx` file
https://github.com/JacquesLucke/animation_nodes/blob/e2195df285ff9182e3b1cadbf7624ed9bfa8da9a/animation_nodes/data_structures/meshes/mesh_data.pyx#L181-L188

OmarEmaraDev: I am not sure. I will have to investigate this myself. So just ignore it for now.

3DSinghVFX: Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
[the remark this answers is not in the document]
3DSinghVFX: Removed.

animation_nodes/data_structures/meshes/mesh_data.pyx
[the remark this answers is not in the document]
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
...
+
+            _newIndices[index] = _oldIndices[start]
+            _newIndices[index + 1] = _oldIndices[start + 1 + j]
+            _newIndices[index + 2] = _oldIndices[start + 2 + j]
+            index += 3
+
+    return newPolygons
+
+@cython.cdivision(True)
+def triangulatePolygonsUsingEarClipMethod(Vector3DList oldVertices, PolygonIndicesList polygons):
```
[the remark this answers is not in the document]
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: The _inner angle_ and _convex_ methods need a counter-clockwise polygon. If the input polygon is clockwise then we have to reverse it to make it counter-clockwise.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: I have modified the algorithm, now no need for deletion.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: > Do we really need to recalculate everything everytime? I assume only two angles will change here. 

I have modified the algorithm, now no need to recalculate everything everytime.

3DSinghVFX: > `<int*>a` will cast `a` to an integer pointer, so you still need to dereference it to get the actual value `(<int*>a)[0]`.

I have done a test, but it is not working.
```
from libc.stdlib cimport qsort

qsort(newAngles, newAngles.length, sizeof(int), compare)

cdef compare(const void * a, const void * b):
   return (<int*>a)[0] - (<int*>b)[0]
```
Screenshot from 2020-04-23 00-41-33

OmarEmaraDev: Looks like you need a pointer to the function. So try `qsort(newAngles, newAngles.length, sizeof(int), &compare)`.

3DSinghVFX: Sorry for late reply. It gave the same error, `Cannot assign type 'object (*)(const void *, const void *)' to 'int (*)(const void *, const void *) nogil'`

OmarEmaraDev: We forgot the function type:
```
cdef int compare(const void * a, const void * b):
```

3DSinghVFX: Still the error is there,
Screenshot from 2020-04-23 02-23-37

3DSinghVFX: Is this related, https://stackoverflow.com/a/50777451

3DSinghVFX: This code is working, 
```
from cpython cimport array
import array

cdef quickSort(LongList newAngles):
    cdef Py_ssize_t amount = newAngles.length
    cdef array.array b = array.array('i', newAngles.asNumpyArray())
    cdef int[:] a = b
    qsort(&a[0], amount, sizeof(int), &compare)
    print(b)

cdef int compare(const void * a, const void * b) with gil:
   return (<int*>a)[0] - (<int*>b)[0]
```

OmarEmaraDev: You need to release the GIL for the compare function. See https://cython.readthedocs.io/en/latest/src/userguide/external_C_code.html#declaring-a-function-as-callable-without-the-gil

3DSinghVFX: Okay, but is it correct way to do that?

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -370,3 +384,272 @@ def calculateLinkedVertices(int verticesAmount, EdgeIndicesList edges):
         usedSlots.data[v2] += 1
 
     return neighboursAmounts, neighboursStarts, neighbours, neighbourEdges
+
+def triangulatePolygonsUsingFanSpanMethod(PolygonIndicesList polygons):
```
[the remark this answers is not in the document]
3DSinghVFX: The functions are moved to a separate file.

animation_nodes/data_structures/meshes/mesh_data.pyx
OmarEmaraDev: It is in the signature of the expected callback, so I guess so.

3DSinghVFX: Okay, 
```
cdef quickSort(LongList newAngles):
    cdef Py_ssize_t amount = newAngles.length
    cdef array.array b = array.array('i', newAngles.asNumpyArray())
    cdef int[:] a = b
    with nogil:
        qsort(&a[0], amount, sizeof(int), &compare)
    print(b)

cdef int compare(const void * a, const void * b) with gil:
   return (<int*>a)[0] - (<int*>b)[0]
```

3DSinghVFX: This is a complete function,
```
cdef quickSort(LongList newAngles):
    cdef Py_ssize_t amount = newAngles.length
    cdef array.array b = array.array('i', newAngles.asNumpyArray())
    cdef int[:] a = b
    with nogil:
        qsort(&a[0], amount, sizeof(int), &compare)

    cdef LongList sortAngles = LongList(length = amount)
    cdef Py_ssize_t i
    for i in range(amount):
        sortAngles.data[i] = b[i]
    return sortAngles

cdef int compare(const void * a, const void * b) with gil:
   return (<int*>a)[0] - (<int*>b)[0]
```
Speed testing:
**With qsort:**
test1_gil

**With Numpy:**
test1_nogil

3DSinghVFX: Can we use `sort` instead of `qsort` which is fast according this site: https://www.geeksforgeeks.org/c-qsort-vs-c-sort/?ref=rp

OmarEmaraDev: I meant like this.

```cython
from libc.stdlib cimport qsort

cdef int compare(const void * a, const void * b) nogil:
   return (<int*>a)[0] - (<int*>b)[0]

qsort(newAngles, newAngles.length, sizeof(int), &compare)

```

3DSinghVFX: It gave an error, maybe need to `newAngles` to C-object,
Screenshot from 2020-04-23 17-40-47

OmarEmaraDev: `newAngles` is a `LongList`. Use `newAngles.data`.

3DSinghVFX: Thank you so much for the help.

It is working But the output is not the same,
Screenshot from 2020-04-23 17-50-38

OmarEmaraDev: A LongList is composed of longs, so I think `sizeof(int)` should be `sizeof(long)`.

3DSinghVFX: Ahhh, it worked. 

Thank you so much :)

OmarEmaraDev: How is the performance now? Is it better than numpy? I am not sure if we are allowed to use c++ here, so not sure about the STL sort function.

3DSinghVFX: Yes, it is better.

But we use c++ in the Fast Noise.

3DSinghVFX: Based on your suggestions, now performance is far better. For example triangulation of a sphere, timing reduces to `~17ms` to `1.45ms`.

3DSinghVFX: Thanks. Okay.

3DSinghVFX:
Hi @OmarEmaraDev 

> I just skimmed over the paper you used to implement the triangulation. Its seems their improved algorithm aims to reduced thin triangles by choosing the ear with the minimum angle and swapping edges when needed. It seems you implemented the former but not the latter, or did you?
 
I used minimum angle approach.

> Implementation wise, here are some points:
> 
>     * You copy and resort the whole angles array for every iteration which is bad, I think I mentioned this before, and while you only compute the two changed angles, we also need to avoid resorting the whole array. To solve this an incremental data structure could be used, like a Heap.

Okay. It computes only two angles i.e., angles of vertices adjacent to the ear-vertex. I do not know Heap method, but I'll try to implement it. 

>     * Shouldn't we skip reflex vertices in the `earNextNeighborIndices` function?

As mentioned in that paper, the removal of an Ear-Vertex can change the status of neighbour vertices (reflex or convex). To compute the angles of neighbour vertices (of Ear-Vertex), we need next-neighbour vertices whether they are reflex or convex.

>     * I don't see why conversion to degrees is necessary here, won't radians do just fine?

I added `(angle - angle % 0.001) * 180.0 / pi` to avoid the precision problem. If the double list with `(angle - angle % 0.001)` can handle that problem, we can remove degree conversion. 

>     * Can we compute directed angles and use them to identify reflex and convex vertices?

Sorry, what do you mean by _directed angles_? I had tried with polygon normal that did not work.



> 
> So generally, the structure should be as follows:
> 
>     1. Compute the **directed** interior angles and the ear status of the vertices. The sign of the angle indicates the reflex and convex status of the vertices.
> 
>     2. Construct a heap from the vertices.
> 
>     3. Extract the vertex with the minimum angle, do the triangulation.

During extraction, we have to check that the triangle formed by the Ear-Vertex with minimum angle does not include any other vertex. If it does then we have to go for next Ear-Vertex with minimum angle. 
 
>     4. Update the heap based on the two changed angles.
> 
>     5. Go back step 3 until done.

The overall structure of the current algorithm is similar to the above. We have to add the `Vertex Status` and `Heap` method.

Current algorithm:
0. Check a polygon (after projection) is counter-clockwise.
1. Compute the interior angles of each vertex.
2. Sort the angles.
3. Remove the Ear (convex, min interior angle) by the condition that formed triangle does not include any other vertex.
4. Update angles of neighbours of the Ear-Vertex.
5. Repeat step 2.

3DSinghVFX:
> > I used minimum angle approach.
> 
> So you are not doing edge swapping?

Yes. I'm not doing edge swapping.
> 
> > To compute the angles of neighbour vertices (of Ear-Vertex), we need next-neighbour vertices whether they are reflex or convex.
> 
> I am aware, but you are looping over all vertices, while you should only loop over convex vertices. Right?

Yes. If we store the Ear Vertex-Status (as 1) then we can just `continue` the loop for `reflex` vertices:
```
for i in range(amount):
    if vertexStatus[i] == 0: continue
    check for possible trianguation...
    update the angles and vertex-status
```
> > What do you mean by directed angles?
> 
> A directed angle is a signed angle. See the following diagram:
> 
> https://animation-nodes-manual.readthedocs.io/en/latest/user_guide/nodes/vector/vector_angle.html#rotation-difference

Ahh, that will work, we can compute angle and vertex-status simultaneously. Thanks for the info.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+        verticesData = <Vertex*>PyMem_Malloc(polyLength * sizeof(Vertex))
+        projectPolygonVertices(vertices, polyStart, polyLength, oldIndices, verticesData)
+
+        # Check polygon polarity.
+        if not polyCounterClockwise(verticesData):
+            for j in range(polyLength):
+                previousIndex = verticesData[j].previous
+                nextIndex = verticesData[j].next
+                verticesData[j].previous = nextIndex
+                verticesData[j].next = previousIndex
```
OmarEmaraDev, line 93: Put those in a function, just to make it clear what is happening.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+                triangleIndex += 1
+
+                index = 0
+                for k in range(polyLength):
+                    if verticesData[k].isEar:
+                        newIndices[polyIndex + index] = verticesData[k].index
+                        index += 1
+                        if index == 3: break
+                polyIndex += 3
+                break
```
OmarEmaraDev, line 118: It appears you are checking if this is the last triangle and skipping the extra checks if it is. What I would do is move this check after the loop. So you loop one time less `for j in range(polyLength - 3):` and then add the last triangle manually.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+    return newPolygons
+
+# Remove ear-vertex and update neighbor-vertices.
+cdef void removeEarVertex(Py_ssize_t polyLength, Vertex* verticesData, Py_ssize_t earIndex):
+    cdef Py_ssize_t previousIndex, nextIndex
+    previousIndex = verticesData[earIndex].previous
+    nextIndex = verticesData[earIndex].next
+
+    verticesData[earIndex].isEar = False
```
OmarEmaraDev, line 144: The ear is being removed, so you don't need to update any of its values. More on this later.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+        verticesData[previousIndex].isEar = True
+    else:
+        verticesData[previousIndex].isEar = False
+
+    angle = calculateAngleWithSign(verticesData, nextIndex)
+    verticesData[nextIndex].angle = angle
+    if angle >= 0.0 and not pointsInTriangle(polyLength, verticesData, nextIndex):
+        verticesData[nextIndex].isEar = True
+    else:
+        verticesData[nextIndex].isEar = False
```
OmarEmaraDev, line 161: It seems we can move this to a function since it is repeated 3 times, 2 here and one above.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+cdef struct Vertex:
+    Py_ssize_t index
+    Vector3 location
+    float angle
+    bint isEar
+    Py_ssize_t previous
+    Py_ssize_t next
```
OmarEmaraDev, line 169: Move this to the top of the file.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+# Make sure normal (+z) is correct.
+@cython.cdivision(True)
+cdef bint polyCounterClockwise(Vertex* verticesData):
+    cdef Vector3 v1, v2
+    cdef float area = 0.0
+    cdef Py_ssize_t headIndex = 0
```
OmarEmaraDev, line 176: Head index is not zero. More on this later.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+    normalizeVec3_InPlace(&ab)
+    normalizeVec3_InPlace(&bc)
+
+    angle = angleNormalizedVec3(&ab, &bc)
+
+    # Make sure normal (+z) is correct, and the polygon is counter clockwise. Convex vertex has sign >= 0.
+    cdef float sign = (v2.x - v1.x) * (v3.y - v1.y) - (v2.y - v1.y) * (v3.x - v1.x)
+    if sign < 0.0: return -(angle - angle % 0.001)
+    return (angle - angle % 0.001)
```
OmarEmaraDev, line 209: Why the need for this `angle - angle % 0.001`?

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+    cdef float angle
+    cdef Py_ssize_t i, angleMinIndex
+
+    angleMinIndex = 0
+    for i in range(polyLength):
+        if verticesData[i].isEar:
+            angle = abs(verticesData[i].angle)
+            if angleMin > angle:
+                angleMin = angle
+                angleMinIndex = i
```
OmarEmaraDev, line 223: You are not looping over the linked list correctly. More on this later.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+    cdef Vector3 v1 = verticesData[previousIndex].location
+    cdef Vector3 v2 = verticesData[earIndex].location
+    cdef Vector3 v3 = verticesData[nextIndex].location
+
+    cdef Py_ssize_t i
+    for i in range(polyLength):
+        if i != previousIndex and i != earIndex and i != nextIndex and pointInsideTriangle(v1, v2, v3, verticesData[i].location):
+            return True
+    return False
```
OmarEmaraDev, line 240: Not looping correctly here either. More on this later.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+    Py_ssize_t index
+    Vector3 location
+    float angle
+    bint isEar
+    Py_ssize_t previous
+    Py_ssize_t next
+
+# Make sure normal (+z) is correct.
+@cython.cdivision(True)
+cdef bint polyCounterClockwise(Vertex* verticesData):
```
OmarEmaraDev, line 173: For functions that return bint, name it as `is*`, in this case, `isPolyCounterClockwise`.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+        v2 = vertices.data[oldIndices[polyStart + (i + 1) % polyLength]]
+
+        polyNormal.x += (v1.y - v2.y) * (v1.z + v2.z)
+        polyNormal.y += (v1.z - v2.z) * (v1.x + v2.x)
+        polyNormal.z += (v1.x - v2.x) * (v1.y + v2.y)
+
+        verticesData[i].index = oldIndices[polyStart + i]
+        verticesData[i].location = v1
+        verticesData[i].previous = (polyLength + i - 1) % polyLength
+        verticesData[i].next = (i + 1) % polyLength
```
OmarEmaraDev, line 288: I prefer to handle the first and last vertex independently, to avoid the remainder operator here.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,316 @@
...
+
+    polyIndex = 0
+    triangleIndex = 0
+    for i in range(polygonAmount):
+        polyStart = oldPolyStarts[i]
+        polyLength = oldPolyLengths[i]
+
+        # Initialization of polygon's vertices.
+        verticesData = <Vertex*>PyMem_Malloc(polyLength * sizeof(Vertex))
+        projectPolygonVertices(vertices, polyStart, polyLength, oldIndices, verticesData)
```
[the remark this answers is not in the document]
3DSinghVFX: It is done.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is added into `makeSurePolygonIsCounterClockwise(verticesData)` function.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is done.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: `verticesData[earIndex].isEar = False` is removed.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is added into `setVertexEarStatus` function

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is moved.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: `struct Vertices` is added to track the `head`.
```
cdef struct Vertices:
    Py_ssize_t head
    Vertex *data
```

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: I added this to handle the precision problem. It mainly happens when angles are equal to each other e.g., square polygon.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is changed.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is changed.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: Okay. This function is changed to `makeSurePolygonIsCounterClockwise`.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: It is done.

3DSinghVFX:
> The code is much cleaner now, I think. But you are still having trouble with the linked list, so let me clarify.
> 
> In our case, the doubly liked list is circular, meaning the first element references the last. So the list is fully defined by two values, the array of nodes, which you already have, and a head index, which you don't seem to keep track of. We can put both values in a struct as follows and pass it as a single value to functions:
> 
> ```cython
> cdef struct Vertices:
>     Py_ssize_t head
>     Vertex *data
> ```
> 
> The head index is the index of any of the nodes of the list. When you remove a node, you have to check if it is the head, and if it is, you have to change the head to any other node in the list. In other words, you have to keep track of your head and make sure it is correctly set.
> 
> Secondly, when looping over the linked list, you have two cases: (1) The list is full (No elements were removed) and (2) some elements were removed. If the case is (1), then you may loop over the list just like you would loop over any other list, however, once an element was removed (2), you have to loop over it as a linked list, just like I described before.
> 
> ```python
> currentVertex = headVertex
> while True:
>     # Do something with the current vertex.
>     currentVertex = currentVertex.next
>     if currentVertex == headVertex:
>         break
> ```
> 
> In your case, initialization, projection, and `polyCounterClockwise` all deal with full lists, so you can loop over them normally. In any other case, you have to loop as a linked list. Moreover, a node is removed simply by switching the next and previous indices of its neighbours, don't do anything else, because I see you are setting the ear status, which is unnecessary.

@OmarEmaraDev Thanks for information and clarification.

I have implemented `Vertex` and `Vertices` structures as you suggested above. Based on that I have modified and refined the code.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
OmarEmaraDev: Can you elaborate? What happens if the angles are equal?

3DSinghVFX: I have tried without that, now everything is fine.

However, before the angles were changing after some decimal places. May be due to the `pi` that I used before.

3DSinghVFX: It is happening when the input mesh is animated.

3DSinghVFX: Here is an example:
Peek 2020-06-18 19-46

OmarEmaraDev: I see. Do we need `@cython.cdivision(True)` here since `%` is used?

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,353 @@
...
+            newPolyStarts[triangleIndex] = polyIndex
+            newPolyLengths[triangleIndex] = 3
+            triangleIndex += 1
+
+            newIndices[polyIndex] = vertexData[vertexData[earIndex].previous].index
+            newIndices[polyIndex + 1] = vertexData[earIndex].index
+            newIndices[polyIndex + 2] = vertexData[vertexData[earIndex].next].index
+            polyIndex += 3
+
+            if earIndex == verticesData.head: verticesData.head = vertexData[earIndex].next
```
OmarEmaraDev, line 142: Put this line inside the `removeEarVertex` function.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
```diff
@@ -0,0 +1,353 @@
...
+
+    return newPolygons
+
+# Find the index of minimum angle.
+cdef int findEarHasMinAngle(Vertices verticesData):
+    cdef Py_ssize_t headIndex = verticesData.head
+    cdef Vertex* vertexData = verticesData.data
+    cdef Py_ssize_t currentIndex
+
+    cdef float angleMin = 3.14
```
OmarEmaraDev, line 169: `angleMin` should be the angle of the head vertex.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: I think no. `cython -a` shows white color for those lines.
Screenshot from 2020-06-18 19-58-27

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: I have tried, it does not work inside the `removeEarVertex` function.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: Okay.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
OmarEmaraDev: Can you show me what you tried?

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: Omar, we can't set it as the angle of the head-vertex because head-vertex may not be an ear but can have small angle.

OmarEmaraDev: I see. Then you should increase the value to 3.15, and add a comment explaining this.

3DSinghVFX: Okay.

3DSinghVFX: Done.

animation_nodes/algorithms/mesh/triangulate_mesh.pyx
3DSinghVFX: > Put this line inside the `removeEarVertex` function.

Done.

Issue #1414, opened 2020-05-05
Possible to keep UV Map when transforming the meshes list


kdsin23:
![uv error](https://user-images.githubusercontent.com/64368228/81029515-d7dd8600-8eb7-11ea-8eb7-3021f11099ad.png)

when transform the mesh list ,why the "UV Map" is None,is it possible to generate the uv?i think if possible,it is good for us to texture.

OmarEmaraDev:
It seems the append method of the mesh structure doesn't support UVs. Thanks for the report!

OmarEmaraDev:
Fixed in 1a73bec.

PR #1415, opened 2020-05-05
Added Load UVs option in the Transform Mesh node.

3DSinghVFX:
I have added the option _Load UVs_ in the _Transform Mesh_ node. This option loads the UVs of input mesh-list. Fix the issue https://github.com/JacquesLucke/animation_nodes/issues/1414

Example:
mesh_transform1
mesh_transform2


OmarEmaraDev:
This shouldn't be an option. It should always be done. Also, can we generalize this to work for other mesh properties? Something like the `transferMeshProperties` method of meshes. Also, this only works for a single UV Map, it should set all of them.

3DSinghVFX:
@OmarEmaraDev Hi
> This shouldn't be an option. It should always be done.

If we _always done_ this then it will cost some performance. Or we can add an option (ON by default) in the advanced setting that loads all the mesh-properties (UVs, Vertex Colors, etc).

>Also, can we generalize this to work for other mesh properties? Something like the `transferMeshProperties` method of meshes.

Now, it is for _UVs_ and _Vertex Colors_.

>Also, this only works for a single UV Map, it should set all of them.

It is done, can add all _UV Maps_ and _Vertex Colors_.

If we add similar functionality for _Material Indices_ (Insert Material Indices and Get Material Indices) in the mesh-data then we can add the same feature for _Material Indices_ as well.

OmarEmaraDev:
@3DSinghVFX The performance penalty will not be significant as it will just be some memory copying. And I think it is essential functionality, so the option shouldn't be there.

3DSinghVFX:
> @3DSinghVFX The performance penalty will not be significant as it will just be some memory copying. And I think it is essential functionality, so the option shouldn't be there.

Thanks for clarification. I'll commit the changes.

3DSinghVFX:
> If we add similar functionality for _Material Indices_ (Insert Material Indices and Get Material Indices) in the mesh-data then we can add the same feature for _Material Indices_ as well.

@OmarEmaraDev What about the _Material Indices_?

OmarEmaraDev:
@3DSinghVFX We don't store materials indices in meshes, do we? If you mean adding support for material indices, then don't do that now.

3DSinghVFX:
> @3DSinghVFX We don't store materials indices in meshes, do we? If you mean adding support for material indices, then don't do that now.

We don't store material indices. Okay.

OmarEmaraDev:
I apologize @3DSinghVFX, but I opted for a different solution with a bit of refactoring in 1a73bec.

3DSinghVFX:
> I apologize @3DSinghVFX, but I opted for a different solution with a bit of refactoring in [1a73bec](https://github.com/JacquesLucke/animation_nodes/commit/1a73bec52c5560ea2a4aae73b7f739cb18650ce3).

It is fine.

PR #1449, opened 2020-06-11
XoShiRo256 Random Number Generators.

OmarEmaraDev:
This patch adds a number of random number generators, namely the XoShiRo256**, XoShiRo256+, and the SplitMix64 generators.

## SplitMix64

This is used internally by the XoShiRo256** and the XoShiRo256+ generators to initialize their states. It should not be used directly.

## XoShiRo256**

This generator can be used to get random integer values. Three classes of methods exist:

- *Standard* : Returns a random integer.
- *With Max* : Returns a random integer between `0` and some number `max`, excluding `max`.
- *In Range* : Returns a random integer between `start` and `end`, excluding `end`.

## XoShiRo256+

This generator can be used to get random floating point numbers. Tow classes of methods exist:

- *Standard* : Returns a random floating point number between 0 and 1.
- *In Range* : Returns a random floating point number between `start` and `end`.

---

Finally, a utility method `cantorPair` was added to be used to compute a "final seed" from two seed values. This is especially useful for nodes with a node seed and an input seed.


OmarEmaraDev:
@JacquesLucke Based on the code of the [Java DSI utilities](http://dsiutils.di.unimi.it/docs/it/unimi/dsi/util/XoShiRo256StarStarRandomGenerator.html), also by the same author. The relevant Java code is as follows:
```java
public long nextLong(final long n) {
	if (n <= 0) throw new IllegalArgumentException("illegal bound " + n + " (must be positive)");
	long t = nextLong();
	final long nMinus1 = n - 1;
	// Rejection-based algorithm to get uniform integers in the general case
	for (long u = t >>> 1; u + nMinus1 - (t = u % n) < 0; u = nextLong() >>> 1);
	return t;
}
```

JacquesLucke:
thanks

PR #1492, opened 2020-08-06
Add menu fix

3DSinghVFX:
I have fixed the issue of adding *Sort Numbers* node from the *Add* menu, and update the *CHANGELOG.md* for rename of the *Sort* node as *Sort Numbers* node.


OmarEmaraDev:
Thanks!

PR #1501, opened 2020-08-14
Add Normals output in Mesh Points Scatter node.

3DSinghVFX:
I have added *Normals* output to *Mesh Points Scatter* node.


3DSinghVFX:
> Can you extract it from the matrices if required just like you do with the output vectors?

Okay. Do we have a direct method for that?

OmarEmaraDev:
Not as far as I know. Create your own function. To get the `normal` from the matrix `m` you do this:
```
normal.x, normal.y, normal.z = m.a13, m.a23, m.a33
```

OmarEmaraDev:
Rebase this, then we can merge.

PR #1504, opened 2020-08-14
 Add Axis Scale option in Viewer 3D node

3DSinghVFX:
I have added `XYZ` options for matrices in *Viewer 3D* node. These options allow to choose the axis for drawing.
Screenshot from 2020-08-14 21-00-15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Added edges mode in *Mesh Points Scatter* node.
 - Added *Vertex Color Fill* attribute to GP strokes.
 - Added *Radial Falloff* node.
+- Added *Axis Scale* option in *Viewer 3D* node.
 
 ### Fixed
 
--- a/animation_nodes/graphics/c_utils.pyx
+++ b/animation_nodes/graphics/c_utils.pyx
@@ -1,7 +1,7 @@
 from .. math cimport Vector3
 from .. data_structures cimport Vector3DList, Matrix4x4List, IntegerList
 
-def getMatricesVBOandIBO(Matrix4x4List matrices, float scale):
+def getMatricesVBOandIBO(Matrix4x4List matrices, float scaleX, float scaleY, float scaleZ):
     cdef Py_ssize_t i
     cdef int length = len(matrices)
     cdef Vector3DList vectors = Vector3DList(length = length * 4)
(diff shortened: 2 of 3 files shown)


OmarEmaraDev:
I am not sure if that's a good idea. I would rather leave that to the user if the user needed it.

3DSinghVFX:
> I am not sure if that's a good idea. I would rather leave that to the user if the user needed it.

Omar, by default all the axis are enabled.

Gerstmann-Bradley:
> > I am not sure if that's a good idea. I would rather leave that to the user if the user needed it.
> 
> Omar, by default all the axis are enabled.

Do you mean that, in this update, all axis are enabled by default as in the past?

OmarEmaraDev:
I mean generally. If the user wants to zero some axis, then the user should do that explicitly, we shouldn't do it in the node.

If we want to make it easier to visualizer normals somehow, then we should add an independent drawing method for that. But I don't consider this a good solution.

3DSinghVFX:
> > > I am not sure if that's a good idea. I would rather leave that to the user if the user needed it.
> > 
> > 
> > Omar, by default all the axis are enabled.
> 
> Do you mean that, in this update, all axis are enabled by default as in the past?

Yes. But you can disable them.
Screenshot from 2020-08-14 21-35-34

Gerstmann-Bradley:
I can see some usage of this in some extreme conditions. Personally speaking having these options makes me stressful, because this is unique for matrices, and such uniqueness seems to highlight the importance of these options, but they aren't that important and will not be frequently used in my opinion. I may fear to click them accidentally somehow.
Is there anyway to hide these options if possible? so that if people want to hide it, they can hide it and otherwise everything will kept as before?

3DSinghVFX:
> I mean generally. If the user wants to zero some axis, then the user should do that explicitly, we shouldn't do it in the node.
> 
> If we want to make it easier to visualizer normals somehow, then we should add an independent drawing method for that. But I don't consider this a good solution.

> I can see some usage of this in some extreme conditions. Personally speaking having these options makes me stressful, because this is unique for matrices, and such uniqueness seems to highlight the importance of these options, but they aren't that important and will not be frequently used in my opinion. I may fear to click them accidentally somehow.

I also realise that problem.

>Is there anyway to hide these options if possible? so that if people want to hide it, they can hide it and otherwise everything will kept as before?

I think that is not possible.

Okay. I'll close this.

OmarEmaraDev:
We can always move such options to the advanced node settings panel. But even implementation wise, in the common case, you are creating a VBO of size double that of the size of the data that you will actually use for drawing, and that's beside the drawing itself. So the solution seems a bit hacky.

3DSinghVFX:
> We can always move such options to the advanced node settings panel. 

Okay. I have moved those options to the advanced settings.

> But even implementation wise, in the common case, you are creating a VBO of size double that of the size of the data that you will actually use for drawing, and that's beside the drawing itself. So the solution seems a bit hacky.

Okay.

PR #1516, opened 2020-08-24
Vectorize Separate quaternion node

harisreedhar:
Vectorized separate quaternion node


OmarEmaraDev (commented):
Update the CHANGELOG file.

OmarEmaraDev (approved):
Thanks!

PR #1517, opened 2020-08-24
Vectorize Combine Quaternion node

harisreedhar:
Vectorized Combine Quaternion node


OmarEmaraDev (changes requested):
Looks good. But you need to normalize the quaternions.

harisreedhar:
> Looks good. But you need to normalize the quaternions.

Should i write Quaternion Normalize function on [/math/quaternion.pyx](https://github.com/JacquesLucke/animation_nodes/blob/master/animation_nodes/math/quaternion.pyx)?

OmarEmaraDev:
Yes.

harisreedhar (commented):
Normalized quaternion output

animation_nodes/math/quaternion.pyx
```diff
@@ -47,3 +47,14 @@ cdef void quatFromAxisAngle(Quaternion *q, Vector3 *axis, float angle):
...
+@cython.cdivision(True)
+cdef void quaternionNormalize_InPlace(Quaternion *q):
+    cdef float length = sqrt(q.w * q.w + q.x * q.x + q.y * q.y + q.z * q.z)
+    if length != 0:
+        q.w /= length
+        q.x /= length
+        q.y /= length
+        q.z /= length
+    else:
+        q.w = q.x = q.y = q.z = 0
```
OmarEmaraDev, line 60: When the length is zero, those values are already zero, so this line is redundant.

OmarEmaraDev (changes requested):

animation_nodes/math/quaternion.pyx
harisreedhar: removed else statement

OmarEmaraDev (approved):
Thanks!

PR #1521, opened 2020-08-29
Add Mesh Falloff node.

3DSinghVFX:
I have added the *Mesh Falloff* node. This node use mesh surface or volume to compute falloff.
Screenshot from 2020-08-29 11-12-36

Example:
Peek 2020-08-29 11-18


animation_nodes/nodes/falloff/mesh_falloff.py
```diff
@@ -0,0 +1,64 @@
+import bpy
+from bpy.props import *
+from mathutils.bvhtree import BVHTree
+from ... events import propertyChanged
+from ... base_types import AnimationNode
+from . mesh_falloff_utils import calculateMeshSurfaceFalloff, calculateMeshVolumeFalloff
+
+modeItems = [
+    ("SURFACE", "Surface", "Falloff w.r.t mesh surface", "NONE", 0),
+    ("VOLUME", "Volume", "Falloff w.r.t mesh volume", "NONE", 1)
```
OmarEmaraDev, line 10: What does w.r.t mean here?

animation_nodes/nodes/falloff/mesh_falloff.py
```diff
@@ -0,0 +1,64 @@
...
+class MeshFalloffNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_MeshFalloffNode"
+    bl_label = "Mesh Falloff"
+    errorHandlingType = "EXCEPTION"
+
+    mode: EnumProperty(name = "Mode", default = "SURFACE",
+        items = modeItems, update = AnimationNode.refresh)
+
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
```
OmarEmaraDev, line 22: Perhaps we should replace this by a BVH Tree instead.

animation_nodes/nodes/falloff/mesh_falloff.py
```diff
@@ -0,0 +1,64 @@
...
+        self.newOutput("Falloff", "Falloff", "falloff")
+
+    def draw(self, layout):
+        layout.prop(self, "mode", text = "")
+
+    def getExecutionFunctionName(self):
+        if self.mode == "SURFACE":
+            return "execute_MeshSurfaceFalloff"
+        elif self.mode == "VOLUME":
+            return "execute_MeshVolumeFalloff"
```
OmarEmaraDev, line 43: As far as I understand Volume is a binary version of Surface with zero size and unit falloff width. So it seem to me we can join both methods into a single one.

animation_nodes/nodes/falloff/mesh_falloff.py
```diff
@@ -0,0 +1,64 @@
...
+    def validMesh(self, mesh):
+        vectorList = mesh.vertices
+        if len(vectorList) == 0:
+            self.raiseErrorMessage("Invalid Mesh.")
+
+        polygonsIndices = mesh.polygons
+        if len(polygonsIndices.indices) == 0:
+            self.raiseErrorMessage("Invalid Mesh.")
+
+        return vectorList, polygonsIndices
```
OmarEmaraDev, line 64: I wouldn't call this an invalid mesh. Instead, return a constant falloff.

animation_nodes/nodes/falloff/mesh_falloff_utils.pyx
```diff
@@ -0,0 +1,130 @@
...
+    cdef Py_ssize_t hits = 0
+
+    offset = direction * 0.0001
+    location = bvhTree.ray_cast(start, direction)[0]
+
+    while location is not None:
+        hits += 1
+        location = bvhTree.ray_cast(location + offset, direction)[0]
+
+    return hits
```
OmarEmaraDev, line 130: Move those into somewhere common between the BVH nodes and this node, perhaps in `/utils`. The same goes for other shared code.

animation_nodes/nodes/falloff/mesh_falloff_utils.pyx
```diff
@@ -0,0 +1,130 @@
...
+        self.dataType = "LOCATION"
+        self.clamped = True
+
+    cdef float evaluate(self, void *value, Py_ssize_t index):
+        return calculateStrength(self, <Vector3*>value)
+
+    cdef void evaluateList(self, void *values, Py_ssize_t startIndex, Py_ssize_t amount, float *target):
+        cdef Py_ssize_t i
+        for i in range(amount):
+            target[i] = calculateStrength(self, <Vector3*>values + i)
```
OmarEmaraDev, line 97: The evaluateList method doesn't seem to be needed in this case.

OmarEmaraDev (changes requested):

animation_nodes/nodes/falloff/mesh_falloff.py
3DSinghVFX: w.r.t is 'with respect to'.

animation_nodes/nodes/falloff/mesh_falloff.py
3DSinghVFX: Okay.

animation_nodes/nodes/falloff/mesh_falloff_utils.pyx
3DSinghVFX: Okay.

animation_nodes/nodes/falloff/mesh_falloff.py
3DSinghVFX: Okay.

animation_nodes/nodes/falloff/mesh_falloff_utils.pyx
3DSinghVFX: Omar, Can you please elaborate why we do not need the `evaluateList` method here?

animation_nodes/nodes/falloff/mesh_falloff.py
3DSinghVFX: Replaced.

animation_nodes/nodes/falloff/mesh_falloff.py
3DSinghVFX: Done.

animation_nodes/utils/bvh.pyx
```diff
@@ -0,0 +1,29 @@
+from random import random
```
OmarEmaraDev, line 1: This should be a normal python file.

animation_nodes/utils/bvh.pyx
```diff
@@ -0,0 +1,29 @@
...
+from mathutils import Vector
+
+cdef float hitsPerLocation(bvhTree, vector):
```
OmarEmaraDev, line 4: Doesn't this return a boolean that says if the input is inside the volume? The function should be renamed accordingly.

animation_nodes/utils/bvh.pyx
```diff
@@ -0,0 +1,29 @@
...
+    direction1 = Vector((random(), random(), random())).normalized()
```
OmarEmaraDev, line 5: Move those variables into global variables like in `is_inside_volume.py`. Also, call the function from the util file in the `is_inside_volume.py` file. The main reason we did that was to avoid duplicate code.

animation_nodes/nodes/falloff/mesh_falloff.pyx
```diff
@@ -0,0 +1,86 @@
...
+from mathutils import Vector
+from ... math cimport Vector3
+from ... base_types import AnimationNode
+from ... data_structures cimport BaseFalloff
+from . constant_falloff import ConstantFalloff
+
+class MeshFalloffNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_MeshFalloffNode"
+    bl_label = "Mesh Falloff"
+    errorHandlingType = "EXCEPTION"
```
OmarEmaraDev, line 13: This is not needed.

animation_nodes/nodes/falloff/mesh_falloff.pyx
```diff
@@ -0,0 +1,86 @@
...
+
+    def create(self):
+        self.newInput("BVHTree", "BVHTree", "bvhTree")
+        self.newInput("Float", "Size", "size")
+        self.newInput("Float", "Falloff Width", "falloffWidth", value = 1)
+        self.newInput("Boolean", "Use Surface", "useSurface", value = True)
+        self.newInput("Boolean", "Use Volume", "useVolume", value = False)
+        self.newInput("Boolean", "Invert", "invert", value = False)
```
OmarEmaraDev, line 21: No need for an invert option here. The user should invert manually if needed.

animation_nodes/nodes/falloff/mesh_falloff.pyx
```diff
@@ -0,0 +1,86 @@
...
+        self.newInput("Float", "Max Distance", "bvhMaxDistance", minValue = 0, value = 1e6, hide = True)
+
+        self.newOutput("Falloff", "Falloff", "falloff")
+
+    def execute(self, bvhTree, size, falloffWidth, useSurface, useVolume, invert, bvhMaxDistance):
+        if bvhTree is None: return ConstantFalloff(0)
+        if not useSurface and not useVolume: return ConstantFalloff(0)
+        return calculateMeshFalloff(bvhTree, bvhMaxDistance, size, falloffWidth, useSurface, useVolume, invert)
+
+cdef class calculateMeshFalloff(BaseFalloff):
```
OmarEmaraDev, line 31: This should be named `MeshFalloff`.

animation_nodes/nodes/falloff/mesh_falloff.pyx
```diff
@@ -0,0 +1,86 @@
...
+    errorHandlingType = "EXCEPTION"
+
+    def create(self):
+        self.newInput("BVHTree", "BVHTree", "bvhTree")
+        self.newInput("Float", "Size", "size")
+        self.newInput("Float", "Falloff Width", "falloffWidth", value = 1)
+        self.newInput("Boolean", "Use Surface", "useSurface", value = True)
+        self.newInput("Boolean", "Use Volume", "useVolume", value = False)
+        self.newInput("Boolean", "Invert", "invert", value = False)
+        self.newInput("Float", "Max Distance", "bvhMaxDistance", minValue = 0, value = 1e6, hide = True)
```
OmarEmaraDev, line 22: Maybe this should be removed and implicitly set to a sufficiently large number.

OmarEmaraDev (changes requested):

animation_nodes/nodes/falloff/mesh_falloff_utils.pyx
OmarEmaraDev: The base falloff implement the `evaluateList` method:

```
    cdef void evaluateList(self, void *objects, Py_ssize_t startIndex,
                           Py_ssize_t amount, float *target):
        cdef Py_ssize_t i
        cdef Py_ssize_t elementSize = getSizeOfFalloffDataType(self.dataType)
        for i in range(amount):
            target[i] = self.evaluate(<char*>objects + i * elementSize, i + startIndex)
```

In this case, your overriding method doesn't seem to provide any gain, so we can just leave it as is.

OmarEmaraDev: @3DSinghVFX It should be noted that your new version of the code could use an `evaluateList`  method to avoid all of the unnecessary branching and checks.

animation_nodes/nodes/falloff/mesh_falloff.pyx
3DSinghVFX: Is `1e10` fine?

OmarEmaraDev: Yes.

animation_nodes/nodes/bvh_tree/is_inside_volume.py
```diff
@@ -30,24 +23,4 @@ def getExecutionCode(self, required):
...
-        hits = 0
-        offset = direction * 0.0001
-        location = bvhTree.ray_cast(start, direction)[0]
-
-        while location is not None:
-            hits += 1
-            location = bvhTree.ray_cast(location + offset, direction)[0]
-
-        return hits
+        return isInsideVolume(bvhTree, vector)
```
OmarEmaraDev, line 26: Call the function inside the execution code directly.

animation_nodes/nodes/falloff/mesh_falloff.pyx
```diff
@@ -0,0 +1,75 @@
...
+        self.newInput("Float", "Falloff Width", "falloffWidth", value = 1)
+        self.newInput("Boolean", "Use Surface", "useSurface", value = True)
+        self.newInput("Boolean", "Use Volume", "useVolume", value = False)
+
+        self.newOutput("Falloff", "Falloff", "falloff")
+
+    def execute(self, bvhTree, size, falloffWidth, useSurface, useVolume):
+        if bvhTree is None: return ConstantFalloff(0)
+        if not useSurface and not useVolume: return ConstantFalloff(0)
+        return MeshFalloff(bvhTree, 1e10, size, falloffWidth, useSurface, useVolume)
```
OmarEmaraDev, line 26: What I meant is that you should pass this value directly to `self.bvhTree.find_nearest`, it shouldn't be part of the falloff class.

OmarEmaraDev (changes requested):

animation_nodes/nodes/falloff/mesh_falloff.pyx
3DSinghVFX: Okay.

animation_nodes/nodes/bvh_tree/is_inside_volume.py
3DSinghVFX: If I write like this 
```
        return "isInside = isInsideVolume(bvhTree, vector)"
```
then it gives an error `NameError: name 'isInsideVolume' is not defined`.

Should I write normal `execute(self, )` function?

3DSinghVFX:
Omar, another thing is I have changed offset
https://github.com/JacquesLucke/animation_nodes/blob/616a3d5a9d8c0733023b39598941c08b6df29a1c/animation_nodes/utils/bvh.py#L24

It was `0.0001` and gave some fluctuations,
untitled

animation_nodes/nodes/bvh_tree/is_inside_volume.py
OmarEmaraDev: It should be called from the AN module `isInside = AN.utils.bvh.isInsideVolume(bvhTree, vector)`.

3DSinghVFX: Okay, Thanks.

OmarEmaraDev:
Reducing the length to avoid misses is fine.

Gerstmann-Bradley:
Hi, I like this node, but I am also concerned with how this node will effect rendering.
In short, if @3DSinghVFX you keyframes your mesh, will you be able to see the animation?
In my test with mesh falloff in EN, it doesn't work if I disable the mesh in render/viewport/collection.
and I disable the mesh in render because I don't want to see it in render/viewport/collection.

Would you please check on this to see if this issue is replicable?

OmarEmaraDev:
@Gerstmann-Bradley If you want to test this pull request, you can go to the Checks tab above, then download the produced artifacts.



For the rendering issue, I am guessing this is related to how Kuldeep constructed the BVH tree from the object in the old version. In this version we leave the construction to the user, so this should fix the issue as far as I can tell.

3DSinghVFX:
> Hi, I like this node, but I am also concerned with how this node will effect rendering.
> In short, if @3DSinghVFX you keyframe your mesh, will you be able to render the animation using Ctrl+F12?
> In my test with mesh falloff in EN, it doesn't work if I disable the mesh in render/viewport/collection (Camera icon, and screen icon in outliner).
> and I disable the mesh in render because I don't want to see it in render/viewport/collection.
> 
> Would you please check on this to see if this issue is replicable?

Yes, I can replicate that.

> @Gerstmann-Bradley If you want to test this pull request, you can go to the Checks tab above, then download the produced artifacts.
> 
> 
> 
> For the rendering issue, I am guessing this is related to how Kuldeep constructed the BVH tree from the object in the old version. In this version we leave the construction to the user, so this should fix the issue as far as I can tell.


Omar, I can replicate the issue in this version as well. 
Disabling the *render* or *collection* also cause the problem for *Mesh Input* node especially when it has *modifiers*.

OmarEmaraDev:
I see. I guess this is the issue we discussed before, maybe T71908 for the keyframe issue.

3DSinghVFX:
> I see. I guess this is the issue we discussed before, maybe T71908 for the keyframe issue.

Yes, I also think so.

Gerstmann-Bradley:
Just to mention there surely will be workaround to solve it.
For example "viewport render animation" as discussed in the past in other post about mesh data reading. But A side effect is that people can't use "compositor" at the moment.
People can also disable "use world space", move falloff object beyond render region, move mesh using nodes...But if we have to build this part of node tree every time by our own as a work around, it may also be unpleasant.
If this turns to be the workaround you guys choose to do, I hope node can be made easier for people to use in such a case?
Or maybe there will be other solutions. Unfortunately I can't help on these issues.

Best regards.

animation_nodes/nodes/falloff/mesh_falloff.pyx
```diff
@@ -0,0 +1,73 @@
...
+
+class MeshFalloffNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_MeshFalloffNode"
+    bl_label = "Mesh Falloff"
+
+    def create(self):
+        self.newInput("BVHTree", "BVHTree", "bvhTree")
+        self.newInput("Float", "Size", "size")
+        self.newInput("Float", "Falloff Width", "falloffWidth", value = 1)
+        self.newInput("Boolean", "Use Surface", "useSurface", value = True)
```
OmarEmaraDev, line 18: `Use Surface` can be removed, as in, it should always be True. A False `Use Surface` and a True `Use Volume` is equivalent to a True `Use Surface` and a True `Use Volume` when Scale and Falloff Width are both zero. You may check for this condition to avoid extra computations implicitly. Moreover, `Use Volume` should be renamed to something more clear, maybe `Fill Inside`.

OmarEmaraDev (changes requested):

animation_nodes/nodes/falloff/mesh_falloff.pyx
3DSinghVFX: Okay.

OmarEmaraDev (approved):
Thanks!

We will come back to this later to try to improve the performance and usability.

3DSinghVFX:
> We will come back to this later to try to improve the performance and usability.

Sure :)

PR #1523, opened 2020-09-05
Add Filter Material List By Name.

3DSinghVFX:
I have added *Filter Material List By Name* node.


OmarEmaraDev (approved):
Thanks! Update the CHANGELOG and then we can merge.

PR #1524, opened 2020-09-05
Add Find Shortest Path node.

3DSinghVFX:
I have added *Find Shortest Path* node. It finds shortest paths (based on **[Dijkstra's algorithm](https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm)**) from given vertex or vertices to all other vertices of the mesh. It also works for mesh islands as well.

Inputs:
  - Mesh
  - Sources: It is an integer list for vertex-indices.
  - Change Direction: It allows to change the direction of paths.

Output:
  -  This node output *Mesh(s), Splines, GP Strokes*.
Screenshot from 2020-09-05 12-06-37

Examples:
untitled

Dijkstra_Shortest_Path7


OmarEmaraDev (commented):
Design-wise, can the node really find the shortest path between two vertices? It seems what we do is compute a shortest-path tree rooted at the source, with no way of actually finding any paths after generation.

3DSinghVFX:
> Design-wise, can the node really find the shortest path between two vertices? It seems what we do is compute a shortest-path tree rooted at the source, with no way of actually finding any paths after generation.

Omar, the *Dijkstra's algorithm* finds the paths from a vertex (source) to all other vertices. So, it is possible to get path between two vertices.

OmarEmaraDev:
Consider the use case where I want to find the shortest path between vertex A and vertex B. What will the node tree for this be?

3DSinghVFX:
> Consider the use case where I want to find the shortest path between vertex A and vertex B. What will the node tree for this be?

I have not added a method for Vertex A to Vertex B. The current version is for paths from Vertex A to other vertices.
However, it is possible to add an input for End / Destiny vertex-index(ies).

OmarEmaraDev:
That's my point. A *Find Shortest Path* node should at least find the shortest path, not output something that resembles a shortest-path tree in a form of a mesh or spline. Either we redesign the node to account for this or we make it a specialized node for this particular use case.

3DSinghVFX:
> That's my point. A _Find Shortest Path_ node should at least find the shortest path, not output something that resembles a shortest-path tree in a form of a mesh or spline. Either we redesign the node to account for this or we make it a specialized node for this particular use case.

I think we can add two options in this node:
  - **Index** - This will allow to find the shortest path from Vertex A to Vertex B.
  - **All** -  This will allow to find the shortest paths from Vertex(ies) A to all other vertices.

OmarEmaraDev:
Ok. Feel free to implement that and we will handle the details in review.

3DSinghVFX:
Omar, I have added *Index* and *All* modes.

Here is node-tree for shortest path from Vertex A to Vertex B:
find_shortest_path1

OmarEmaraDev (changes requested):
I don't think "Destiny" is the best word to use here. Perhaps "Target" is better.

Is there a good reason why Change Direction an option in this node? Wouldn't a simple Change Spline Direction node or a Reverse List node do the job?

I think the most important output for the single case is the actual indices of the path. I also don't think the output structures (Spline, Mesh, Stroke) are particularly useful in the single case. So I think there should be a single output, the indices of the path.

Finally, what do you think about "Path & Tree" instead of "Index & All"?

3DSinghVFX:
> I don't think "Destiny" is the best word to use here. Perhaps "Target" is better.

Okay.

> Is there a good reason why Change Direction an option in this node? Wouldn't a simple Change Spline Direction node or a Reverse List node do the job?

It can be done for single path `Reverse List node` and for spline-paths, but not for other path-types. 
 
> I think the most important output for the single case is the actual indices of the path. I also don't think the output structures (Spline, Mesh, Stroke) are particularly useful in the single case. So I think there should be a single output, the indices of the path.

Okay.
 
> Finally, what do you think about "Path & Tree" instead of "Index & All"?

I agree with you, these are more clearer.

OmarEmaraDev:
But does order matter in Mesh and Stroke structures? What are the possible applications?

3DSinghVFX:
> But does order matter in Mesh and Stroke structures? What are the possible applications?

Yes, *Build* modifier workes based on the order.

OmarEmaraDev:
Then add a Change Direction node for strokes later.

Gerstmann-Bradley:
Sorry to interrupt, will the "change direction" option still be kept? Won't that be easier for user if there is already an option available instead of adding an additional node?
I agree with adding Change Direction node for strokes though.

3DSinghVFX:
@OmarEmaraDev 
> Then add a Change Direction node for strokes later.

Okay. What about the mesh-paths?

OmarEmaraDev:
I think the option shouldn't be kept. I am not a fan of options that have a very narrow use cases especially that those cases are usually achieved through a single additional node. For mesh-paths, what is the use case? You mentioned the build modifier, but I don't think this is convincing enough to keep the option.

3DSinghVFX:
> I think the option shouldn't be kept. I am not a fan of options that have a very narrow use cases especially that those cases are usually achieved through a single additional node. For mesh-paths, what is the use case? You mentioned the build modifier, but I don't think this is convincing enough to keep the option.

Okay. I'll remove that option.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+
+from . line import getLinesMesh
+from ... math cimport distanceVec3
+
+# Dijkstra's algorithm (https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm) but is implemented
+# such a way to handle multi-sources and mesh with multiple islands.
+def getShortestPath(Mesh mesh, LongList sources, LongList targets, str pathType, str mode):
+    cdef Vector3DList vertices = mesh.vertices
+    cdef Py_ssize_t vertexCount = vertices.length
+    cdef float maxWeight = 1000000
```
OmarEmaraDev, line 20: Use INFINITY from `math.h`.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+    cdef DoubleList weights = DoubleList.fromValue(maxWeight, length = vertexCount)
+    cdef LongList previousVertices = LongList.fromValue(-1, length = vertexCount)
+    cdef LongList visitedVertices = LongList.fromValue(0, length = vertexCount)
```
OmarEmaraDev, line 23: Would a BooleanList make more sense here?

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+from . line import getLinesMesh
+from ... math cimport distanceVec3
+
+# Dijkstra's algorithm (https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm) but is implemented
+# such a way to handle multi-sources and mesh with multiple islands.
+def getShortestPath(Mesh mesh, LongList sources, LongList targets, str pathType, str mode):
+    cdef Vector3DList vertices = mesh.vertices
+    cdef Py_ssize_t vertexCount = vertices.length
+    cdef float maxWeight = 1000000
+    cdef DoubleList weights = DoubleList.fromValue(maxWeight, length = vertexCount)
```
OmarEmaraDev, line 21: Call this `distances`. Since our graph is unweighted.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+        if minWeight == maxWeight:
+            break
+
+        for currentIndex in range(vertexCount):
+            if abs(minWeight - weights.data[currentIndex]) < 1.0e-6:
+                visitedVertices.data[currentIndex] = 1
+
+                linkedVertices = mesh.getVertexLinkedVertices(currentIndex)[0]
+
+                for l in range(linkedVertices.length):
```
OmarEmaraDev, line 50: Shouldn't we filter the neighbors based on the visited array instead?

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+                minWeight = weight
+
+        if minWeight == maxWeight:
+            break
+
+        for currentIndex in range(vertexCount):
+            if abs(minWeight - weights.data[currentIndex]) < 1.0e-6:
+                visitedVertices.data[currentIndex] = 1
+
+                linkedVertices = mesh.getVertexLinkedVertices(currentIndex)[0]
```
OmarEmaraDev, line 48: Can we use `getLinkedVertices` directly for lower overhead?

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+
+                for l in range(linkedVertices.length):
+                    linkedIndex = linkedVertices.data[l]
+                    weight = distanceVec3(vertices.data + currentIndex, vertices.data + linkedIndex) + weights.data[currentIndex]
+                    if weight < weights.data[linkedIndex]:
+                        weights.data[linkedIndex] = weight
+                        previousVertices.data[linkedIndex] = currentIndex
```
OmarEmaraDev, line 55: For the PATH version, we can break early after we reach our target, right?

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,126 @@
...
+            strengths = FloatList(length = amount)
+            pressures = FloatList(length = amount)
+            uvRotations = FloatList(length = amount)
+            vertexColors = ColorList(length = amount)
+            strengths.fill(1)
+            pressures.fill(1)
+            uvRotations.fill(0)
+            vertexColors.fill((0, 0, 0, 0))
+            strokes.append(GPStroke(sortLocations.reversed(), strengths, pressures,
+                                    uvRotations, vertexColors, 10))
```
OmarEmaraDev, line 124: Seems like all of those share the same code. You should deduplicate.

animation_nodes/nodes/mesh/find_shortest_path.py
```diff
@@ -0,0 +1,108 @@
...
+        layout.prop(self, "mode", text = "")
+        layout.prop(self, "pathType", text = "")
+        if self.pathType == "MESH":
+            layout.prop(self, "joinMeshes")
+
+    def getExecutionFunctionName(self):
+        if self.mode == "PATH":
+            return "execute_Index"
+        else:
+            return "execute_All"
```
OmarEmaraDev, line 66: Rename those to Path and Tree

animation_nodes/nodes/mesh/find_shortest_path.py
```diff
@@ -0,0 +1,108 @@
...
+        targets = LongList.fromValue(target)
+        if self.pathType == "MESH":
+            if self.joinMeshes:
+                return Mesh.join(*getShortestPath(mesh, sources, targets, "MESH", "PATH"))
+            else:
+                return getShortestPath(mesh, sources, targets, "MESH", "PATH")
+        elif self.pathType == "SPLINE":
+            return getShortestPath(mesh, sources, targets, "SPLINE", "PATH")
+        elif self.pathType == "STROKE":
+            return getShortestPath(mesh, sources, targets, "STROKE", "PATH")
```
OmarEmaraDev, line 87: Why do we have this code if the path type won't matter?

animation_nodes/nodes/mesh/find_shortest_path.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if self.pathType == "MESH":
+            if self.joinMeshes:
+                return Mesh.join(*getShortestPath(mesh, sources, LongList(), "MESH", "TREE"))
+            else:
+                return getShortestPath(mesh, sources, LongList(), "MESH", "TREE")
+        elif self.pathType == "SPLINE":
+            return getShortestPath(mesh, sources, LongList(), "SPLINE", "TREE")
+        elif self.pathType == "STROKE":
+            return getShortestPath(mesh, sources, LongList(), "STROKE", "TREE")
```
OmarEmaraDev, line 108: `self.pathType` can be passed directly instead of branching.

OmarEmaraDev (changes requested):

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Omar, this `getLinkedVertices` gives
```
neighboursAmounts, neighboursStarts, neighbours, neighbourEdges = self.getLinkedVertices()
```
Do you mean we calculate this at the beginning then use it later?

animation_nodes/nodes/mesh/find_shortest_path.py
3DSinghVFX: Done.

animation_nodes/nodes/mesh/find_shortest_path.py
3DSinghVFX: Sorry. I did not realize that. It is removed.

animation_nodes/nodes/mesh/find_shortest_path.py
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Yes. Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Omar, We have to update the weight/distance of each neighbor vertex and update their previous vertex index. The current vertex can have visited and non-visited vertices.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
OmarEmaraDev: Yes, call the function one time at the beginning outside the loop. You will then sample the data of `neighboursAmounts, neighboursStarts, neighbours` to get the indices you want.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
OmarEmaraDev: Ok. I will look at this in more details once I get the time.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Okay.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Okay.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

OmarEmaraDev:
@JacquesLucke Any thoughts about the Tree option here? The option produces n number of partially duplicate paths for n number of vertices, which seems rather messy to me. To see this better, I randomized the control points a bit.



@3DSinghVFX I am looking at the output tree and it doesn't seem to be correct to me. Shouldn't the shortest path between 0 and 16 in the following image be a single edge? Is it not supposed to be a shortest path tree?

3DSinghVFX:
> @3DSinghVFX I am looking at the output tree and it doesn't seem to be correct to me. Shouldn't the shortest path between 0 and 16 in the following image be a single edge? Is it not supposed to be a shortest path tree?
> 
> 

@OmarEmaraDev What is the *source index*? Can you please upload the file?
It looks fine on my side:
Screenshot from 2020-09-14 00-45-10

3DSinghVFX:
> @JacquesLucke Any thoughts about the Tree option here? The option produces n number of partially duplicate paths for n number of vertices, which seems rather messy to me. To see this better, I randomized the control points a bit.
> 
> 

If required, then overlapping can easily be removed with help of *BMesh* nodes or *Weld* modifier:
Screenshot from 2020-09-14 00-58-16

OmarEmaraDev:
The source is zero. Here is the file:

[shortest_path_tree.zip](https://github.com/JacquesLucke/animation_nodes/files/5214994/shortest_path_tree.zip)

OmarEmaraDev:
> If required, then overlapping can easily be removed with help of BMesh nodes or Weld modifier.

Sure. I am not exactly against, I just want to look at other possibilities.

3DSinghVFX:
> The source is zero. Here is the file:
> 
> [shortest_path_tree.zip](https://github.com/JacquesLucke/animation_nodes/files/5214994/shortest_path_tree.zip)

@OmarEmaraDev It is working fine,
shortest_path_tree

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,112 @@
...
+    Vector3DList,
+)
+
+from libc.math cimport INFINITY
+from . line import getLinesMesh
+from ... math cimport distanceVec3
+
+# Dijkstra's algorithm (https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm) but is implemented
+# such a way to handle multi-sources and mesh with multiple islands.
+def getShortestPath(Mesh mesh, LongList sources, LongList targets, str pathType, str mode):
```
OmarEmaraDev, line 19: Why is `targets` a list?

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,112 @@
...
+            distance = distances.data[k]
+            if distance < minDistance and not visitedVertices.data[k]:
+                minDistance = distance
+
+        if minDistance == maxDistance:
+            break
+
+        for currentIndex in range(vertexCount):
+            if abs(minDistance - distances.data[currentIndex]) < 1.0e-6:
+                visitedVertices.data[currentIndex] = True
```
OmarEmaraDev, line 54: If you want the index of the vertex with the smallest distance, you compute it directly, don't do two nested loops. Every time you assign `minDistance = distance` you also assign `k` to an index variable.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,112 @@
...
+        targetIndex = targets.data[0]
+    else:
+        targetIndex = <int>INFINITY
+
+    cdef float distance, minDistance
+    cdef Py_ssize_t k, l, currentIndex, neighboursStart, neighbourIndex
+    cdef LongList neighboursAmounts, neighboursStarts, neighbours
+    neighboursAmounts, neighboursStarts, neighbours = mesh.getLinkedVertices()[:3]
+
+    for i in range(vertexCount):
```
OmarEmaraDev, line 41: This can be `while numberOfVisitedVertices != vertexCount:`. This should avoid mid-function break.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,112 @@
...
+    cdef BooleanList visitedVertices = BooleanList.fromValue(False, length = vertexCount)
+    cdef Py_ssize_t i, targetIndex
+
+    for i in range (sources.length):
+        distances.data[sources.data[i]] = 0.0
+
+    if mode == "PATH":
+        targetIndex = targets.data[0]
+    else:
+        targetIndex = <int>INFINITY
```
OmarEmaraDev, line 34: You should split this function into multiple functions. A function `computeShortestPathTree` which returns the `previousVertices` list. This function takes a target, this target can be set to `-1` if we want to compute the whole tree. The function implements that transparently as you do now.

Then, the two separate functions that takes the previouseVertices array will generate the path array and the tree structures.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,112 @@
...
+
+        if minDistance == maxDistance:
+            break
+
+        for currentIndex in range(vertexCount):
+            if abs(minDistance - distances.data[currentIndex]) < 1.0e-6:
+                visitedVertices.data[currentIndex] = True
+
+                neighboursStart = neighboursStarts.data[currentIndex]
+                for l in range(neighboursAmounts.data[currentIndex]):
```
OmarEmaraDev, line 57: Don't use `l` as a variable name. `i, j, and k` are usually used.

OmarEmaraDev (changes requested):

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Omar, we need to compute *index of the vertex with the smallest distance* which is not visited yet.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Removed.

3DSinghVFX: However, with `targets` a list we can extend the functionality for target-indices, and outputs tree.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
OmarEmaraDev: Isn't this why we have `not visitedVertices.data[k]:`?

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
OmarEmaraDev: Okay, but lets finish an initial version first then look at possible ways to extend it.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Yes.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Okay.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
OmarEmaraDev: So what is the difficulty?

3DSinghVFX: I do not how to compute it directly.

OmarEmaraDev:
```cython
        minDistance = maxDistance
        cdef Py_ssize_t minDistanceIndex 
        for k in range(vertexCount):
            distance = distances.data[k]
            if not visitedVertices.data[k] and distance < minDistance:
                minDistanceIndex = k
                minDistance = distance

```

3DSinghVFX: Okay. Omar this is possible when there is the only one source. If we have multiple sources then we can have then multiple vertices that can have the same `minDistance` value. They generate different results:
**With Loop**
With_Loop

**With minDistanceIndex**
With_Index

3DSinghVFX: Omar, I have tested both methods for different cases, path topology is different but path length is the same.

OmarEmaraDev: So it is different but correct? Are you saying we shouldn't use the index method?

3DSinghVFX: **With Loop**: In this method, we are updating the distance (and previous index) of all `min distance` vertices which lies in region/reach of each source. It is like parallel update. 

**With Index**: In this method, we are updating the distance (and previous index) of the first `min distance` vertex. It is like series update.

The difference in topology is mainly pronouced when base mesh has edges with same length e.g., Grid Mesh.

OmarEmaraDev: Notice that the loop is slower. So if the difference in topology is not important, we should use the index method.

3DSinghVFX: Yes. I agree with you.

OmarEmaraDev (changes requested):
Looks better already. But I want you to split the getShortestPathTree function again into two function, one for PATH and one for the TREE. Removing the `mode` input.

3DSinghVFX:
> Looks better already. But I want you to split the getShortestPathTree function again into two function, one for PATH and one for the TREE. Removing the `mode` input.

Okay.

3DSinghVFX:
@OmarEmaraDev, This method calculates path or tree from source(s) to target(all). Therefore, we can cache them to make it even faster especially in the case of *Path* mode because calculation remains valid unless some change source.

3DSinghVFX:
> @OmarEmaraDev, This method calculates path or tree from source(s) to target(all). Therefore, we can cache them to make it even faster especially in the case of _Path_ mode because calculation remains valid unless some change source.

What do you think?

OmarEmaraDev:
How are you going to implement the caching?

3DSinghVFX:
> How are you going to implement the caching?

I think if we can make these methods as mesh-mesthods then probaly we can cache them. But I do not know whether it will work in Group or Loop nodes.

OmarEmaraDev:
Yes, but notice that the current caching mechanism we have doesn't take inputs as keys, so you will have to extend that in order to handle cases like this.

3DSinghVFX:
> Yes, but notice that the current caching mechanism we have doesn't take inputs as keys, so you will have to extend that in order to handle cases like this.

I know that but it seems difficult for me to do.

Query:
Do the  *execution or any function* of the node has a unique ID?

OmarEmaraDev:
If you mean a per-node thing. The node identifier is what you are looking for.

OmarEmaraDev:
Lets finish this one first and I will work on caching later.

3DSinghVFX:
> Lets finish this one first and I will work on caching later.

Okay.

3DSinghVFX:
> If you mean a per-node thing. The node identifier is what you are looking for.

Omar, thanks, I know that. I want to understand how do Group and Loop nodes work. Is they make a gaint function from individual nodes then make it unique ID according to Invoke Subprogram or something else?

OmarEmaraDev:
Yes. Something like that. Subprograms are converted into a function. When you invoke a subprogram, Animation Nodes will make this function available in the variable space of the execution unit and will essentially call it just like any function.

For instance, if I create a group with a single Number Add node and two inputs, Animation Nods will construct this function:

```
 17.  def main(group_input_0, group_input_1):
 18.
 19.
 20.      # Node: 'NodeTree' - 'Float Math'
 21.      _result_cgt2 = group_input_0 + group_input_1
 22.
 23.
 24.      return _result_cgt2
```

And if I invoke this subprogram, Animation Nodes will emit this:

```
  1.
  2.  # Node: 'NodeTree' - 'Invoke Subprogram'
  3.  _xbddtetjqf_ao82 = _subprogram_sggiqt6oucqrc3y(_hweuuxewgr_ao80, _dgfhqkevns_ao81)
```

`_subprogram_sggiqt6oucqrc3y` here is the `main` function above, but it was renamed to have a unique name. The `sggiqt6oucqrc3y` is the subprogram identifier.

The same goes for loops. Except Animation Nodes insert a for loop inside the function to implement iteration.

3DSinghVFX:
> Yes. Something like that. Subprograms are converted into a function. When you invoke a subprogram, Animation Nodes will make this function available in the variable space of the execution unit and will essentially call it just like any function.
> 
> For instance, if I create a group with a single Number Add node and two inputs, Animation Nods will construct this function:
> 
> ```
>  17.  def main(group_input_0, group_input_1):
>  18.
>  19.
>  20.      # Node: 'NodeTree' - 'Float Math'
>  21.      _result_cgt2 = group_input_0 + group_input_1
>  22.
>  23.
>  24.      return _result_cgt2
> ```
> 
> And if I invoke this subprogram, Animation Nodes will emit this:
> 
> ```
>   1.
>   2.  # Node: 'NodeTree' - 'Invoke Subprogram'
>   3.  _xbddtetjqf_ao82 = _subprogram_sggiqt6oucqrc3y(_hweuuxewgr_ao80, _dgfhqkevns_ao81)
> ```
> 
> `_subprogram_sggiqt6oucqrc3y` here is the `main` function above, but it was renamed to have a unique name. The `sggiqt6oucqrc3y` is the subprogram identifier.
> 
> The same goes for loops. Except Animation Nodes insert a for loop inside the function to implement iteration.

Omar, thank you.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,113 @@
...
+def getShortestPath(Mesh mesh, LongList sources, Py_ssize_t target):
+    cdef LongList previousVertices = computeShortestPathTree(mesh, sources, target)
+    cdef LongList indices
+    cdef Py_ssize_t index
+
+    indices = LongList()
+    index = target
+    if previousVertices.data[index] == -1: return indices
+    while index != -1:
+        indices.append(index)
```
OmarEmaraDev, line 28: Use `append_LowLevel` instead.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,113 @@
...
+        index = previousVertices.data[index]
+
+    return indices.reversed()
+
+def getShortestTree(Mesh mesh, LongList sources, str pathType):
+    cdef Vector3DList vertices = mesh.vertices
+    cdef long vertexCount = vertices.length
+    cdef LongList previousVertices = LongList(length = vertexCount)
```
OmarEmaraDev, line 36: This is redundant. You are allocating here and another time in in the function below.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,113 @@
...
+    cdef Vector3DList sortLocations
+    cdef Py_ssize_t i, index
+    cdef list tree = []
+
+    for i in range(vertexCount):
+        sortLocations = Vector3DList()
+        index = i
+        if previousVertices.data[index] == -1: continue
+        while index != -1:
+            sortLocations.append(vertices[index])
```
OmarEmaraDev, line 48: Use `append_LowLevel` here.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,113 @@
...
+    elif pathType == "STROKE":
+        amount = sortLocations.length
+        strengths = FloatList(length = amount)
+        pressures = FloatList(length = amount)
+        uvRotations = FloatList(length = amount)
+        vertexColors = ColorList(length = amount)
+        strengths.fill(1)
+        pressures.fill(1)
+        uvRotations.fill(0)
+        vertexColors.fill((0, 0, 0, 0))
```
OmarEmaraDev, line 70: It will be faster if this was `fill(0)`.

animation_nodes/nodes/mesh/find_shortest_path.py
```diff
@@ -0,0 +1,93 @@
...
+                else:
+                    self.newOutput("Mesh", "Mesh", "outMesh")
+            elif self.pathType == "SPLINE":
+                self.newOutput("Spline List", "Splines", "outSplines")
+            elif self.pathType == "STROKE":
+                self.newOutput("GPStroke List", "Strokes", "outStrokes")
+
+    def draw(self, layout):
+        layout.prop(self, "mode", text = "")
+        layout.prop(self, "pathType", text = "")
```
OmarEmaraDev, line 58: This shouldn't be drawn if the mode is PATH.

animation_nodes/nodes/mesh/find_shortest_path.py
```diff
@@ -0,0 +1,93 @@
...
+
+    joinMeshes: BoolProperty(name = "Join Meshes", default = True,
+        update = AnimationNode.refresh)
+
+    useSourceList: VectorizedSocket.newProperty()
+
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
+        if self.mode == "PATH":
+            self.newInput("Integer", "Source", "source")
```
OmarEmaraDev, line 37: Isn't it possible to make this vectorized even in the PATH mode. In this case, it will be the shortest path to the closest source.

animation_nodes/nodes/mesh/find_shortest_path.py
```diff
@@ -0,0 +1,93 @@
...
+            return "execute_Path"
+        else:
+            return "execute_Tree"
+
+    def execute_Path(self, mesh, source, target):
+        if mesh is None:
+            return LongList()
+
+        if source < 0 or source >= len(mesh.vertices):
+            self.raiseErrorMessage("Some indices are out of range.")
```
OmarEmaraDev, line 73: mention specifically which indices. Same for the one below.

OmarEmaraDev (changes requested):

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
```diff
@@ -0,0 +1,113 @@
...
+    BooleanList,
+    Vector3DList,
+)
+
+from libc.math cimport INFINITY
+from . line import getLinesMesh
+from ... math cimport distanceVec3
+
+# Dijkstra's algorithm (https://en.wikipedia.org/wiki/Dijkstra%27s_algorithm) but is implemented
+# such a way to handle multi-sources and mesh with multiple islands.
```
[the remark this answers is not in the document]
3DSinghVFX: Okay.

animation_nodes/nodes/mesh/find_shortest_path.py
3DSinghVFX: Omar, I think that may confuse the user.
To find the *shortest path to the closest source*
1. We have to find a path from all sources to the target.
2. Then check the path length for all the paths.
3. Output the path which is the shortest or closest source path.

OmarEmaraDev: I am not sure about the confusion that this might cause. By default, it will be a single value. So a user will not have to deal with multiple values unless they want. Do we really have to go through all those steps? I though simply setting a list of sources will make things work. If it it is not that case, then yo can ignore this for now.

3DSinghVFX: > I am not sure about the confusion that this might cause. By default, it will be a single value. So a user will not have to deal with multiple values unless they want. 

Okay.

> Do we really have to go through all those steps?

Yes.

> I though simply setting a list of sources will make things work. If it it is not that case, then you can ignore this for now.

Okay.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: But we need *Vertices* to construct the paths. Do I make the change in the `computeShortestPathTree` method so that it also returns *Vertices* as well?

OmarEmaraDev: I mean `LongList(length = vertexCount)` is reduandant. It will be overwritten anyway in the next line.

animation_nodes/nodes/mesh/find_shortest_path.py
3DSinghVFX: Done.

animation_nodes/algorithms/mesh_generation/find_shortest_path.pyx
3DSinghVFX: Okay.

animation_nodes/nodes/mesh/find_shortest_path.py
3DSinghVFX: Done.

OmarEmaraDev (approved):
Thanks!

PR #1528, opened 2020-09-11
Add a "Random Boolean" node

kdsin23:


OmarEmaraDev (changes requested):
Perhaps it would be more useful if we have a probability input. 0 being all false and 1 being all true.

Also, merge upstream and fix the conflicts.

kdsin23:
> Perhaps it would be more useful if we have a probability input. 0 being all false and 1 being all true.
> 
> Also, merge upstream and fix the conflicts.

good idea! i will try

OmarEmaraDev:
One thing you shouldn't do is commit your changes to master in your fork. Create another branch and commit to that instead. Your master should be identical to Jacques' master.

I am currently seeing changes from previous commits, can you fix that?

kdsin23:
> One thing you shouldn't do is commit your changes to master in your fork. Create another branch and commit to that instead. Your master should be identical to Jacques' master.
> 
> I am currently seeing changes from previous commits, can you fix that?

i am sorry for this, i not sure i can fix this, because i am not familar with the github

OmarEmaraDev:
Can you close this and create a new pull request to fix this?

OmarEmaraDev (approved):
Thanks!

Issue #1363, opened 2020-03-16
When to use cython ?


harisreedhar:
I have few doubts in coding on AN,

1. What are the best cases to use cython for animation nodes ? 
2. What is a virtuallist ?
3. How yield works to return list ? I saw they are written inside double quotes
4. How to check a socket is empty or not ?

OmarEmaraDev:
1. We use Cython in two cases. (1) When we want to use C/C++ code in Animation Nodes. (2) If writing the code in Cython would result in better performance. This is usually the case for algorithms that have a large input size. Some code is always written in Cython, like Falloffs.
2. A virtual list is a list that allow indexing outside of its bound by repeating or filling itself. For instance, if `a = virtual([0, 1, 2, 3])`, then `a[3] -> 3, a[4] -> 0, a[5] -> 1, ...`.
3. This is called a "Generator Function". Consult the documentation for how this works.
4. I assume you mean if a socket is linked or not. In this case, you can use the `required` input of the `getExecutionCode` function. For example, look at the sort node.

harisreedhar:
Thanks @OmarEmaraDev

PR #1530, opened 2020-09-15
Vectorized Switch Node

kdsin23:
vectorized switch node

### DIFF
--- a/animation_nodes/nodes/boolean/switch_node.py
+++ b/animation_nodes/nodes/boolean/switch_node.py
@@ -1,21 +1,29 @@
 import bpy
 from bpy.props import *
-from ... base_types import AnimationNode, DataTypeSelectorSocket
+from ... sockets.info import toListDataType, isList
+from ... base_types import AnimationNode, DataTypeSelectorSocket, VectorizedSocket
 
 class SwitchNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_SwitchNode"
     bl_label = "Switch"
 
     assignedType: DataTypeSelectorSocket.newProperty(default = "Float")
+    useList: VectorizedSocket.newProperty()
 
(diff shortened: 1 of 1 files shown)


animation_nodes/nodes/boolean/switch_node.py
```diff
@@ -27,5 +35,24 @@ def assignType(self, dataType):
...
+            else:
+                outputs.append(ifFalse)
+        if not isList(self.assignedType):
+            listDataType = toListDataType(self.assignedType)
+            return self.outputs[0].correctValue(outputs)[0]
+        else:
+            outs = []
+            for i in range(len(conditions)):
+                outs.extend(outputs[i])
+            return self.outputs[0].correctValue(outs)[0]
```
OmarEmaraDev, line 58: A virtual list can be used here. Are you familiar with those?

OmarEmaraDev (changes requested):
I am not sure if this is the best way to vectorize this node. We probably need to do work around socket templates to make things easier in the future. I will get back to you on this later.

animation_nodes/nodes/boolean/switch_node.py
kdsin23: Sorry, i am not familiar with virtual list, and have never used it.

OmarEmaraDev: Virtual Lists are described in https://docs.animation-nodes.com/release_notes/v2_1/#virtual-lists. Also see #1363. A lot of nodes use them, for instance, look at any of the Geometry nodes. Let me know if you need further help in understanding them.

kdsin23: Thank you very much! i will try my best to understand them.

PR #1537, opened 2020-09-18
Fix wrong order of inputs in Random Color node.

3DSinghVFX:
I have fixed the wrong order of *seed* and *count* inputs in the *Random Color* node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@
 ### Fixed
 
 - Fixed nextBoolean method for XoShiRo256StarStar generators.
-- Fixed wrong order of seed and count inputs in *Random Color* node.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 ### Fixed
 
 - Fixed nextBoolean method for XoShiRo256StarStar generators.
+- Fixed wrong order of seed and count inputs in *Random Color* node.
```
OmarEmaraDev, line 14: We don't add CHANGELOG entries for changes in the same subversion.

OmarEmaraDev (changes requested):

CHANGELOG.md
3DSinghVFX: Okay, thanks for the info.

OmarEmaraDev (approved):
Thanks!

PR #1549, opened 2020-10-05
Revectorized the convert rotations node

kdsin23:
clean up the codes


animation_nodes/nodes/rotation/c_utils.pyx
```diff
@@ -121,3 +122,139 @@ def randomQuaternionList(int seed, int amount):
         quaternionNormalize_InPlace(result.data + i)
 
     return result
+
```
OmarEmaraDev, line 125: As we said in the original pull request. I want `cdef` functions to do each of the conversions in the math module. Then those functions can be used here.

animation_nodes/nodes/rotation/convert_rotations.py
```diff
@@ -13,6 +15,16 @@
     ("QUATERNION_TO_AXIS_ANGLE", "Quaternion to Axis Angle", "", "NONE", 6),
     ("AXIS_ANGLE_TO_QUATERNION", "Axis Angle to Quaternion", "", "NONE", 7) ]
 
+ListTypeItems = {
```
OmarEmaraDev, line 18: This is not used.

animation_nodes/nodes/rotation/convert_rotations.py
```diff
@@ -24,39 +36,60 @@ class ConvertRotationsNode(bpy.types.Node, AnimationNode):
 
     conversionType: EnumProperty(name = "Conversion Type", default = "QUATERNION_TO_EULER",
         items = conversionTypeItems, update = AnimationNode.refresh)
+        
+    listTypeItems: StringProperty(default = "QUATERNION_TO_EULER", update = AnimationNode.refresh)
```
OmarEmaraDev, line 40: Also not used.

animation_nodes/nodes/rotation/convert_rotations.py
```diff
@@ -66,30 +99,74 @@ def draw(self, layout):
 
     def drawLabel(self):
         for item in conversionTypeItems:
-            if self.conversionType == item[0]: return item[1]
+            if self.conversionType == item[0]:
+                if self.useList:
+                    return listTypeItems[item[0]]
```
OmarEmaraDev, line 104: You can just leave this method as it was.

OmarEmaraDev (changes requested):

animation_nodes/nodes/rotation/c_utils.pyx
kdsin23: OK, let me try.

animation_nodes/nodes/rotation/convert_rotations.py
kdsin23: then, how to change the label draw when chenging to list type? can you help me?

3DSinghVFX:
@kdsin23 I think you accidentally merged the https://github.com/JacquesLucke/animation_nodes/pull/1549/commits/c092cbc5b9ec869c20fd03be71adb64fe2889c69.

kdsin23:
> @kdsin23 I think you accidentally merged the [c092cbc](https://github.com/JacquesLucke/animation_nodes/commit/c092cbc5b9ec869c20fd03be71adb64fe2889c69).

yes, but i have a trouble with some functions
qta
why the angles are not correct?

3DSinghVFX:
> > @kdsin23 I think you accidentally merged the [c092cbc](https://github.com/JacquesLucke/animation_nodes/commit/c092cbc5b9ec869c20fd03be71adb64fe2889c69).
> 
> yes, but i have a trouble with some functions
> qta
> why the angles are not correct?

The reason for that is you did not declare `a` (angle) as a pointer

https://github.com/JacquesLucke/animation_nodes/blob/e55204b07b782870c7993dc31d89b98f79210e54/animation_nodes/math/conversion.pyx#L197-L205

So, you can write something like,
```
cdef quaternionToAxisAngle(Vector3* v, float* a, Quaternion* q):
    quaternionNormalize_InPlace(q)
    cdef float k = sqrt(1 - q.w * q.w)

    v.x = q.x / k
    v.y = q.y / k
    v.z = q.z / k

    a[0] = <float>(2 * acos(q.w))
```
and

```
def quaternionListToAxisListAngleList(QuaternionList qs, bint useDegree = False):
    cdef long amount = qs.length
    cdef Vector3DList vs = Vector3DList(length = amount)
    cdef DoubleList angles = DoubleList(length = amount)
    cdef float angle
    cdef Py_ssize_t i

    for i in range(amount):
        quaternionToAxisAngle(vs.data + i, &angle, qs.data + i)

        normalizeVec3_InPlace(vs.data + i)
        if useDegree:
            angles *= radianToDegreeFactor
        angles.data[i] = angle

    return vs, angles
```

kdsin23:
> > > @kdsin23 I think you accidentally merged the [c092cbc](https://github.com/JacquesLucke/animation_nodes/commit/c092cbc5b9ec869c20fd03be71adb64fe2889c69).
> > 
> > 
> > yes, but i have a trouble with some functions
> > qta
> > why the angles are not correct?
> 
> The reason for that is you did not declare `a` (angle) as a pointer
> 
> https://github.com/JacquesLucke/animation_nodes/blob/e55204b07b782870c7993dc31d89b98f79210e54/animation_nodes/math/conversion.pyx#L197-L205
> 
> So, you can write something like,
> 
> ```
> cdef quaternionToAxisAngle(Vector3* v, float* a, Quaternion* q):
>     quaternionNormalize_InPlace(q)
>     cdef float k = sqrt(1 - q.w * q.w)
> 
>     v.x = q.x / k
>     v.y = q.y / k
>     v.z = q.z / k
> 
>     a[0] = <float>(2 * acos(q.w))
> ```
> 
> and also make the changes in the `conversion.pxd` file, and
> 
> ```
> def quaternionListToAxisListAngleList(QuaternionList qs, bint useDegree = False):
>     cdef long amount = qs.length
>     cdef Vector3DList vs = Vector3DList(length = amount)
>     cdef DoubleList angles = DoubleList(length = amount)
>     cdef float angle
>     cdef Py_ssize_t i
> 
>     for i in range(amount):
>         quaternionToAxisAngle(vs.data + i, &angle, qs.data + i)
> 
>         normalizeVec3_InPlace(vs.data + i)
>         if useDegree:
>             angle *= radianToDegreeFactor
>         angles.data[i] = angle
> 
>     return vs, angles
> ```

Thank you so much!

kdsin23:
is it nessiary to move the `quaternionToAxisAngle()` to the place where the quaternion.pyx file inside? because there is a quaFromAxisAngle() funciton in the quaternion.pyx file.

3DSinghVFX:
> is it nessiary to move the `quaternionToAxisAngle()` to the place where the quaternion.pyx file inside? because there is a quaFromAxisAngle() funciton in the quaternion.pyx file.

@kdsin23 Yes.

PR #1542, opened 2020-10-01
Vectorized Mix nodes

3DSinghVFX:
I have vectorized all *Mix* nodes.

PR #1552, opened 2020-10-08
Added Virtual Quaternion List

harisreedhar:
This patch adds virtual list support for Quaternion  List


OmarEmaraDev:
Thanks! But Kuldeep added that already in #1542, which I will merge in a moment.

PR #1554, opened 2020-10-10
Improved matrix mix function

harisreedhar:
Added matrix mix functions in cython. 
old_matrixlerp

newMtarixlerp


3DSinghVFX:
@harisreedhar I have tried that but the result did not match if you mix matrices with `Blender` function (thin lines) and component-wise mixing (thick lines),
Mix_Matrix_Issue

and difference became significant if `factor > 1',
Mix_Matrix_Issue1

Can you please check that?

harisreedhar:
> @harisreedhar Hi. I have tried that but the result did not match if you mix matrices with the `Blender` function (long lines) and component-wise mixing (small lines),
> Mix_Matrix_Issue
> 
> and the difference became significant if `factor > 1` or `< 0`,
> Mix_Matrix_Issue1
> 
> Can you please check that?

That's because quaternion mix is a lerp. In blender they use slerp(spherical lerp) to mix matrices. If we use quaternion slerp we get same result.

matlerpslerp

3DSinghVFX:
Okay, then it is fine.

OmarEmaraDev:
@3DSinghVFX So you agree with approach as opposed to the one you were researching?

@harisreedhar Would you implement slerp-based Quaternion mixing in a separate pull request first?

3DSinghVFX:
> @3DSinghVFX So you agree with the approach as opposed to the one you were researching?

@OmarEmaraDev Yes. I found that there is a relation between 4x4 real orthogonal matrices and quaternions ([link1](https://archive.gamedev.net/archive/reference/articles/article428.html), [link2](https://www.euclideanspace.com/maths/algebra/matrix/orthogonal/)) so we can use quaternion matrix. 

This is the result of the current patch (long lines) and blender's method (short lines), both are same,
Mix_Matrix_Issue2


But we need to handle the axis flip as mentioned in [Blender Code](https://github.com/blender/blender/blob/743eca0f316d1ec2ef16e6066d7cfc5233c92657/source/blender/blenlib/intern/math_matrix.c#L2390-L2404), here is the result from the current patch (long lines) and blender's method (short lines),
Mix_Matrix_Issue
Mix_Matrix_Issue


For negative scale (one or more axis), they do not agree with each other and both do not give expected results, e.g.,
Mix_Matrix_Issue1

OmarEmaraDev:
@3DSinghVFX The flip problem is the one we handled before as in 55ccdb9b5a5555127c1f6916e2aa85a3cce9d808?

3DSinghVFX:
> @3DSinghVFX The flip problem is the one we handled before as in [55ccdb9](https://github.com/JacquesLucke/animation_nodes/commit/55ccdb9b5a5555127c1f6916e2aa85a3cce9d808)?

@OmarEmaraDev This flip problem is related to the technique that is used in this patch and mentioned in [Blender's Code](https://github.com/blender/blender/blob/743eca0f316d1ec2ef16e6066d7cfc5233c92657/source/blender/blenlib/intern/math_matrix.c#L2390-L2404)

harisreedhar:
> @harisreedhar Would you implement slerp-based Quaternion mixing in a separate pull request first?

ok will do that

harisreedhar:
> > @3DSinghVFX The flip problem is the one we handled before as in [55ccdb9](https://github.com/JacquesLucke/animation_nodes/commit/55ccdb9b5a5555127c1f6916e2aa85a3cce9d808)?
> 
> @OmarEmaraDev This flip problem is related to the technique that is used in this patch and mentioned in [Blender's Code](https://github.com/blender/blender/blob/743eca0f316d1ec2ef16e6066d7cfc5233c92657/source/blender/blenlib/intern/math_matrix.c#L2390-L2404)

This patch is actually implemented based on [blend_m4_m4m4](https://github.com/blender/blender/blob/743eca0f316d1ec2ef16e6066d7cfc5233c92657/source/blender/blenlib/intern/math_matrix.c#L2331) function.

OmarEmaraDev:
Will you update this patch?

harisreedhar:
> Will you update this patch?

I haven't found solution for flipping and negative scale issue mentioned by @3DSinghVFX. If you are ok with current patch i can update this.

OmarEmaraDev:
I will take a look and get back to you.

OmarEmaraDev:
@harisreedhar If you are still interested in this patch, @IyadAhmed previously proposed an implementation of polar decomposition based on numpy's [`svd`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.svd.html) method. Maybe you can consider using that to implement the Blender function in AN.

harisreedhar:
@OmarEmaraDev Thanks, Where i can find his implementation?

OmarEmaraDev:
@harisreedhar He didn't have a complete implementation, but you can implement it yourself. You need to implement the [`interp_m3_m3m3`](https://github.com/blender/blender/blob/master/source/blender/blenlib/intern/math_matrix.c#L2530) function instead of `blend_m4_m4m4`. This function calls [`mat3_polar_decompose`](https://github.com/blender/blender/blob/master/source/blender/blenlib/intern/math_matrix.c#L2304), which in turn, calls `BLI_svd_m3`. This function is essentially what the numpy `svd` method does, so you can use it and the rest of the implementation should be straightforward. Since we want to apply the function on a list of matrices, you can use broadcasting as described [here](https://numpy.org/doc/stable/reference/routines.linalg.html#linear-algebra-on-several-matrices-at-once).

But no pressure, I just though to let you know and see if you want to take another look at it.

harisreedhar:
@OmarEmaraDev Thanks for the explanation I'll take a look.

iyadahmed:
> @harisreedhar If you are still interested in this patch, @IyadAhmed previously proposed an implementation of polar decomposition based on numpy's [`svd`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.svd.html) method. Maybe you can consider using that to implement the Blender function in AN.

Hi, yes I tried but performance didn't improve, maybe I'm missing something 

Also recently I tried using NumPy with Cython, had success but not complete yet

OmarEmaraDev:
@iyadahmed Do you still have the implementation? I can have a look.

iyadahmed:
> @iyadahmed Do you still have the implementation? I can have a look.

hi not complete, I tested performance by timing it, with broadcasting not much improvement, timing is same or slower than old method

here's the non complete script I sent before https://gist.github.com/iyadahmed/e168b7726606206563b1774b1c5a6b3b

OmarEmaraDev:
Alright, thanks @iyadahmed! I guess the overhead of the python call is not significant relative to the execution time of the decomposition. If @harisreedhar doesn't have anything else to add, then we can close this.

iyadahmed:
Great, I might tinker at it too and report results if any

harisreedhar:
Yes, Polar decomposition alone is slower than old implementation.

decompose_mat

OmarEmaraDev:
Then I guess there is no reason to attempt a custom implementation for now. Thank you both!

PR #1556, opened 2020-10-11
Added Quaternion Slerp function

harisreedhar:


OmarEmaraDev:
Update the Quaternion mix node with this function. No need to make it an option, it should be the default.

harisreedhar:
> Update the Quaternion mix node with this function. No need to make it an option, it should be the default.

Updated Quaternion mix code

OmarEmaraDev (changes requested):
This doesn't handle the single case in the Mix Quaternion node. Quaternion mathutils already have a slerp, so you can use that.

harisreedhar:
Updated single case in the Mix Quaternion node

OmarEmaraDev:
Can you confirm that the single and multiple cases yield identical or sufficiently similar results?

harisreedhar:
> Can you confirm that the single and multiple cases yield identical or sufficiently similar results?

There are some mismatch in signs and blender's slerp can't go below or above 0 - 1 range
quat_test

harisreedhar:
The code is opted from [here](https://gitlab.com/bztsrc/slerp-opt)

OmarEmaraDev:
We probably don't need to worry about the sign here, because as a representation of rotation, they are both identical. However, we can't have the [0, 1] limit. So you should probably implement your own single method as well. Maybe put it in utils.math just like the euler method.

harisreedhar:
> We probably don't need to worry about the sign here, because as a representation of rotation, they are both identical. However, we can't have the [0, 1] limit. So you should probably implement your own single method as well. Maybe put it in utils.math just like the euler method.

should i add same copy in utils/math?

OmarEmaraDev:
Yes. Add the same function to utils/math, but as a normal python function, without the Cython stuff.

OmarEmaraDev (approved):
Looks good. Thanks!

Next time. Setup your editor to add new lines at the end of files.

OmarEmaraDev:
Can you update the CHANGELOG as well? Mentioned in the Changed section that we now slerp for Quaternion mixing.

harisreedhar:
Updated change log

PR #1557, opened 2020-10-11
Fix triangulateMesh method of Mesh.

3DSinghVFX:
I have fixed the *triangulateMesh* method of Mesh for *materialIndices*.


OmarEmaraDev (approved):
Thanks!

PR #1574, opened 2020-10-28
Added isUsed property on Shade Object Smooth node

harisreedhar:
This addition can help to avoid unwanted calculation on every execution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 - *Mix Quaternions* node now uses slerp for mixing
 - Vectorized *Convert Angle* node
 - Vectorized *Vector Dot Product* node
+- Socket enable/disable option for *Shade Object Smooth* node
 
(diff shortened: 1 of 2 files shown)


OmarEmaraDev:
How is this supposed to avoid unwanted calculation on every execution?

harisreedhar:
> How is this supposed to avoid unwanted calculation on every execution?



> How is this supposed to avoid unwanted calculation on every execution?

This option allows users to enable and disable socket. Thus node is only evaluated when socket is enabled. For shade smooth i think it is useful.
smoothShade

OmarEmaraDev:
Sounds reasonable. Can you update the CHANGELOG?

harisreedhar:
> Sounds reasonable. Can you update the CHANGELOG?

Updated

OmarEmaraDev (approved):
Thanks!

PR #1594, opened 2020-11-25
Fix Grease Pencil Nodes for API changes.

3DSinghVFX:
I have fixed the *Grease Pencil Nodes* for API changes of Blender 2.91.


OmarEmaraDev (approved):
Thanks for the prompt fix!

PR #1617, opened 2021-01-06
improve Spiral in Distribute Matrix Node

kdsin23:
final effect 

spiral improve

### DIFF
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -79,7 +79,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         description = "Increase to have a more accurate evaluation if the type is set to Uniform",
         update = propertyChanged)
     
-    __annotations__["centerAxis"] =  BoolProperty(name = "Center Axis",
+    __annotations__["centerSpiral"] =  BoolProperty(name = "Center Spiral",
         description = "Center the spiral along Z axis",
         default = False, update = propertyChanged)
 
@@ -116,9 +116,9 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
-            self.newInput("Float", "Altitude", "altitude", value = 0)
-            self.newInput("Interpolation", "Interpolation_1", "interpolation_1", defaultDrawType = "PROPERTY_ONLY", hide = True)
-            self.newInput("Interpolation", "Interpolation_2", "interpolation_2", defaultDrawType = "PROPERTY_ONLY", hide = True)
+            self.newInput("Float", "Spiral Height", "spiralHeight", value = 0)
+            self.newInput("Interpolation", "Spiral Interpolation", "spiralInterpolation", defaultDrawType = "PROPERTY_ONLY", hide = True)
+            self.newInput("Interpolation", "Height Interpolation", "heightInterpolation", defaultDrawType = "PROPERTY_ONLY", hide = True)
         elif self.mode == "SPLINE":
             self.newInput("Spline", "Spline", "spline", defaultDrawType = "PROPERTY_ONLY")
             if self.splineDistributionMethod == "STEP":
@@ -147,7 +147,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         if self.mode == "SPLINE":
             col.prop(self, "splineDistributionMethod", text = "")
         if self.mode == "SPIRAL":
-            layout.prop(self, "centerAxis")
+            layout.prop(self, "centerSpiral", toggle = True)
 
     def drawAdvanced(self, layout):
         if self.mode == "CIRCLE":
@@ -175,7 +175,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             elif self.meshMode == "POLYGONS":
                 yield "matrices = self.execute_Polygons(mesh)"
         elif self.mode == "SPIRAL":
-            yield "matrices = self.execute_Spiral(amount, startRadius, endRadius, startSize, endSize, startAngle, endAngel, altitude, interpolation_1, interpolation_2)"
+            yield "matrices = self.execute_Spiral(amount, startRadius, endRadius, startSize, endSize, startAngle, endAngel, spiralHeight, spiralInterpolation, heightInterpolation)"
         elif self.mode == "SPLINE":
             if self.splineDistributionMethod == "STEP":
                 yield "matrices = self.execute_SplineStep(spline, step, start, end)"
@@ -273,7 +273,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
 
     def execute_Spiral(self, Py_ssize_t _amount, float startRadius, float endRadius,
                              float startSize, float endSize, float startAngle, float endAngle,
-                             float altitude, Interpolation interpolation_1, Interpolation interpolation_2):
+                             float spiralHeight, Interpolation radiusInterpolation, Interpolation heightInterpolation):
         cdef Py_ssize_t i
         cdef Vector3 position
         cdef float iCos, iSin, stepCos, stepSin, f, size
@@ -287,18 +287,15 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         stepCos = cos(angleStep)
         stepSin = sin(angleStep)
         
-        Zoffset = altitude / 2 * int(self.centerAxis)
-
         for i in range(amount):
             f = <float>i * factor
 
             size = f * (endSize - startSize) + startSize
-            radius = (endRadius - startRadius) * interpolation_1(max(min(f, 1.0), 0.0)) + startRadius
-            Zstep = altitude * interpolation_2(max(min(f, 1.0), 0.0))
+            radius = radius = (endRadius - startRadius) * radiusInterpolation.evaluate(f) + startRadius
 
             position.x = iCos * radius
             position.y = iSin * radius
-            position.z = Zstep - Zoffset
+            position.z = spiralHeight * heightInterpolation.evaluate(f) - spiralHeight / 2 * self.centerSpiral
 
             setTranslationMatrix(matrices.data + i, &position)
             setMatrixCustomZRotation(matrices.data + i, iCos, iSin)
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -5,7 +5,7 @@ from libc.math cimport sin, cos
 from ... utils.limits cimport INT_MAX
 from ... events import propertyChanged
 from ... base_types import AnimationNode
-from ... data_structures cimport Mesh, Matrix4x4List, Vector3DList, Spline
+from ... data_structures cimport Mesh, Matrix4x4List, Vector3DList, Spline, Interpolation
 from ... algorithms.rotations.rotation_and_direction cimport directionToMatrix_LowLevel
 from ... math cimport (Matrix4, Vector3, setTranslationMatrix,
     setMatrixTranslation, setRotationZMatrix, toVector3, scaleMatrix3x3Part)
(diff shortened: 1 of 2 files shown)


animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -78,6 +78,10 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
     __annotations__["splineResolution"] = IntProperty(name = "Spline Resolution", min = 2, default = 20,
         description = "Increase to have a more accurate evaluation if the type is set to Uniform",
         update = propertyChanged)
+    
+    __annotations__["centerAxis"] =  BoolProperty(name = "Center Axis",
```
OmarEmaraDev, line 82: Maybe call this Center Spiral.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -112,6 +116,9 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
+            self.newInput("Float", "Altitude", "altitude", value = 0)
```
OmarEmaraDev, line 119: Let us call this Height to be consistent with other modes.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -139,6 +146,8 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             row.prop(self, "centerAlongZ", text = "Z", toggle = True)
         if self.mode == "SPLINE":
             col.prop(self, "splineDistributionMethod", text = "")
+        if self.mode == "SPIRAL":
+            layout.prop(self, "centerAxis")
```
OmarEmaraDev, line 150: Pass `toggle = True` here.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -276,16 +286,19 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
         stepSin = sin(angleStep)
+        
+        Zoffset = altitude / 2 * int(self.centerAxis)
 
         for i in range(amount):
             f = <float>i * factor
 
             size = f * (endSize - startSize) + startSize
-            radius = f * (endRadius - startRadius) + startRadius
+            radius = (endRadius - startRadius) * interpolation_1(max(min(f, 1.0), 0.0)) + startRadius
```
OmarEmaraDev, line 296: The logic doesn't seem to be correct here. If the end and start radius were equal, the expression would evaluate to zero, and you will end up with just `startRadius`, which I suppose is not what you intended. Can you explain the intended logic?

Also `f`  is guaranteed to be in the `[0, 1]` range, so you don't need the min and max.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -276,16 +286,19 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            Zstep = altitude * interpolation_2(max(min(f, 1.0), 0.0))
```
OmarEmaraDev, line 297: Min and max are not needed here as well. You can just inline this variable as well, no need to store this value in a variable.

Interpolations in Cython should be evaluated using the `evaluate` method instead.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -276,16 +286,19 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         iSin = sin(startAngle)
         stepCos = cos(angleStep)
         stepSin = sin(angleStep)
+        
+        Zoffset = altitude / 2 * int(self.centerAxis)
```
OmarEmaraDev, line 290: We don't use capital letters as the start of variable names. Also, put parentheses for clarity here:
```
zOffset = (height / 2.0) * self.centerSpiral
```

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -112,6 +116,9 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
+            self.newInput("Float", "Altitude", "altitude", value = 0)
+            self.newInput("Interpolation", "Interpolation_1", "interpolation_1", defaultDrawType = "PROPERTY_ONLY", hide = True)
+            self.newInput("Interpolation", "Interpolation_2", "interpolation_2", defaultDrawType = "PROPERTY_ONLY", hide = True)
```
OmarEmaraDev, line 121: You should give better names to those.

OmarEmaraDev (changes requested):

OmarEmaraDev:
@Gerstmann-Bradley You can provide feedback here.

animation_nodes/nodes/matrix/distribute_matrices.pyx
kdsin23: acturally, at the begining i  used the "Height", but i found it overlay the grid mode Height, and can not set default value to 0, i don't know why.

animation_nodes/nodes/matrix/distribute_matrices.pyx
kdsin23: i just wanted to modify the interval between each turn in XY plane, so i mutiply the` interpolation_1(f)` , If the end and start radius were equal, means interal equal zero, in this case interpolation lose effect.

animation_nodes/nodes/matrix/distribute_matrices.pyx
kdsin23: is like this `interpolation_XY` and `interpolation_Z` ?

animation_nodes/nodes/matrix/distribute_matrices.pyx
OmarEmaraDev: In this case, you can just change the identifier to `spiralHeight` (Third argument).

animation_nodes/nodes/matrix/distribute_matrices.pyx
OmarEmaraDev: `radiusInterpolation` and `zInterpolation`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
OmarEmaraDev: I am not sure about that behavior. When I tried it the first time, I expected something different and it seemed to me like it didn't work. Users might feel the same way. Maybe the interpolation should be over the whole expression after the addition.
What do you think @Gerstmann-Bradley?

Gerstmann-Bradley: I need to think more about this, will get back to this later.

Gerstmann-Bradley: I have tested build, since I don't have a build by Omar upon this issue, I can't get a comparison about which is better.
I can only tell this curvature interpolation looks normal to me, while I think Omar's logic is right.
If no one is pointing out, I probably already think we are interpolate over the whole expression.
I am satisfied with the current result because I can't tell the difference, but I think I agree with Omar unless I find issues otherwise.

animation_nodes/nodes/matrix/distribute_matrices.pyx
Gerstmann-Bradley: Hi, I agree with Omar's suggestions, but since we are naming "Height" as "Height" not "z", perhaps it makes more sense and consistent to name it "heightInterpolation" instead of "zInterpolation". This is a really minor issue though, so I don't mind which name to use eventually, because I understand "heightInterpolation" maybe a little bit too long?

animation_nodes/nodes/matrix/distribute_matrices.pyx
kdsin23: @OmarEmaraDev you means code like this `radius = ((endRadius - startRadius) + startRadius) * radiusInterpolation.evaluate(f)`, is it right? but i tried this, the startRadius lose effect.

spiral 3

Gerstmann-Bradley: I have just come up with a way to build spiral using AN from scratch. I'm getting the same answer as @kdsin23. I am not sure the logic behind his coding, but the result matches my expectation. I have no idea now, I will leave the discussion to you two.
example

OmarEmaraDev: I idea was to let the base case be a constant interpolation at value 1. But I guess this one might be confusing as well as evident by this discussion. A base linear interpolation might be more familiar to users. So just leave it like you had originally for now. We will make sure to document those cases properly.

animation_nodes/nodes/matrix/distribute_matrices.pyx
OmarEmaraDev: Sure, doesn't matter that much.

animation_nodes/nodes/matrix/distribute_matrices.pyx
kdsin23: ok, i will do that

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -112,6 +116,9 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             self.newInput("Float", "End Size", "endSize", value = 0.5, minValue = 0)
             self.newInput("Float", "Start Angle", "startAngle", value = 0)
             self.newInput("Float", "End Angle", "endAngel", value = 6 * PI)
+            self.newInput("Float", "Spiral Height", "spiralHeight", value = 0)
```
OmarEmaraDev, line 119: Keep the name as `Height`. Only the identifier matter here.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -112,6 +116,9 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            self.newInput("Interpolation", "Spiral Interpolation", "spiralInterpolation", defaultDrawType = "PROPERTY_ONLY", hide = True)
```
OmarEmaraDev, line 120: `Spiral Interpolation` is not very informative either. I suggest `Radius Interpolation`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -166,7 +175,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             elif self.meshMode == "POLYGONS":
                 yield "matrices = self.execute_Polygons(mesh)"
         elif self.mode == "SPIRAL":
-            yield "matrices = self.execute_Spiral(amount, startRadius, endRadius, startSize, endSize, startAngle, endAngel)"
+            yield "matrices = self.execute_Spiral(amount, startRadius, endRadius, startSize, endSize, startAngle, endAngel, spiralHeight, spiralInterpolation, heightInterpolation)"
```
OmarEmaraDev, line 178: Break this line.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -276,16 +286,16 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
         stepCos = cos(angleStep)
         stepSin = sin(angleStep)
-
+        
         for i in range(amount):
             f = <float>i * factor
 
             size = f * (endSize - startSize) + startSize
-            radius = f * (endRadius - startRadius) + startRadius
+            radius = radius = (endRadius - startRadius) * radiusInterpolation.evaluate(f) + startRadius
```
OmarEmaraDev, line 294: You got two `radius = ` here.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -276,16 +286,16 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
 
             position.x = iCos * radius
             position.y = iSin * radius
-            position.z = 0
+            position.z = spiralHeight * heightInterpolation.evaluate(f) - spiralHeight / 2 * self.centerSpiral
```
OmarEmaraDev, line 298: Sorry for the miscommunication. I meant to inline only the step, the offset should be left as is outside the loop. With parentheses for clarity. 
```
(spiralHeight / 2) * self.centerSpiral
```

OmarEmaraDev (changes requested):

OmarEmaraDev (commented):
Make sure to update the CHANGELOG as well.

OmarEmaraDev (approved):
Thanks!

PR #1622, opened 2021-01-09
Added "Center Linear " option to linear mode in Distribute Matrix node

kdsin23:
center linear

### DIFF
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -46,6 +46,8 @@ searchItems = {
     "Distribute Spline" : "SPLINE",
 }
 
+directionAxisItems = [(axis, axis, "") for axis in ("X", "Y", "Z")]
+
 class DistributeMatricesNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_DistributeMatricesNode"
     bl_label = "Distribute Matrices"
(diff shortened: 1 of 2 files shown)


Gerstmann-Bradley:
If we were to improve the function of linear mode, I wonder if it's possible to decide which direction this linear will go towards. like X, Y, Z.  I mostly use grid mode instead of linear mode because of this reason.
And Center Linear option is great. Thanks!

OmarEmaraDev:
Feel free to implement @Gerstmann-Bradley's suggestions.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -137,6 +137,8 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         col.prop(self, "mode", text = "")
         if self.mode in ("LINEAR", "GRID"):
             col.prop(self, "distanceMode", text = "")
+            if self.mode == "LINEAR":
+                layout.prop(self, "centerAlongX", text = "Center Linear", toggle = True)
```
OmarEmaraDev, line 141: If we will be sharing those properties between modes, make the property description more general above.

OmarEmaraDev (changes requested):

animation_nodes/nodes/matrix/distribute_matrices.pyx
kdsin23: linear mode

is this OK?

Gerstmann-Bradley: looks good to me at least. Thanks for development.

OmarEmaraDev (approved):
Thanks!

PR #1633, opened 2021-01-14
Fix bl_idname of Evaluate MIDI Track node.

3DSinghVFX:
I have fixed the `bl_idname` of the **Evaluate MIDI Track** node.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 ### Fixed
 
-- Fixed `bl_idname` of *Evaluate MIDI Track* node.
+
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@
 
 ### Fixed
 
-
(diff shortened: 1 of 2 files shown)


OmarEmaraDev:
What was wrong with it?

3DSinghVFX:
> What was wrong with it?

Omar, the `bl_idname` was wrong in the node file.

OmarEmaraDev:
Wrong how I mean.

3DSinghVFX:
> Wrong how I mean.

In all MIDI nodes, the `bl_idname` convension is `an_...MIDI...` but in this case it is `an_...Midi...`.

OmarEmaraDev:
Right, thanks! But you shouldn't add that to the CHANGELOG, we are still in the same subversion.

3DSinghVFX:
> Right, thanks! But you shouldn't add that to the CHANGELOG, we are still in the same subversion.

Okay. It removed.

OmarEmaraDev (approved):
Thanks!

PR #1641, opened 2021-01-16
Matrix socket usability improvement

harisreedhar:
Added ``Compose Matrix Node`` invoke button on Matrix socket.

matrix_socket_improvement

### DIFF
--- a/animation_nodes/sockets/matrix.py
+++ b/animation_nodes/sockets/matrix.py
@@ -1,7 +1,9 @@
 import bpy
+from bpy.props import *
 from mathutils import Matrix
 from .. data_structures import Matrix4x4List
 from .. base_types import AnimationNodeSocket, CListSocket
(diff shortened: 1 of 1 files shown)


OmarEmaraDev:
This is already possible with the W shortcut. Or this something different?


https://user-images.githubusercontent.com/25300994/104807977-9abde780-57eb-11eb-962b-4f82bec900f2.mp4

harisreedhar:
> This is already possible with the W shortcut. Or this something different?
> 20210116-111513.mp4

Yeah it is similar, but for multiple Matrix inputs it saves two clicks.

OmarEmaraDev:
I don't think it is worth it in this case. I would rather keep the socket UI as clean as possible.

harisreedhar:
@OmarEmaraDev Ok

PR #1644, opened 2021-01-18
Vectorize Delay Time node

kdsin23:


animation_nodes/nodes/animation/delay_time.py
```diff
@@ -20,4 +26,7 @@ def drawLabel(self):
         else: return "Delay Time"
 
     def getExecutionCode(self, required):
-        return "outTime = time - delay"
+        if self.useList:
+            return "outTimes = DoubleList.fromValues((time - delay) for delay in delays)"
```
OmarEmaraDev, line 30: Might as well make the input time vectorized and both inputs virtual lists. You should implement that in Cython. See any virtual node for reference.

animation_nodes/nodes/animation/delay_time.py
```diff
@@ -1,15 +1,21 @@
...
     dynamicLabelType = "HIDDEN_ONLY"
 
+    useList: VectorizedSocket.newProperty()
+
     def create(self):
         self.newInput("Float", "Time", "time")
-        self.newInput("Float", "Delay", "delay", value = 10)
-        self.newOutput("Float", "Time", "outTime")
+        self.newInput(VectorizedSocket("Float", "useList",
+            ("Delay", "delay",dict(value = 10, minValue = 0)),
```
OmarEmaraDev, line 15: No need for the minimum value here. Also, space between `,` and `dict`.

OmarEmaraDev (changes requested):

animation_nodes/nodes/animation/delay_time.py
kdsin23: Ok,I will do that in Cython.

kdsin23: do you think where `c_utils` file should be placed is more better, i templary put it in `animation`, if not better, i will change.

animation_nodes/nodes/animation/delay_time.py
```diff
@@ -20,4 +31,31 @@ def drawLabel(self):
...
+        amount = VirtualDoubleList.getMaxRealLength(virtualA, virtualB)
+
+        return delayTime_Multiple(virtualA, virtualB, amount)
+
+    def executeList_Delays(self, time, delays):
+        times = DoubleList.fromValue(time)
+        virtualA, virtualB = VirtualDoubleList.createMultiple((times, 0), (delays, 0))
+        amount = VirtualDoubleList.getMaxRealLength(virtualA, virtualB)
+
+        return delayTime_Multiple(virtualA, virtualB, amount)
```
OmarEmaraDev, line 61: All of those methods are identical. You can just execute one of them if if any of the inputs is a list. Also, avoid using the names like A and B. We know exactly what those lists contain and they should be named accordingly.

animation_nodes/nodes/animation/c_utils.pyx
```diff
@@ -0,0 +1,12 @@
+from ... data_structures cimport (
+    LongList,
```
OmarEmaraDev, line 2: `LongList` is no needed.

animation_nodes/nodes/animation/c_utils.pyx
```diff
@@ -0,0 +1,12 @@
...
+    DoubleList,
+    VirtualDoubleList
+)
+
+def delayTime_Multiple(VirtualDoubleList a, VirtualDoubleList b, long amount):
+    cdef DoubleList result = DoubleList(length = amount)
+    cdef Py_ssize_t i
+    for i in range(amount):
+        result.data[i] = a.get(i) - b.get(i)
+    return result
```
OmarEmaraDev, line 12: Setup your editor to add a new line at the end of files.

OmarEmaraDev (changes requested):

kdsin23:
BTW, Now am i in a right workfow?

OmarEmaraDev (approved):
Looks good. Thanks!

OmarEmaraDev:
@kdsin23 Seems like you are using the right workflow. Maybe just name your branches more meaningfully.

PR #1650, opened 2021-01-21
Correct mesh components' order and names.

3DSinghVFX:
I have corrected the mesh components' order and names.
Old_New


OmarEmaraDev (approved):
Thanks!

PR #1651, opened 2021-01-22
Vectorize animate data node

### LINKED ISSUE #1638: Proposal: vectorize "delay time" & "animate" nodes
The attached is an example of what I mean. **I mostly want vectorization in "delay" in delay time, & "time" input in animate node**.
(Indeed, more vectorizations will be welcome but I don't think they will be too useful; so if these are too heavy to implement then I won't ask for it.)
(Delay in delay time is also tricky, because technically we can use math node instead.)

I usually use this method to create a sequential event, examples include "flip book pages", etc.
Here I am not talking about "delay action" because we can't decompose action,
while numerical control in "animate float", "animate vector", etc. are very helpful as they can do math in animation nodes for certain purposes. 

Other insights will be welcome.
Example

kdsin23:
Resolves #1638.

### DIFF
--- a/animation_nodes/nodes/generic/animate_data.py
+++ b/animation_nodes/nodes/generic/animate_data.py
@@ -1,8 +1,29 @@
 import bpy
 from bpy.props import *
 from . mix_data import getMixCode
-from ... base_types import AnimationNode
+from ... utils.math import mixEulers
 from ... events import executionCodeChanged
+from ... base_types import AnimationNode, VectorizedSocket
+
+from . c_utils import (
+    mixColorLists,
+    mixEulerLists,
+    mixDoubleLists,
+    mixVectorLists,
+    mixQuaternionLists,
+    executeTimeList,
+    calculateInfluenceList,
+)
+from ... data_structures import (
+    Color,
+    Matrix4x4List,
+    VirtualColorList,
+    VirtualEulerList,
+    VirtualDoubleList,
+    VirtualVector3DList,
+    VirtualMatrix4x4List,
+    VirtualQuaternionList,
+)
 
 class AnimateDataNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_AnimateDataNode"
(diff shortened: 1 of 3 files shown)


Gerstmann-Bradley:
I think it's fine not to vectorize interpolation (unless it can be easily done).
I can barely come up a case in which people really need to vectorize interpolation,
More importantly, We can't directly generate random interpolations, as we can do with float, vector, etc. Even we vectorize interpolation input, it may not be very straightforward to get a list of various interpolations to use. Thus this vectorization is less important than other input.
At the end, thanks a lot for your contributions!

OmarEmaraDev:
What is the issue with vectorizing the interpolation socket?

animation_nodes/nodes/generic/animate_data.py
```diff
@@ -20,22 +41,81 @@ class AnimateDataNode(bpy.types.Node, AnimationNode):
...
+            yield "results, outTimes = self.executeDataList({})".format(args)
+
+        else:
+            yield "finalDuration = max(duration, 0.0001)"
+            yield "influence = max(min(time / finalDuration, 1.0), 0.0)"
+            yield "influence = interpolation(influence)"
+            yield getMixCode(self.dataType, "start", "end", "influence", "result")
+            yield "outTime = time - finalDuration"
+
+    def executeDataList(self, time, start, end, interpolation, duration):
```
OmarEmaraDev, line 81: You can probably just add that to the execution code directly. This way, you can dynamically select the virtual list type and the needed execution function directly.

animation_nodes/nodes/generic/animate_data.py
```diff
@@ -20,22 +41,81 @@ class AnimateDataNode(bpy.types.Node, AnimationNode):
...
 
     def getExecutionCode(self, required):
-        yield "finalDuration = max(duration, 0.0001)"
-        yield "influence = max(min(time / finalDuration, 1.0), 0.0)"
-        yield "influence = interpolation(influence)"
-        yield getMixCode(self.dataType, "start", "end", "influence", "result")
-        yield "outTime = time - finalDuration"
+        if any([self.useListA, self.useListB, self.useListC, self.useListD]):
+            args = ", ".join(socket.identifier for socket in self.inputs)
+            yield "results, outTimes = self.executeDataList({})".format(args)
```
OmarEmaraDev, line 72: Use f-strings instead.

OmarEmaraDev (changes requested):

kdsin23:
> What is the issue with vectorizing the interpolation socket?

interpolationList is not allowed in `cython`? why?
i coded like this:
``def testfunction( InterpolationList interpolations)`
is not allowed.

animation_nodes/nodes/generic/animate_data.py
kdsin23: sorry, Can U make more detailed about this? how to wirite this ? Thank you!

kdsin23:
> I think it's fine not to vectorize interpolation (unless it can be easily done).
> I can barely come up a case in which people really need to vectorize interpolation,
> More importantly, We can't directly generate random interpolations, as we can do with float, vector, etc. Even we vectorize interpolation input, it may not be very straightforward to get a list of various interpolations to use. Thus this vectorization is less important than other input.
> At the end, thanks a lot for your contributions!

Thank you for your feeding back.

animation_nodes/nodes/generic/animate_data.py
kdsin23: maybe i guess you meant like this:
 ```
        if any([self.useListA, self.useListB, self.useListC, self.useListD]):
            args = ", ".join(socket.identifier for socket in self.inputs)
            yield f"time, start, end, interpolation, duration = {args}"
            yield f"starts, ends = AN.data_structures.Virtual{self.dataType}List.createMultiple((start, 0), (end, 0))"
            yield "amount = AN.data_structures.VirtualDoubleList.getMaxRealLength(times, starts, ends, durations)"
            yield "factors = AN.nodes.generic.c_utils.calculateInfluenceList(times, interpolation, durations, amount)"
            yield "_factors = AN.data_structures.VirtualDoubleList.create(factors, 0)"
            yield f"outTimes = AN.nodes.generic.c_utils.executeTimeList(times, durations, amount)"
            yield "results = AN.nodes.generic.c_utils.mix{self.dataType}Lists(starts, ends, _factors, amount)"
```

Comment (OmarEmaraDev) [reply animation_nodes/nodes/generic/animate_data.py:81]: Yes. Something like that. Though this probably will not work because `Virtual{self.dataType}List` will not be correct for vector lists and maybe others. So, a dictionary mapping the data type to the appropriate virtual list can be created outside the node class and used to retrieve the name of the virtual list.

Also, to avoid this line:
```
args = ", ".join(socket.identifier for socket in self.inputs)
```
You can just use the same identifier for both base and list inputs. So instead of `("Time", "time"), ("Times", "times")`, you do `("Time", "time"), ("Times", "time")`. Notice the name stays the same, but the identifier is made identical.

**Review by OmarEmaraDev — CHANGES_REQUESTED** [review]

File: animation_nodes/nodes/generic/animate_data.py (hunk 1 above)
Comment (kdsin23) [reply animation_nodes/nodes/generic/animate_data.py]: sorry, i don't know how to use a dictionary mapping the data type, can U give me example? Thank you!

**Comment by kdsin23** [comment]
> > > What is the issue with vectorizing the interpolation socket?
> > 
> > 
> > interpolationList is not allowed in `cython`? why?
> > i coded like this:
> > ``def testfunction( InterpolationList interpolations)`
> > is not allowed.
> 
> There is no InterpolationList structure. You can use list directly `list interpolations`. If you want a virtual list from that, you can also use `VirtualPyList`. But in this node, I would just leave the interpolation a constant anyway.

Thank you! do you think it is need to vectorize interpolation socket?

File: animation_nodes/nodes/generic/animate_data.py (hunk 1 above)
Comment (OmarEmaraDev) [reply animation_nodes/nodes/generic/animate_data.py]: @kdsin23 Outside the class, you should define the following dictionary:

```python
dataTypeToVirtualListMapping = {
    "Float" : "VirtualDoubleList",
    "Vector" : "VirtualVector3DList",
    ...
}
```

Then, in the execution code you do:

```
f"AN.data_structures.{dataTypeToVirtualListMapping[self.dataType]}.createMultiple((start, 0), (end, 0))"
```

animation_nodes/nodes/generic/animate_data.py
```diff
@@ -20,22 +37,50 @@ class AnimateDataNode(bpy.types.Node, AnimationNode):
 
     dataType: StringProperty(default = "Float", update = AnimationNode.refresh)
 
+    useListA: VectorizedSocket.newProperty()
+    useListB: VectorizedSocket.newProperty()
+    useListC: VectorizedSocket.newProperty()
+    useListD: VectorizedSocket.newProperty()
```
OmarEmaraDev, line 43: Use more meaningful names here.

animation_nodes/nodes/generic/c_utils.pyx
```diff
@@ -5,11 +5,12 @@ from ... data_structures cimport (
     DoubleList,
     Vector3DList,
     QuaternionList,
+    Interpolation,
     VirtualColorList,
     VirtualEulerList,
     VirtualDoubleList,
     VirtualVector3DList,
-    VirtualQuaternionList,
+    VirtualQuaternionList,
```
OmarEmaraDev, line 13: Remove that extra space at the end.

OmarEmaraDev (changes requested):

OmarEmaraDev (approved):
Thanks!

PR #1655, opened 2021-02-01
Add Start Time to Animate Data nodes.

3DSinghVFX:
I have added the *Start Time* input to **Animate Data** nodes that will allow to set the start time for animation.


OmarEmaraDev:
While at it, can you change the order of the outputs. Also, maybe rename the input Start and End  into Start Value and End Value.

3DSinghVFX:
> While at it, can you change the order of the outputs. Also, maybe rename the input Start and End into Start Value and End Value.

Okay.

OmarEmaraDev (approved):
Thanks!

Issue #1658, opened 2021-02-11
Proposal: be able to set plane for "distribute circle"


Gerstmann-Bradley:
The circle is generated in XY plane. it perhaps will be more convenient if we can set it to other like XZ/YZ plane.
This maybe a minor thing since you can accomplish with two additional nodes, but as the time goes I think it would be more convenient to set it at the first place without additional nodes.
What do you think?

OmarEmaraDev:
I think that's fine. I will leave this as a good first contribution.

PR #1659, opened 2021-02-11
Add Plane-Axis for Circle mode of Distribute Matrices node

3DSinghVFX:
I have added the Plane-Axis option in the *Circle* mode of the **Distribute Matrices** node.

Resolves: #1658

### DIFF
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -264,22 +264,39 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         stepCos = cos(angleStep)
         stepSin = sin(angleStep)
 
-        for i in range(amount):
-            if self.planeAxis == "XY":
+        if self.planeAxis == "XY":
+            for i in range(amount):
                 vector.x = iCos * radius
                 vector.y = iSin * radius
                 vector.z = 0
-            elif self.planeAxis == "YZ":
+
+                setTranslationMatrix(matrices.data + i, &vector)
+                setMatrixCustomZRotation(matrices.data + i, iCos, iSin)
+
+                rotateStep(&iCos, &iSin, stepCos, stepSin)
+
+            return matrices
+
+        elif self.planeAxis == "YZ":
+            for i in range(amount):
                 vector.x = 0
-                vector.y = iSin * radius
-                vector.z = iCos * radius
-            else:
-                vector.x = iCos * radius
-                vector.y = 0
+                vector.y = iCos * radius
                 vector.z = iSin * radius
 
+                setTranslationMatrix(matrices.data + i, &vector)
+                setMatrixCustomXRotation(matrices.data + i, iCos, iSin)
+
+                rotateStep(&iCos, &iSin, stepCos, stepSin)
+
+            return matrices
+
+        for i in range(amount):
+            vector.x = iSin * radius
+            vector.y = 0
+            vector.z = iCos * radius
+
             setTranslationMatrix(matrices.data + i, &vector)
-            setMatrixCustomZRotation(matrices.data + i, iCos, iSin)
+            setMatrixCustomYRotation(matrices.data + i, iCos, iSin)
 
             rotateStep(&iCos, &iSin, stepCos, stepSin)
 
@@ -370,6 +387,14 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
 
 cdef int limitAmount(n):
     return max(min(n, INT_MAX), 0)
+
+cdef inline void setMatrixCustomXRotation(Matrix4* m, double iCos, double iSin):
+    m.a22 = m.a33 = iCos
+    m.a23, m.a32 = -iSin, iSin
+
+cdef inline void setMatrixCustomYRotation(Matrix4* m, double iCos, double iSin):
+    m.a11 = m.a33 = iCos
+    m.a13, m.a31 = iSin, -iSin
 
 cdef inline void setMatrixCustomZRotation(Matrix4* m, double iCos, double iSin):
     m.a11 = m.a22 = iCos
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -277,7 +277,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
 
             return matrices
 
-        elif self.planeAxis == "YZ":
+        if self.planeAxis == "YZ":
             for i in range(amount):
                 vector.x = 0
                 vector.y = iCos * radius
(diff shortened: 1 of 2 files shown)


animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -260,9 +265,19 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+                vector.y = iSin * radius
+                vector.z = 0
+            elif self.planeAxis == "YZ":
+                vector.x = 0
+                vector.y = iSin * radius
+                vector.z = iCos * radius
+            else:
+                vector.x = iCos * radius
+                vector.y = 0
+                vector.z = iSin * radius
```
OmarEmaraDev, line 279: Move the condition outside of the loop. This is too much overhead inside the loop.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -260,9 +265,19 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+
             setTranslationMatrix(matrices.data + i, &vector)
             setMatrixCustomZRotation(matrices.data + i, iCos, iSin)
```
OmarEmaraDev, line 282: Should this change as well?

OmarEmaraDev (changes requested):

animation_nodes/nodes/matrix/distribute_matrices.pyx
3DSinghVFX: Okay.

animation_nodes/nodes/matrix/distribute_matrices.pyx
3DSinghVFX: Omar, If I define `setMatrixCustom...` according to the [right-hand rule](https://en.wikipedia.org/wiki/Rotation_matrix) then they do not output correct rotation,
```
cdef inline void setMatrixCustomXRotation(Matrix4* m, double iCos, double iSin):
    m.a22 = m.a33 = iCos
    m.a23, m.a32 = -iSin, iSin

cdef inline void setMatrixCustomYRotation(Matrix4* m, double iCos, double iSin):
    m.a11 = m.a33 = iCos
    m.a13, m.a31 = iSin, -iSin
```
Screenshot from 2021-02-12 00-05-19



However, if I change the -ve sign then they output correct rotation,
```
cdef inline void setMatrixCustomXRotation(Matrix4* m, double iCos, double iSin):
    m.a22 = m.a33 = iCos
    m.a23, m.a32 = -iSin, iSin

cdef inline void setMatrixCustomYRotation(Matrix4* m, double iCos, double iSin):
    m.a11 = m.a33 = iCos
    m.a13, m.a31 = -iSin, iSin
```
Screenshot from 2021-02-12 00-06-32

Is there something that I'm doing wrong?

OmarEmaraDev: @3DSinghVFX Those functions don't really have anything to do with rotations matrices or the right hand rule. You construct the rotation of the matrix by making three decisions:

- What should the local x axis of the matrix be aligned with? In the XY case, we choose to align with the normal to the circle. The normal to the circle is obviously `V(cos, sin, 0)` by looking at the unit circle.
- What should the local y axis of the matrix be aligned with? In the XY case, we choose to align with the tangent to the circle. The tangent to the circle is perpendicular to the normal, which is equivalent to a `pi/2` degree rotation, which we can compute through trigonometric identifies as follows, `V(cos + pi/2, sin + pi/2, 0)` -> `V(-sin, cos, 0)`.
- What should the local z axis of the matrix be aligned with? In the XY case, we choose to align with `V(0, 0, 1)`.

To actually construct the matrix from all of those vectors, you assign them to the 3x3 columns of the matrix. The x-aligned vector is set to the first column, the y-aligned vector is set to the second column, and the z-aligned vector is set to the their column. In the XY case, it looks like this:

```
cos  -sin 0
sin  cos  0
0    0    1
```

The other options can be deduced similarly.

3DSinghVFX: Thanks for the explanation. I have aligned the axis in the cycle: `Y` in `XY`, `Z` in `YZ`, and `X` in `ZX`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -259,12 +264,39 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+                vector.z = 0
+
+                setTranslationMatrix(matrices.data + i, &vector)
+                setMatrixCustomZRotation(matrices.data + i, iCos, iSin)
+
+                rotateStep(&iCos, &iSin, stepCos, stepSin)
+
+            return matrices
+
+        elif self.planeAxis == "YZ":
```
OmarEmaraDev, line 280: If you are going to use a guard condition, this should be an `if`.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -259,12 +264,39 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+                vector.z = iSin * radius
+
+                setTranslationMatrix(matrices.data + i, &vector)
+                setMatrixCustomXRotation(matrices.data + i, iCos, iSin)
+
+                rotateStep(&iCos, &iSin, stepCos, stepSin)
+
+            return matrices
+
         for i in range(amount):
```
OmarEmaraDev, line 293: For the `ZX` case, the matrices seems to start at pi/2, as opposed to the other modes.

OmarEmaraDev (changes requested):

animation_nodes/nodes/matrix/distribute_matrices.pyx
3DSinghVFX: Changed.

animation_nodes/nodes/matrix/distribute_matrices.pyx
3DSinghVFX: Omar, If correct that then it aligns the `Z` axis with tangent instead the `X` axis,
```
for i in range(amount):
            vector.x = iCos * radius
            vector.y = 0
            vector.z = iSin * radius

            setTranslationMatrix(matrices.data + i, &vector)
            setMatrixCustomYRotation(matrices.data + i, iCos, iSin)

            rotateStep(&iCos, &iSin, stepCos, stepSin)

cdef inline void setMatrixCustomYRotation(Matrix4* m, double iCos, double iSin):
    m.a11 = m.a33 = iCos
    m.a13, m.a31 = -iSin, iSin
```

OmarEmaraDev: @3DSinghVFX  Seems to the the result I was expecting, it matches the other two cases. What exactly is the problem?

3DSinghVFX: > @3DSinghVFX Seems to the the result I was expecting, it matches the other two cases. What exactly is the problem?
> 
> 

I'm also getting this result. I was thinking to align the *X* axis with tangents of the circle.

OmarEmaraDev (approved):
Thanks!

PR #1660, opened 2021-02-14
New DataType

OsmanGny:
Can you add Vector4D type.I need to W variable of Vector on some matrix operations


OmarEmaraDev:
Please don't open merge pull requests. If you have a feature request, you can use issues instead.

PR #1665, opened 2021-02-16
Marching Squares node

3DSinghVFX:
I have added the **Marching Squares** node. It generates contours of falloff based on the [marching-squares](https://en.wikipedia.org/wiki/Marching_squares) method on a grid as well as on mesh surface with quad polygons for multiple thresholds. I have also added the linear-interpolation method to smooth the contours.

*Grid*
screen

*Mesh*
screen1

Ter0016

Comp0017


OmarEmaraDev:
Haven't really looked at the code yet, but right off the bat, as far as I know, marching squares can be implemented using a simple bit encoding and indexing technique. This should skip you all those if statements, switch cases, and base conversions. I recommend you use Paul Bourke's marching cubes implementation as a reference. http://paulbourke.net/geometry/polygonise/

OmarEmaraDev:
Can you provide a more detailed example on the threshold input? How does vectorization affect the result? Are the threshold-results concatenated inside the mesh?

Are segments connected or disconnected?

3DSinghVFX:
> Can you provide a more detailed example on the threshold input? How does vectorization affect the result? Are the threshold-results concatenated inside the mesh?

Vectorization of `Thresholds` allows/samples the falloff based on multiple iso-levels.

*Single Threshold*
Screenshot from 2021-02-17 18-57-01

*Two Thresholds*
Screenshot from 2021-02-17 18-56-19


> Are segments connected or disconnected?

Yes, they are connected.

3DSinghVFX:
> Haven't really looked at the code yet, but right off the bat, as far as I know, marching squares can be implemented using a simple bit encoding and indexing technique. This should skip you all those if statements, switch cases, and base conversions. I recommend you use Paul Bourke's marching cubes implementation as a reference. http://paulbourke.net/geometry/polygonise/

Okay, I'll try.

OmarEmaraDev:
Do you need help with this?

3DSinghVFX:
> Do you need help with this?

Actually, I have been busy with the `Custom Attribute` patch. So, I have not fully studied that algorithm yet.
BTW, I'm trying to solve an issue of marching nodes https://github.com/3DSinghVFX/animation_nodes/issues/166

OmarEmaraDev:
Alright. I wouldn't look at this issue now, we are likely to rewrite some parts of the code, so spending time on this is not a good idea.

OmarEmaraDev:
Feel free to reopen when needed.

3DSinghVFX:
> Feel free to reopen when needed.

Okay.

PR #1672, opened 2021-03-05
Custom Attributes Nodes

3DSinghVFX:
I have added *Custom Attribute* nodes: *Get Custom Attribute, Insert Custom Attribute* and *Set Custom Attribute* nodes.
Screenshot from 2021-03-05 21-37-44


OmarEmaraDev:
I am currently in the midst of exams, so it might take me sometime to review this. Some questions first:

- How does Blender treat vertex colors and uv layers? Are they considered custom attributes? Can attributes exist with the same name? Can we unify custom attributes and vertex and uv layers in the Mesh structure?
- Mesh processing is currently ignoring those attributes. Those needs to be handled. See uses for `getMeshProperties`. Can you look into this?

3DSinghVFX:
> I am currently in the midst of exams, so it might take me sometime to review this. Some questions first:

It is okay.

>     * How does Blender treat vertex colors and uv layers? Are they considered custom attributes? Can attributes exist with the same name? Can we unify custom attributes and vertex and uv layers in the Mesh structure?

Vertex Colors are considered as *Custom Attributes* with the same name. BUT there is a mismatch in color if we set with the custom attribute that I do not why. 

>     * Mesh processing is currently ignoring those attributes. Those needs to be handled. See uses for `getMeshProperties`. Can you look into this?

Okay. I'll look into that.

3DSinghVFX:
>     * Mesh processing is currently ignoring those attributes. Those needs to be handled. See uses for `getMeshProperties`. Can you look into this?

Omar, Now the `getMeshProperties` method also handles the *Custom Attributes* https://github.com/JacquesLucke/animation_nodes/pull/1672/commits/7690f02bdc366b4d9fb2a463648da09e53d9de51

Gerstmann-Bradley:
I suppose Omar said what I wanted to say.

I have one question though. Thinking about this scenario:
I have an attribute "X", I multiply this "X" with whatever and output an "X". (Sounds bizarre but this happens in GN)
This means I will have different values of "X" at different time point.
If I were to get the attribute, which X will I get? The most beginning X upon creation? Or the last X at the end of node tree?

Indeed we can name the X after math to be X2 or Y or others in many cases to avoid such a problem.
But there are other cases we can't do so, for example the mesh has "position", we can overwrite "position" in GN to procedurally deform the mesh. In such a case, we will have to name the position as position after processing. In such a case, If I were to get attribute of position, which position will I get? Am I able to avoid this problem?

At the end, thanks for the development. This is really nice.

Gerstmann-Bradley:
Also I think this has been mentioned by Omar's second point (Or maybe I am wrong)
I thought this should be fixed in 7690f02? but it didn't seem so.
Basically the bug is that if you just simple divide mesh, whatever attribute you set can't be seen or used.
IMG1
IMG2

OmarEmaraDev (changes requested):
You probably still need to update other methods like `appendMeshProperties` and `copyMeshProperties`.

I think the way to go forward is to create a struct for the attribute storing its information like the domain, data type, data and so on. I would also try to incorporate built-in attributes like `material_index` and other data layers such as vertex colors and UVs into this system. Data layers can be identified as such and handled separately if needed.

3DSinghVFX:
> You probably still need to update other methods like `appendMeshProperties` and `copyMeshProperties`.
> 
> I think the way to go forward is to create a struct for the attribute storing its information like the domain, data type, data and so on. I would also try to incorporate built-in attributes like `material_index` and other data layers such as vertex colors and UVs into this system. Data layers can be identified as such and handled separately if needed.

That is really nice. Here is `Attribute`struct
```
cdef struct Attribute:
    str name
    str type
    str domain
    str dataType
    data
```
where 

- *name* is the attribute's name
- *type* is `[UVMAP, MATERIAL_INDEX, VERTEX_COLOR, ATTRIBUTE]`
- *domain* is `[POINT/VERTEX, EDGE, POLYGON, CORNER/LOOP]`
- *dataType* is `[FLOAT, FLOAT2/VECTOR2D,..]`
- *data* is a list of floats/vectors/colors...

Now my question is how we can define the type for *data* in the struct, means `FloatList Data or Vector3DList Data, so on`?

3DSinghVFX:
> I suppose Omar said what I wanted to say.
> 
> I have one question though. Thinking about this scenario:
> I have an attribute "X", I multiply this "X" with whatever and output an "X". (Sounds bizarre but this happens in GN)
> This means I will have different values of "X" at different time point.
> If I were to get the attribute, which X will I get? The most beginning X upon creation? Or the last X at the end of node tree?

I'm aware of this problem. In this case, it will give the last X attribute of the node-chain. But in AN, you can't use the same name.

> Indeed we can name the X after math to be X2 or Y or others in many cases to avoid such a problem.
> But there are other cases we can't do so, for example the mesh has "position", we can overwrite "position" in GN to procedurally deform the mesh. In such a case, we will have to name the position as position after processing. In such a case, If I were to get attribute of position, which position will I get? Am I able to avoid this problem?
> 
> At the end, thanks for the development. This is really nice.

The `position` is not a custom attribute, so you have to store it with another name and you will get the new position.

I think to avoid this problem is to use different names in GN.

3DSinghVFX:
> Also I think this has been mentioned by Omar's second point (Or maybe I am wrong)
> I thought this should be fixed in [7690f02](https://github.com/JacquesLucke/animation_nodes/commit/7690f02bdc366b4d9fb2a463648da09e53d9de51)? but it didn't seem so.

That was to preserve attributes while replication of mesh in AN like UVs, Vertex Colors, etc.

> Basically the bug is that if you just simple divide mesh, whatever attribute you set can't be seen or used.
> This issue is in the current version of EN custom node as well.
> IMG1
> IMG2

Actually, the custom attribute set according to the data of *undivided mesh*. So, you divide the mesh then they will be removed by GN. Even in GN, it does not work,
Screenshot from 2021-03-08 12-35-46

Expected result or correct way,
Screenshot from 2021-03-08 12-36-17
 
So, I do not know is there a way to handle that because we can't get data in between from the GN node-tree to AN.

Gerstmann-Bradley:
> Actually, the custom attribute set according to the data of _undivided mesh_. So, you divide the mesh then they will be removed by GN. Even in GN, it does not work,
> Expected result or correct way,
> So, I do not know is there a way to handle that because we can't get data in between from the GN node-tree to AN.

Thanks for answering these questions!
For the quoted part, after thinking several scenarios,
I may even consider this as a bug, or at least an issue that needs to be handled in blender somehow.
Would you be able to reach the developers like Jacques discussing this issue?

OmarEmaraDev:
@3DSinghVFX The type, domain, and data type should use enums instead of strings. Also `ATTRIBUTE` -> `CUSTOM`.

The type of `data` in this case can be the super class of those lists, so `CList`. If it can be more than a CList, then you can use the type `object`, typing is not very important here.

Now that I think about it, we should probably make that into a class and add methods for attribute replication, copying, and appending and so on. So basically transfer most of the attribute code from the mesh class into the attribute class.

3DSinghVFX:
> @3DSinghVFX The type, domain, and data type should use enums instead of strings. Also `ATTRIBUTE` -> `CUSTOM`.

Okay, that is more clear. 

> The type of `data` in this case can be the super class of those lists, so `CList`. If it can be more than a CList, then you can use the type `object`, typing is not very important here.

Okay.

> Now that I think about it, we should probably make that into a class and add methods for attribute replication, copying, and appending and so on. So basically transfer most of the attribute code from the mesh class into the attribute class.

Ahh, I totally agree with you. Do we put that code in a separate file because in the future we may have custom-attributes for other objects e.g., Curve, Volume, etc?

OmarEmaraDev:
@3DSinghVFX Yes. Put that in a new file.

animation_nodes/data_structures/attributes/attribute.pxd
```diff
@@ -0,0 +1,14 @@
+cdef enum AttributeType:
+    UVMAP, MATERIAL_INDEX, VERTEX_COLOR, CUSTOM
+cdef enum AttributeDomain:
+    POINT, EDGE, POLYGON, CORNER
+cdef enum AttributeDataType:
+    INT, FLOAT, FLOAT2, FLOAT_VECTOR, FLOAT_COLOR, BYTE_COLOR, BOOLEAN
```
OmarEmaraDev, line 6: Put those on different lines. `UVMAP` -> `UV_MAP`.

animation_nodes/data_structures/attributes/attribute.pxd
```diff
@@ -0,0 +1,14 @@
+cdef enum AttributeType:
```
OmarEmaraDev, line 1: Declare this as `cpdef` so that you can access them in Python. They should be accessible in Python in a manner similar to [PEP 435](https://www.python.org/dev/peps/pep-0435/).

animation_nodes/data_structures/attributes/attribute.pxd
```diff
@@ -0,0 +1,14 @@
...
+    POINT, EDGE, POLYGON, CORNER
+cdef enum AttributeDataType:
+    INT, FLOAT, FLOAT2, FLOAT_VECTOR, FLOAT_COLOR, BYTE_COLOR, BOOLEAN
+
+cdef class Attribute:
+    cdef:
+        public str name
+        public AttributeType type
+        public AttributeDomain domain
+        public AttributeDataType dataType
```
OmarEmaraDev, line 13: Those should be readonly. If the domain or dataType change, the data will also need to change, so those should not be mutable.

animation_nodes/data_structures/attributes/attribute.pxd
```diff
@@ -0,0 +1,14 @@
...
+        public object data
```
OmarEmaraDev, line 14: Can this be `CList` of is `object` necessary?

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,120 @@
...
+                         self.data.copy())
+
+    def repeated(self, Py_ssize_t amount = -1):
+        return Attribute(self.name, self.type, self.domain, self.dataType,
+                         self.data.repeated(amount = amount))
+
+    def extend(self, object data):
+        self.data.extend(data)
+
+    @property
```
OmarEmaraDev, line 34: I don't like class properties to be honest. I would use standard methods for this.
Also, most of those asString properties are not needed now that we can access the enums in Python. So those should be removed.

animation_nodes/data_structures/meshes/mesh_data.pxd
```diff
@@ -10,4 +10,4 @@ cdef class Mesh:
         dict derivedMeshDataCache
         object uvMaps
         object vertexColorLayers
```
OmarEmaraDev, line 12: Those should be removed. The same goes for `materialIndices`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -60,11 +61,13 @@ cdef class Mesh:
...
+        self.attributes = OrderedDict()
 
     def getMeshProperties(self):
         return (
-            (self.uvMaps, Vector2DList),
-            (self.vertexColorLayers, ColorList),
+            (self.uvMaps, Attribute),
+            (self.vertexColorLayers, Attribute),
+            (self.attributes, Attribute),
         )
```
OmarEmaraDev, line 71: This should be removed. You can loop over `attributes` directly.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,35 +135,53 @@ cdef class Mesh:
         else:
             raise Exception("invalid length")
 
-    def insertUVMap(self, str name, Vector2DList uvs):
-        if len(uvs) == len(self.polygons.indices):
-            self.uvMaps[name] = uvs
+    def insertAttribute(self, str name, str type, str domain, str dataType, object data):
```
OmarEmaraDev, line 138: Use a guard condition, and use enums instead of str.

```python
if referenceAmount != len(data):
    raise Exception("invalid length")

attribute = Attribute()
...
...
```

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,35 +135,53 @@ cdef class Mesh:
...
+            attribute.typeAsString = type
+            attribute.domainAsString = domain
+            attribute.dataTypeAsString = dataType
+            attribute.data = data
+
+            if type == "UVMAP":
+                self.uvMaps[name] = attribute
+            elif type == "VERTEX_COLOR":
+                self.vertexColorLayers[name] = attribute
+            else:
```
OmarEmaraDev, line 160: Should be removed.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -225,17 +247,20 @@ cdef class Mesh:
...
 
             for name in sourceMeshProperty.keys():
                 if name not in meshProperty:
-                    extension = meshPropertyType(length = self.polygons.indices.length)
+                    attribute = sourceMeshProperty[name]
+                    extension = attribute.getDataType()(length = self.polygons.indices.length)
                     extension.fill(0)
-                    meshProperty[name] = extension + sourceMeshProperty[name]
+                    data = extension + attribute.data
+                    meshProperty[name] = Attribute(name, attribute.type, attribute.domain, attribute.dataType, data)
```
OmarEmaraDev, line 263: Most of this should be be moved to the attribute class. Instead of having a `extend` method in the attribute class, make it `appendAttribute` or something like that. The same for for other attribute methods.

OmarEmaraDev (changes requested):
This is going into the right direction. But there are general things to note.

- We use enums to avoid passing strings. So all strings should be replaced with references to the enum. Check the review for more details.
- We turned material indices, uv maps, and vertex colors into attributes, so they should be removed from the mesh structure.
- Mesh methods like append and replicate should loop over the attributes and call specialized methods that does the job. Attributes shouldn't expose low level data methods like extend and repeat.

3DSinghVFX:
> This is going into the right direction. But there are general things to note.
> 
>     * We use enums to avoid passing strings. So all strings should be replaced with references to the enum. Check the review for more details.

Okay.

>     * We turned material indices, uv maps, and vertex colors into attributes, so they should be removed from the mesh structure.

We can remove that but for UV Nodes (Vertex Colors Nodes) we have to modify them to access respective attributes.
BTW, we can add general `Custom Attribute Nodes -> Attribute Nodes` and these nodes will allow to `get, insert and set` all types of attributes. So, we can remove UV, Vertex Color, and other similar nodes.

>     * Mesh methods like append and replicate should loop over the attributes and call specialized methods that does the job. Attributes shouldn't expose low level data methods like extend and repeat.

Okay.

OmarEmaraDev:
> We can remove that but for UV Nodes (Vertex Colors Nodes) we have to modify them to access respective attributes.

Yes, you should do that. As for the decision to replace other nodes, lets delay this decision for later.

animation_nodes/data_structures/attributes/attribute.pxd
3DSinghVFX: Done.

animation_nodes/data_structures/attributes/attribute.pxd
3DSinghVFX: Thanks for the information. It is done.

animation_nodes/data_structures/attributes/attribute.pxd
3DSinghVFX: It is changed.

animation_nodes/data_structures/attributes/attribute.pyx
3DSinghVFX: Removed.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/attributes/attribute.pxd
3DSinghVFX: Removed.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Added `appendAttribute` method.

animation_nodes/data_structures/meshes/mesh_data.pxd
3DSinghVFX: Removed.

animation_nodes/data_structures/attributes/attribute.pxd
```diff
@@ -0,0 +1,28 @@
...
+    BYTE_COLOR,
+    BOOLEAN
+
+cdef class Attribute:
+    cdef:
+        readonly str name
+        readonly AttributeType type
+        readonly AttributeDomain domain
+        readonly AttributeDataType dataType
+        readonly data
```
OmarEmaraDev, line 28: Can this be `CList`?

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,86 @@
...
+        self.data = data
+
+    def copy(self):
+        return Attribute(self.name, self.type, self.domain, self.dataType,
+                         self.data.copy())
+
+    def extend(self, data):
+        self.data.extend(data)
+
+    def repeated(self, Py_ssize_t amount = -1):
```
OmarEmaraDev, line 30: This should repeat the attribute in-place. It should be called `replicate`. And the amount should not have a default value, -1 doesn't make sense here.

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,86 @@
...
+
+    def appendAttribute(self, sourceAttribute = None, Py_ssize_t amount = -1):
+        if sourceAttribute is not None:
+            self.extend(sourceAttribute.data)
+        else:
+            extension = self.getDataType()(length = amount)
+            extension.fill(0)
+            self.extend(extension)
+
+    def getTypeAsString(self):
```
OmarEmaraDev, line 42: I still don't see why we need all of those string functions. Why not just use the enums we defined?

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,86 @@
...
+            return "FLOAT2"
+        elif self.dataType == FLOAT_VECTOR:
+            return "FLOAT_VECTOR"
+        elif self.dataType == FLOAT_COLOR:
+            return "FLOAT_COLOR"
+        elif self.dataType == BYTE_COLOR:
+            return "BYTE_COLOR"
+        return "BOOLEAN"
+
+    def getDataType(self):
```
OmarEmaraDev, line 75: getListType

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,86 @@
...
+        if self.dataType == INT:
+            return LongList
+        elif self.dataType == FLOAT:
+            return FloatList
+        elif self.dataType == FLOAT2:
+            return Vector2DList
+        elif self.dataType == FLOAT_VECTOR:
+            return Vector3DList
+        elif self.dataType in [FLOAT_COLOR, BYTE_COLOR]:
```
OmarEmaraDev, line 84: `()` instead of `[]`

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -41,30 +55,25 @@ cdef class Mesh:
...
         self.edges = edges
         self.polygons = polygons
-        self.materialIndices = materialIndices
 
         self.derivedMeshDataCache = {}
-        self.uvMaps = OrderedDict()
-        self.vertexColorLayers = OrderedDict()
+        self.attributes = OrderedDict()
 
     def getMeshProperties(self):
```
OmarEmaraDev, line 74: This function should be removed. Loop over attributes directly.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,35 +141,56 @@ cdef class Mesh:
         else:
             raise Exception("invalid length")
 
-    def insertUVMap(self, str name, Vector2DList uvs):
-        if len(uvs) == len(self.polygons.indices):
-            self.uvMaps[name] = uvs
+    def insertAttribute(self, str name, str type, str domain, str dataType, object data):
```
OmarEmaraDev, line 144: Use enums instead of strings.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,35 +141,56 @@ cdef class Mesh:
...
             raise Exception("invalid length")
 
-    def getUVMaps(self):
-        return list(self.uvMaps.items())
+        self.attributes[name] = Attribute(name, AttributeType[type], AttributeDomain[domain],
+                                              AttributeDataType[dataType], data)
 
-    def getUVMapNames(self):
-        return list(self.uvMaps.keys())
+    def getAttributes(self, str type = ""):
```
OmarEmaraDev, line 160: Enums instead of string.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -211,31 +240,30 @@ cdef class Mesh:
...
         Polygons: {len(self.polygons)}
-        UV Maps: {self.getUVMapNames()}
-        Vertex Colors: {self.getVertexColorLayerNames()}""")
+        UV Maps: {self.getAttributeNames("UV_MAP")}
+        Vertex Colors: {self.getAttributeNames("VERTEX_COLOR")}
+        Custom Attributes: {self.getAttributeNames("CUSTOM")}""")
 
     def replicateMeshProperties(self, Mesh source, long amount):
         for ((meshProperty, _), (sourceMeshProperty, _)) in zip(
                 self.getMeshProperties(), source.getMeshProperties()):
```
OmarEmaraDev, line 249: Use attributes directly instead of `getMeshProperties`.

OmarEmaraDev (changes requested):
You are still using strings instead of enums in most of the code.

animation_nodes/data_structures/attributes/attribute.pxd
3DSinghVFX: Omar, What is `CList`? Do you mean the lists that we have in AN?

animation_nodes/data_structures/attributes/attribute.pyx
3DSinghVFX: Okay.

animation_nodes/data_structures/attributes/attribute.pyx
3DSinghVFX: Omar, enums returns `int` not string. So. we have to map them manually.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Okay.

Gerstmann-Bradley:
Hi, I have found an issue in EN, and thought since this should be the technically the newest build, I am reporting the issue here.
**I have just tested with artifact built from this pull request, and the issue is replicable**.

Basically, when I am set the custom attribute.
During the typing of attribute's name, all processes will be recorded,
and thus I am generating numerous attribute that won't be used at the end.

Take "COL" as an example, during the process, I will generate "C", "CO" and finally the "COL".
I am not too sure about the solution, so I will just leave the problem here.

Thanks for the development.
image

3DSinghVFX:
> Hi, I have found an issue in EN, and thought since this should be technically the newest build, I am reporting the issue here.
> **I have just tested with artifact built from this pull request, and the issue is replicable**.

Thanks for the feedback. It is not new issue, we aware of that issue.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Removed.

animation_nodes/nodes/mesh/get_custom_attribute.py
```diff
@@ -0,0 +1,49 @@
...
+from bpy.props import *
+from ... base_types import AnimationNode
+from ... data_structures import DoubleList
+
+class GetCustomAttributeNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_GetCustomAttributeNode"
+    bl_label = "Get Custom Attribute"
+    errorHandlingType = "EXCEPTION"
+
+    dataType: StringProperty(name = "Color Mode", default = "FLOAT", update = AnimationNode.refresh)
```
OmarEmaraDev, line 11: What is this property exactly? It is not drawn, it is not assigned, it is named `Color Mode`. I am not sure what it does.

animation_nodes/nodes/mesh/mesh_info.py
```diff
@@ -31,8 +32,10 @@ def getExecutionCode(self, required):
         if "polygonCenters" in required:
             yield "polygonCenters = mesh.getPolygonCenters()"
         if "materialIndices" in required:
-            yield "materialIndices = mesh.materialIndices"
+            yield "materialIndices = mesh.getMaterialIndices()"
```
OmarEmaraDev, line 35: As we said above, if this returned None, return a zero array instead.

animation_nodes/nodes/mesh/mesh_object_input.py
```diff
@@ -116,17 +119,36 @@ def getPolygonCenters(self, mesh, object, useWorldSpace):
...
 
-    def loadUVs(self, mesh, sourceMesh, object):
-        if object.mode == "OBJECT":
-            for uvMapName in sourceMesh.uv_layers.keys():
-                mesh.insertUVMap(uvMapName, sourceMesh.an.getUVMap(uvMapName))
-        else:
-            self.setErrorMessage("Object has to be in object mode to load UV maps.")
-    
-    def loadVertexColors(self, mesh, sourceMesh, object):
+    def loadAttributes(self, type, mesh, sourceMesh, object):
```
OmarEmaraDev, line 122: Better split this function for each type.

OmarEmaraDev (changes requested):

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,87 @@
...
+
+    def appendAttribute(self, sourceAttribute = None, Py_ssize_t amount = -1):
+        if sourceAttribute is not None:
+            self.extend(sourceAttribute.data)
+        else:
+            extension = self.getListType()(length = amount)
+            extension.fill(0)
+            self.extend(extension)
+
+    def getTypeAsString(self):
```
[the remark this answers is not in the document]
3DSinghVFX: Okay, I'll check.

animation_nodes/nodes/mesh/get_custom_attribute.py
3DSinghVFX: Sorry, the name is wrong. Actually, this is added to change the output type according to attribute data. Is there any other way to do this?

OmarEmaraDev: @3DSinghVFX This is probably not a good idea. What if this node is inside a group and the attribute changes?
I would say the best solution is to make this explicit. Add an enum for the output type. If the attribute does not match the output type, warn the user to change the enum manually.

While at it, `getAllAttributes` seems like it returns a list. Just remove it and use `mesh.attributes` directly.

animation_nodes/data_structures/attributes/attribute.pyx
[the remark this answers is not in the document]
3DSinghVFX: First, I tried in the `attribute.pyx` file,
```
Error compiling Cython file:
------------------------------------------------------------
...
            extension = self.getListType()(length = amount)
            extension.fill(0)
            self.extend(extension)

    def getTypeAsString(self):
        print(self.type.name)
                      ^
------------------------------------------------------------

animation_nodes/data_structures/attributes/attribute.pyx:44:23: Object of type 'AttributeType' has no attribute 'name'

```

Then, I also tried in the `mesh_data.pyx` file,
```
print(attribute.type.name)

------------------------------------------------------------
  File "animation_nodes/data_structures/meshes/mesh_data.pyx", line 166, in animation_nodes.data_structures.meshes.mesh_data.Mesh.getAttributeNames
AttributeError: 'int' object has no attribute 'name'
```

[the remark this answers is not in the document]
OmarEmaraDev: @3DSinghVFX Don't do it in Cython files. This would only works in Python files, which is what we want anyways.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -175,15 +197,15 @@ cdef class Mesh:
 
     def copy(self):
         mesh = Mesh(self.vertices.copy(), self.edges.copy(),
-                    self.polygons.copy(), self.materialIndices.copy())
+                    self.polygons.copy())
         mesh.copyMeshProperties(self)
         return mesh
 
     def copyMeshProperties(self, Mesh source):
```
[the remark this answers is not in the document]
3DSinghVFX: Do we also rename the `replicateMeshProperties -> replicateAttributes`?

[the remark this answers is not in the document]
OmarEmaraDev: Yes.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -132,35 +136,53 @@ cdef class Mesh:
...
+        if attribute is not None:
+            if type == attribute.type: return attribute
+            elif type == AttributeType["VERTEX_COLOR"] and attribute.dataType == AttributeDataType["BYTE_COLOR"]:
+                return attribute
+        return attribute
+
+    def getAllAttributes(self):
+        return self.attributes
+
+    def getMaterialIndices(self):
```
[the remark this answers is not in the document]
3DSinghVFX: Can we write this function like,
```
def getMaterialIndices(self):
        attribute = self.getAttribute("Material Indices", AttributeType["MATERIAL_INDEX"])
        if attribute is None: return LongList(length = len(self.polygons))
        return attribute.data
```

animation_nodes/data_structures/attributes/attribute.pyx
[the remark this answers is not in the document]
3DSinghVFX: Okay.

animation_nodes/data_structures/meshes/mesh_data.pyx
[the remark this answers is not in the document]
3DSinghVFX: Can we write this function like this,
```
def getMaterialIndices(self):
        attribute = self.getAttribute("Material Indices", AttributeType["MATERIAL_INDEX"])
        if attribute is None: return LongList(length = len(self.polygons))
        return attribute.data
```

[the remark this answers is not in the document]
OmarEmaraDev: @3DSinghVFX You can, but you have other places were you just want to check if material indices exist. As far as I can see, you only need that in the Info node, so I would say do it manually there.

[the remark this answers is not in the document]
3DSinghVFX: We also need this in *Create BMesh* and *Mesh Object Output* nodes.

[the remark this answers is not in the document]
OmarEmaraDev: @3DSinghVFX For those nodes it, we don't need a zeros array. We either set an existing material indices, or we don't. We don't create a zeros array.

animation_nodes/nodes/mesh/combine_mesh_data.py
```diff
@@ -36,7 +36,10 @@ def execute(self, vertexLocations, edgeIndices, polygonIndices, materialIndices)
             materialIndices = LongList(length = len(polygonIndices))
             materialIndices.fill(0)
         try:
-            return Mesh(vertexLocations, edgeIndices, polygonIndices,
-                    materialIndices, skipValidation = self.skipValidation)
+            mesh = Mesh(vertexLocations, edgeIndices, polygonIndices, skipValidation = self.skipValidation)
+            mesh.insertAttribute("Material Indices", AttributeType["MATERIAL_INDEX"], AttributeDomain["FACE"],
+                                  AttributeDataType["INT"], materialIndices)
```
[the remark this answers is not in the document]
3DSinghVFX: > Also, it might be a good idea to make `insertAttribute` take an attribute object instead of take the individual elements.

Omar, the *Insert Custom Attribute* node operates on the mesh but not an object.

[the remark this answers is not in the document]
OmarEmaraDev: @3DSinghVFX I meant a python object. What I was saying is that instead of passing those individual values, construct an Attribute object from them and pass that object. So it should have the following interface:
```
insertAttribute(attribute: Attribute)
```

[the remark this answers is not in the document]
3DSinghVFX: Okay.

animation_nodes/nodes/mesh/get_custom_attribute.py
3DSinghVFX: > While at it, `getAllAttributes` seems like it returns a list. Just remove it and use `mesh.attributes` directly.

Omar, this gave an error:
```
attributes = mesh.attributes
AttributeError: 'animation_nodes.data_structures.meshes.mesh_data.M' object has no attribute 'attributes'
```

OmarEmaraDev: @3DSinghVFX  Probably because it is cdef-ed. In this case, just rename it to `getAttributes`.

animation_nodes/data_structures/attributes/attribute.pyx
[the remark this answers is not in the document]
3DSinghVFX: Omar, this gave an error even in Python file,
```
 File "/home/ks/.config/blender/2.93/scripts/addons/animation_nodes/nodes/mesh/get_custom_attribute.py", line 43, in execute
    print(attribute.type.name)
AttributeError: 'int' object has no attribute 'name'
```

animation_nodes/nodes/mesh/get_custom_attribute.py
3DSinghVFX: We have already the `getAttributes` method which gave attributes by type. That is why I added this function.

animation_nodes/data_structures/attributes/attribute.pyx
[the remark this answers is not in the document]
OmarEmaraDev: Okay. Leave it as is and I will sort this later myself.

animation_nodes/nodes/mesh/get_custom_attribute.py
OmarEmaraDev: @3DSinghVFX How about renaming the other one `getAllAttributesOfType`?
I suppose we can remove this function altogether and add another one `getAttributeByName` to hide the underlying dict.

3DSinghVFX: Okay. We all attribute names to show in the `Get Custom Attributes` node.

animation_nodes/data_structures/attributes/attribute.pyx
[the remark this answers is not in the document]
3DSinghVFX: Okay.

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -0,0 +1,87 @@
...
+            return "FLOAT2"
+        elif self.dataType == FLOAT_VECTOR:
+            return "FLOAT_VECTOR"
+        elif self.dataType == FLOAT_COLOR:
+            return "FLOAT_COLOR"
+        elif self.dataType == BYTE_COLOR:
+            return "BYTE_COLOR"
+        return "BOOLEAN"
+
+    def getListType(self):
```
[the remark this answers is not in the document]
3DSinghVFX: Omar, can please elaborate on this? Did you mean something like this,
```
def getListType(self):
        dataTypes = {INT: LongList, FLOAT: FloatList, FLOAT2: Vector2DList,
                     FLOAT_VECTOR: Vector3DList, FLOAT_COLOR: ColorList,
                     BYTE_COLOR: ColorList, BOOLEAN: BooleanList}
        return dataTypes[self.dataType]
```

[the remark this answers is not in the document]
OmarEmaraDev: @3DSinghVFX Yes. But this dictionary should be defined at the module level, outside of the class. Maybe name it `cListFromDataType`.

OmarEmaraDev (approved):
Thanks for working on this @3DSinghVFX!

There are some minor things that we need to take care of, but we can do handle them after merging.

3DSinghVFX:
> Thanks for working on this @3DSinghVFX!

> There are some minor things that we need to take care of, but we can do handle them after merging.

Okay :)

PR #1699, opened 2021-04-15
Fix mesh generation methods for Material Indices attribute

3DSinghVFX:
I have fixed the *Mesh Generation* methods for the *Material Indices* attribute. I have also updated the *CHANGELOG.md*.


OmarEmaraDev (approved):
Thanks!

PR #1705, opened 2021-04-22
Refactor mesh attributes

3DSinghVFX:
In this patch, I have refactored the mesh attribute storage as discussed here https://github.com/JacquesLucke/animation_nodes/commit/558dc15219eae4780517744b0e9c9afbefbebcac#commitcomment-49758333. Now, it is possible to have different type of attributes with the same name. 

The *UV Map* and *Vertex Color* attributes have their own dictionary for storage. However, the *Custom* and *Material Indices* attributes share the same dictionary for storage.

### DIFF
--- a/animation_nodes/data_structures/attributes/attribute.pyx
+++ b/animation_nodes/data_structures/attributes/attribute.pyx
@@ -19,8 +19,8 @@ cListFromDataType = {
 }
 
 stringFromType = {
-    UV_MAP: "UV_MAP",
     MATERIAL_INDEX: "MATERIAL_INDEX",
+    UV_MAP: "UV_MAP",
     VERTEX_COLOR: "VERTEX_COLOR",
     CUSTOM: "CUSTOM",
 }
@@ -85,13 +85,13 @@ cdef class Attribute:
         self.data = extension + self.data
 
     def getTypeAsString(self):
-        return stringFromType.get((self.type))
+        return stringFromType[self.type]
 
     def getDomainAsString(self):
-        return stringFromDomain.get((self.domain))
+        return stringFromDomain[self.domain]
 
     def getListTypeAsString(self):
-        return stringFromDataType.get((self.dataType))
+        return stringFromDataType[self.dataType]
 
     def getListType(self):
-        return cListFromDataType.get(self.dataType)
+        return cListFromDataType[self.dataType]
--- a/animation_nodes/data_structures/attributes/attribute.pyx
+++ b/animation_nodes/data_structures/attributes/attribute.pyx
@@ -85,13 +85,13 @@ cdef class Attribute:
         self.data = extension + self.data
 
     def getTypeAsString(self):
-        return stringFromType(self.type)
+        return stringFromType.get((self.type))
 
     def getDomainAsString(self):
-        return stringFromDomain(self.domain)
+        return stringFromDomain.get((self.domain))
 
     def getListTypeAsString(self):
-        return stringFromDataType(self.dataType)
+        return stringFromDataType.get((self.dataType))
 
     def getListType(self):
         return cListFromDataType.get(self.dataType)
--- a/animation_nodes/data_structures/meshes/mesh_data.pyx
+++ b/animation_nodes/data_structures/meshes/mesh_data.pyx
@@ -72,6 +72,7 @@ cdef class Mesh:
         self.customAttributes = OrderedDict()
         self.uvMapAttributes = OrderedDict()
         self.vertexColorAttributes = OrderedDict()
+        self.materialIndices = None
 
     def getMeshAttributes(self):
         return (
@@ -149,7 +150,9 @@ cdef class Mesh:
         if self.getAttributeDomainLength(attribute.domain) != len(attribute.data):
             raise Exception("invalid length")
 
-        if attribute.type == AttributeType.UV_MAP:
+        if attribute.type == AttributeType.MATERIAL_INDEX:
+            self.materialIndices = attribute
+        elif attribute.type == AttributeType.UV_MAP:
             self.uvMapAttributes[attribute.name] = attribute
         elif attribute.type == AttributeType.VERTEX_COLOR:
             self.vertexColorAttributes[attribute.name] = attribute
@@ -189,7 +192,7 @@ cdef class Mesh:
             return None
 
     def getMaterialIndices(self):
-        return self.getAttribute("Material Indices", AttributeType.MATERIAL_INDEX)
+        return self.materialIndices
 
     def getVertexLinkedVertices(self, long vertexIndex):
         cdef LongList neighboursAmounts, neighboursStarts, neighbours, neighbourEdges
@@ -209,6 +212,10 @@ cdef class Mesh:
         return mesh
 
     def copyAttributes(self, Mesh source):
+        sourceMaterialIndices = source.materialIndices
+        if sourceMaterialIndices is not None:
+            self.materialIndices = sourceMaterialIndices.copy()
+
         for ((meshAttributes, _), (sourceMeshAttributes, _)) in zip(
                 self.getMeshAttributes(), source.getMeshAttributes()):
             for name, value in sourceMeshAttributes.items():
@@ -244,6 +251,10 @@ cdef class Mesh:
         Custom Attributes: {self.getAttributeNames(AttributeType.CUSTOM)}""")
 
     def replicateAttributes(self, Mesh source, long amount):
+        sourceMaterialIndices = source.materialIndices
+        if sourceMaterialIndices is not None:
+            self.materialIndices = sourceMaterialIndices.replicate(amount)
+
         for ((meshAttributes, _), (sourceMeshAttributes, _)) in zip(
                 self.getMeshAttributes(), source.getMeshAttributes()):
             for name, attribute in sourceMeshAttributes.items():
@@ -260,6 +271,20 @@ cdef class Mesh:
             return self.polygons.indices.length
 
     def appendAttributes(self, Mesh source):
+        sourceMaterialIndices = source.materialIndices
+        meshMaterialIndices = self.materialIndices
+        if meshMaterialIndices is not None:
+            if sourceMaterialIndices and sourceMaterialIndices.similar(meshMaterialIndices):
+                meshMaterialIndices.append(sourceMaterialIndices)
+            else:
+                length = source.getAttributeDomainLength(meshMaterialIndices.domain)
+                meshMaterialIndices.appendZeros(length)
+        elif sourceMaterialIndices is not None:
+            length = self.getAttributeDomainLength(sourceMaterialIndices.domain)
+            newAttribute = sourceMaterialIndices.copy()
+            newAttribute.prependZeros(length)
+            self.materialIndices = newAttribute
+
         for ((meshAttributes, _), (sourceMeshAttributes, _)) in zip(
                 self.getMeshAttributes(), source.getMeshAttributes()):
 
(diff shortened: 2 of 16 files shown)


3DSinghVFX:
@OmarEmaraDev Do you think we need to separate the *Custom and Material Indices* attributes dictionary?

OmarEmaraDev:
@3DSinghVFX Material indices should be declared as an Attribute in the class, not in a dictionary. It can be None initialized by default.

3DSinghVFX:
> @3DSinghVFX Material indices should be declared as an Attribute in the class, not in a dictionary. It can be None initialized by default.

Okay.

3DSinghVFX:
> @3DSinghVFX Material indices should be declared as an Attribute in the class, not in a dictionary. It can be None initialized by default.

@OmarEmaraDev I have added Material Indices as an Attribute.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -140,37 +150,51 @@ cdef class Mesh:
...
+        elif attribute.type == AttributeType.UV_MAP:
+            self.uvMapAttributes[attribute.name] = attribute
+        elif attribute.type == AttributeType.VERTEX_COLOR:
+            self.vertexColorAttributes[attribute.name] = attribute
+        elif attribute.type == AttributeType.MATERIAL_INDEX:
+            self.materialIndices = attribute
+        else:
+            self.customAttributes[attribute.name] = attribute
 
     def getAttributes(self, AttributeType type):
```
OmarEmaraDev, line 164: I would specialize all of those method. `getAllCustomAttributes`, `getAllVertexColors`, `getAllUVMaps`.

Also, you don't need to return the name, the name is already a member of the attribute.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -140,37 +150,51 @@ cdef class Mesh:
...
-        return attributes
+        if type == AttributeType.UV_MAP:
+            return list(self.uvMapAttributes.items())
+        elif type == AttributeType.VERTEX_COLOR:
+            return list(self.vertexColorAttributes.items())
+        else:
+            attributes = list()
+            for name, attribute in self.customAttributes.items():
+                if type == attribute.type: attributes.append((name, attribute))
+            return attributes
```
OmarEmaraDev, line 173: Why the special handing here. Won't `type` always be equal to `attribute.type`?

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -140,37 +150,51 @@ cdef class Mesh:
...
 
     def getAttributeNames(self, AttributeType type):
```
OmarEmaraDev, line 175: This is almost identical to the above method. Just call the above method in a list comprehension and extract the names.

This also should be specialized. `getAllCustomAttributeNames`, `getAllVertexColorNames`, `getAllUVMapNames`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -140,37 +150,51 @@ cdef class Mesh:
...
+        if type == AttributeType.UV_MAP:
+            return self.uvMapAttributes.get(name, None)
+        elif type == AttributeType.VERTEX_COLOR:
+            return self.vertexColorAttributes.get(name, None)
+        else:
+            attribute = self.customAttributes.get(name, None)
+            if attribute is not None and type == attribute.type: return attribute
+            return None
 
     def getMaterialIndices(self):
```
OmarEmaraDev, line 196: `getMaterialIndicesAttribute`.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -140,37 +150,51 @@ cdef class Mesh:
...
+            return list(self.uvMapAttributes.keys())
+        elif type == AttributeType.VERTEX_COLOR:
+            return list(self.vertexColorAttributes.keys())
+        else:
+            attributeNames = list()
+            for name, attribute in self.customAttributes.items():
+                if type == attribute.type: attributeNames.append(name)
+            return attributeNames
 
     def getAttribute(self, str name, AttributeType type):
```
OmarEmaraDev, line 186: Also specialize this method.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -190,10 +214,14 @@ cdef class Mesh:
...
-        for name, value in sourceMeshAttributes.items():
-            meshAttributes[name] = value.copy()
+        sourceMaterialIndices = source.materialIndices
+        if sourceMaterialIndices is not None:
+            self.materialIndices = sourceMaterialIndices.copy()
+
+        for ((meshAttributes, _), (sourceMeshAttributes, _)) in zip(
+                self.getMeshAttributes(), source.getMeshAttributes()):
+            for name, value in sourceMeshAttributes.items():
+                meshAttributes[name] = value.copy()
```
OmarEmaraDev, line 224: I would remove the `getMeshAttributes` and loop explicitly over each of the dictionaries.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -225,10 +253,14 @@ cdef class Mesh:
...
     def replicateAttributes(self, Mesh source, long amount):
-        meshAttributes = self.attributes
-        sourceMeshAttributes = source.attributes
-        for name, attribute in sourceMeshAttributes.items():
-            meshAttributes[name] = attribute.replicate(amount)
+        sourceMaterialIndices = source.materialIndices
+        if sourceMaterialIndices is not None:
+            self.materialIndices = sourceMaterialIndices.replicate(amount)
+
+        for ((meshAttributes, _), (sourceMeshAttributes, _)) in zip(
```
OmarEmaraDev, line 260: Same here. Loop explicitly over each of the dictionaries.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -241,21 +273,38 @@ cdef class Mesh:
             return self.polygons.indices.length
 
     def appendAttributes(self, Mesh source):
```
OmarEmaraDev, line 275: You can probably add a method that appends a single attribute, then in this method, loop explicitly over each of the dictionaries and call the newly created method.

OmarEmaraDev (changes requested):

animation_nodes/data_structures/attributes/attribute.pyx
```diff
@@ -85,13 +85,13 @@ cdef class Attribute:
...
-        return stringFromType(self.type)
+        return stringFromType.get((self.type))
 
     def getDomainAsString(self):
-        return stringFromDomain(self.domain)
+        return stringFromDomain.get((self.domain))
 
     def getListTypeAsString(self):
-        return stringFromDataType(self.dataType)
+        return stringFromDataType.get((self.dataType))
```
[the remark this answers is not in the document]
3DSinghVFX: Yes, we can.

animation_nodes/data_structures/meshes/mesh_data.pyx
```diff
@@ -140,37 +150,51 @@ cdef class Mesh:
...
 
-        self.attributes[attribute.name] = attribute
+        if attribute.type == AttributeType.MATERIAL_INDEX:
+            self.materialIndices = attribute
+        elif attribute.type == AttributeType.UV_MAP:
+            self.uvMapAttributes[attribute.name] = attribute
+        elif attribute.type == AttributeType.VERTEX_COLOR:
+            self.vertexColorAttributes[attribute.name] = attribute
+        elif attribute.type == AttributeType.MATERIAL_INDEX:
+            self.materialIndices = attribute
```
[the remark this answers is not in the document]
3DSinghVFX: Removed.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: I have a query, are we going to separate other `types` of attributes as well *e.g.,* Vertex Weight, Bevel Vertex/Edge Weight, etc.? If so, then we do not need this special handling?

OmarEmaraDev: @3DSinghVFX That's a good point. I didn't though about future additions. Give me second to think about this.

OmarEmaraDev: @3DSinghVFX What do you think about creating a so called `builtInAttributes`. That dictionary should hold single things  like Material Indices, Bevel Vertex/Edge weights, and so on.

3DSinghVFX: Other possible attributes are: 
1. Vertex Weight
2. Bevel Vertex Weight
3. Bevel Edge Weight
4. Crease
5. Seams
6. Face Groups

OmarEmaraDev: @3DSinghVFX  Vertex Weights and Face Maps are not really built-in, they are more similar to vertex colors and uv maps.

An ordered dictionary `builtInAttributes` can hold all of the single built-in attributes like your list [2, 5]. Rename the `getMeshAttributes` method to `getAttributeDictionaries`. Instead of returning a list of tuples, just return a list of dictionaries, the second element is never user in the tuple. Then adjust your code accordingly. Some of my review items will no longer stand, but finish up and I will rereview later.

3DSinghVFX: Okay.

3DSinghVFX: Another thing I would like to mention here is that *Vertex Color* is shared in *Custom Attribute*. So, in the current patch, I have added a condition in Vertex Color and Custom Attribute Nodes to avoid the same name conflict.

Should we handle that in the *InsertAttribute* method in `mesh_data.pyx` file?

OmarEmaraDev: @3DSinghVFX hmmm, at this point I would say lets separate them completely. We can divert from Blender in this case. Custom attribute nodes should only deal with custom attributes no other data. We already having uv and vertex color nodes to deal with other things.

3DSinghVFX: Omar, they are separated but we need to check only the name of Vertex Color and Custom Attribute (Byte Color). Otherwise, Vertex Color and Custom Attribute (Byte Color) with the same name will override each other in the node-tree,
Screenshot from 2021-04-24 09-06-45

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: This method is removed.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Done.

animation_nodes/data_structures/meshes/mesh_data.pyx
3DSinghVFX: Omar, the `getAttributeDictionaries` method works for all dictionaries.

3DSinghVFX:
@OmarEmaraDev I have modified codes based on the 1st review.

animation_nodes/data_structures/meshes/mesh_data.pyx
OmarEmaraDev: @3DSinghVFX  Lets leave it at that even with the override. The last thing I want is for the Get Custom Attribute node to only return custom attributes, we are diverting from the Blender logic to make things more intuitive.

3DSinghVFX: Okay.

OmarEmaraDev (approved):
Code looks much better now. Thanks!

PR #1707, opened 2021-04-27
Added Hexagon Grid node

harisreedhar:
hexagonGrid


OmarEmaraDev:
Thanks for picking this up!

I think the grid should be connected though. If the user wishes to have separate cells, one can use one of the separate polygons node. Moreover, that generation algorithm should be added in `animation_nodes/algorithms/mesh_generation`.

harisreedhar:
In sverchok they are using ``bmesh.ops.remove_doubles`` to make connected polygons. But it is very slow to compute.

OmarEmaraDev:
Right. This should be computed directly. Give it a try and if proved too hard, I will look into it.

harisreedhar:
> Right. This should be computed directly. Give it a try and if proved too hard, I will look into it.

I'll try

harisreedhar:
So far, I didn't find any algorithms i can understand. So i'm closing this for now.

PR #1708, opened 2021-04-28
Fix Mesh Info node.

3DSinghVFX:
I have fixed the wrong methods for names of attributes in the *Mesh Info* node.


OmarEmaraDev (approved):
Thanks!

Issue #1671, opened 2021-03-01
Proposal: Vectorize "transform" series of nodes


Gerstmann-Bradley:
**Not an urgent proposal**. But I think it will very useful to have these at some point.
Also it seems if people don't propose, certain development will never be made...

**In terms of priority, I mostly recommend "Vector" & "Spline"**;
and "matrix" will be the least to consider since we have "offset matrices".
(To rotate vectors, we can only use transform, or do a double conversion and then borrow the vectorization from "offset matrix". Thus having vectorization of "Transform Vector" will be the most helpful imo).
The other members in the series can wait. 
Glad we have loop function to help doing all these vectorizations for the time being.

![image](https://user-images.githubusercontent.com/55994236/109573320-d9b6ac80-7abc-11eb-99a2-912c2325cc86.png)

OmarEmaraDev:
Other transform nodes still need to be fully vectorized.

PR #1712, opened 2021-05-03
Vectorize Transform Vector node

harisreedhar:
Vectorized Matrix input of `Transform Vector` node.

Resolves: #1671

### DIFF
--- a/animation_nodes/nodes/vector/c_utils.pyx
+++ b/animation_nodes/nodes/vector/c_utils.pyx
@@ -116,8 +116,8 @@ def calculateVectorDotProducts(Py_ssize_t amount,
     return dotProducts
 
 def transformVirtualVectorList(Py_ssize_t amount,
-                        VirtualVector3DList vectors,
-                        VirtualMatrix4x4List matrices):
+                               VirtualVector3DList vectors,
+                               VirtualMatrix4x4List matrices):
     cdef Py_ssize_t i
     cdef Vector3DList output = Vector3DList(length = amount)
     for i in range(amount):
--- a/animation_nodes/nodes/vector/c_utils.pyx
+++ b/animation_nodes/nodes/vector/c_utils.pyx
@@ -1,7 +1,7 @@
-from ... math cimport Vector3, distanceVec3, lengthVec3, dotVec3
+from ... math cimport Vector3, distanceVec3, lengthVec3, dotVec3, transformVec3AsPoint
 from ... data_structures cimport (
     DoubleList, Vector3DList, CDefaultList, Vector2DList,
-    VirtualDoubleList, VirtualVector3DList, FloatList)
+    VirtualDoubleList, VirtualVector3DList, VirtualMatrix4x4List, FloatList)
 
 def combineVectorList(Py_ssize_t amount,
                       VirtualDoubleList x, VirtualDoubleList y, VirtualDoubleList z):
@@ -114,3 +114,12 @@ def calculateVectorDotProducts(Py_ssize_t amount,
     for i in range(amount):
         dotProducts.data[i] = dotVec3(vectors1.get(i), vectors2.get(i))
     return dotProducts
+
+def transformVirtualVectorList(Py_ssize_t amount,
+                        VirtualVector3DList vectors,
+                        VirtualMatrix4x4List matrices):
+    cdef Py_ssize_t i
+    cdef Vector3DList output = Vector3DList(length = amount)
+    for i in range(amount):
+        transformVec3AsPoint(output.data + i, vectors.get(i), matrices.get(i))
+    return output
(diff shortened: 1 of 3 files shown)


animation_nodes/nodes/vector/c_utils.pyx
```diff
@@ -114,3 +114,12 @@ def calculateVectorDotProducts(Py_ssize_t amount,
     for i in range(amount):
         dotProducts.data[i] = dotVec3(vectors1.get(i), vectors2.get(i))
     return dotProducts
+
+def transformVirtualVectorList(Py_ssize_t amount,
+                        VirtualVector3DList vectors,
+                        VirtualMatrix4x4List matrices):
```
OmarEmaraDev, line 120: Align those with the first argument.

animation_nodes/nodes/vector/transform_vector.py
```diff
@@ -6,19 +6,27 @@ class TransformVectorNode(bpy.types.Node, AnimationNode):
...
     def create(self):
         self.newInput(VectorizedSocket("Vector", "useVectorList",
             ("Vector", "vector"),
             ("Vectors", "vectors", dict(dataIsModified = True))))
+        self.newInput(VectorizedSocket("Matrix", "useMatrixList",
+            ("Matrix", "matrix"), ("Matrices", "matrices")))
 
-        self.newInput("Matrix", "Matrix", "matrix")
-
-        self.newOutput(VectorizedSocket("Vector", "useVectorList",
```
OmarEmaraDev, line 17: VectorizedSocket can take multiple properties as input. `["useVectorList", "useMatrixList"]`. This should avoid the if statement.

animation_nodes/nodes/vector/transform_vector.py
```diff
@@ -6,19 +6,27 @@ class TransformVectorNode(bpy.types.Node, AnimationNode):
     bl_label = "Transform Vector"
 
     useVectorList: VectorizedSocket.newProperty()
+    useMatrixList: VectorizedSocket.newProperty()
 
     def create(self):
         self.newInput(VectorizedSocket("Vector", "useVectorList",
             ("Vector", "vector"),
             ("Vectors", "vectors", dict(dataIsModified = True))))
```
OmarEmaraDev, line 14: `dict(dataIsModified = True)` is no longer needed.

animation_nodes/nodes/vector/transform_vector.py
```diff
@@ -6,19 +6,27 @@ class TransformVectorNode(bpy.types.Node, AnimationNode):
...
 
     def getExecutionCode(self, required):
-        if self.useVectorList:
-            return "vectors.transform(matrix)"
+        if any((self.useVectorList, self.useMatrixList)):
+            if not self.useVectorList: yield "vectors = Vector3DList.fromValue(vector)"
+            if not self.useMatrixList: yield "matrices = Matrix4x4List.fromValue(matrix)"
+            yield "_vectors = VirtualVector3DList.create(vectors, (0,0,0))"
+            yield "_matrices = VirtualMatrix4x4List.create(matrices, Matrix.Identity(4))"
+            yield "amount = max(len(vectors), len(matrices))"
```
OmarEmaraDev, line 29: Use `getMaxRealLength`.

OmarEmaraDev (changes requested):

animation_nodes/nodes/vector/transform_vector.py
harisreedhar: By using this method there is an issue with socket update.

https://user-images.githubusercontent.com/46858047/116970318-5a6c7180-acd5-11eb-96ec-9a123ef63421.mp4

animation_nodes/nodes/vector/transform_vector.py
harisreedhar: Done

animation_nodes/nodes/vector/transform_vector.py
harisreedhar: Done

animation_nodes/nodes/vector/transform_vector.py
OmarEmaraDev: Seems like a standard behavior for the vectorization selector. Perhaps this behavior should change globally then.

OmarEmaraDev (approved):
Thanks!

Gerstmann-Bradley:
Sorry I haven't checked yet, but in description it only mentions "Transform Vector" which raises my concern.
I was asking the whole series: "Transform Matrix", "Transform Spline" both of high priority, and then Transform Vector, etc.
Are the other nodes being vectorized as well? If so it will be so nice. Thank you both.

OmarEmaraDev:
Transform Matrix and Transforms Spline are already vectorized, no?

Gerstmann-Bradley:
> Transform Matrix and Transforms Spline are already vectorized, no?

No, at least none of the series member is vectorized before this commit, and I still haven't checked this commit yet.
image

OmarEmaraDev:
Okay right. One input is vectorized and the other isn't.

PR #1715, opened 2021-05-05
Vectorize Transform Matrix node

harisreedhar:
Vectorized all inputs of ``Transform Matrix`` node.

Resolves: #1671.


OmarEmaraDev (approved):
Thanks!

PR #1716, opened 2021-05-05
Added Hexagon distribution matrix

harisreedhar:
Added Hexagon mode for ``Distribute Matrix`` node.

### DIFF
--- a/animation_nodes/nodes/matrix/distribute_matrices.pyx
+++ b/animation_nodes/nodes/matrix/distribute_matrices.pyx
@@ -18,7 +18,7 @@ modeItems = [
     ("MESH", "Mesh", "", "NONE", 3),
     ("SPIRAL", "Spiral", "", "NONE", 4),
     ("SPLINE", "Spline", "", "NONE", 5),
-    ("HEXAGON", "Hexagon", "", "NONE", 6),
+    ("HEXAGONAL_GRID", "Hexagonal Grid", "", "NONE", 6),
 ]
 
 distanceModeItems = [
@@ -45,7 +45,7 @@ searchItems = {
     "Distribute Mesh" : "MESH",
     "Distribute Spiral" : "SPIRAL",
     "Distribute Spline" : "SPLINE",
-    "Distribute Hexagon" : "HEXAGON",
+    "Distribute Hexagonal Grid" : "HEXAGONAL_GRID",
 }
 
 directionAxisItems = [(axis, axis, "") for axis in ("X", "Y", "Z")]
@@ -96,6 +96,11 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         description = "Increase to have a more accurate evaluation if the type is set to Uniform",
         update = propertyChanged)
 
+    __annotations__["hexGridCenterX"] = BoolProperty(name = "Center Along X", default = True,
+        description = "Center hexagonal grid along the x axis", update = propertyChanged)
+    __annotations__["hexGridCenterY"] = BoolProperty(name = "Center Along Y", default = True,
+        description = "Center hexagonal grid along the y axis", update = propertyChanged)
+
     def create(self):
         if self.mode == "LINEAR":
             self.newInput("Integer", "Amount", "amount", value = 5)
@@ -141,16 +146,12 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             if self.splineDistributionMethod != "VERTICES":
                 self.newInput("Float", "Start", "start", value = 0.0).setRange(0.0, 1.0)
                 self.newInput("Float", "End", "end", value = 1.0).setRange(0.0, 1.0)
-        elif self.mode == "HEXAGON":
-            self.newInput("Integer", "X Divisions", "xDiv", value = 6, minValue = 0)
-            self.newInput("Integer", "Y Divisions", "yDiv", value = 6, minValue = 0)
-            if self.distanceMode == "STEP":
-                self.newInput("Float", "X Distance", "size1", value = 1)
-                self.newInput("Float", "Y Distance", "size2", value = 1)
-            elif self.distanceMode == "SIZE":
-                self.newInput("Float", "Width", "size1", value = 5)
-                self.newInput("Float", "Length", "size2", value = 5)
-            self.newInput("Float", "Z Scale", "size3", value = 1)
+        elif self.mode == "HEXAGONAL_GRID":
+            self.newInput("Integer", "X Divisions", "xDivisions", value = 6, minValue = 0)
+            self.newInput("Integer", "Y Divisions", "yDivisions", value = 6, minValue = 0)
+            self.newInput("Float", "X Size", "xSize", value = 1)
+            self.newInput("Float", "Y Size", "ySize", value = 1)
+            self.newInput("Float", "Z Size", "zSize", value = 1)
 
         self.newOutput("Matrix List", "Matrices", "matrices")
         self.newOutput("Vector List", "Vectors", "vectors")
@@ -158,7 +159,7 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
     def draw(self, layout):
         col = layout.column()
         col.prop(self, "mode", text = "")
-        if self.mode in ("LINEAR", "GRID", "HEXAGON"):
+        if self.mode in ("LINEAR", "GRID"):
             col.prop(self, "distanceMode", text = "")
         if self.mode == "LINEAR":
             layout.prop(self, "directionAxis", expand = True)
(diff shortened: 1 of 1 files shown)


animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -139,14 +141,24 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
                 self.newInput("Float", "End", "end", value = 1.0).setRange(0.0, 1.0)
+        elif self.mode == "HEXAGON":
+            self.newInput("Integer", "X Divisions", "xDiv", value = 6, minValue = 0)
+            self.newInput("Integer", "Y Divisions", "yDiv", value = 6, minValue = 0)
+            if self.distanceMode == "STEP":
+                self.newInput("Float", "X Distance", "size1", value = 1)
+                self.newInput("Float", "Y Distance", "size2", value = 1)
+            elif self.distanceMode == "SIZE":
+                self.newInput("Float", "Width", "size1", value = 5)
+                self.newInput("Float", "Length", "size2", value = 5)
```
OmarEmaraDev, line 152: Use less ambiguous naming here. `xDivisions`, `xDistance`, `Width`, ...

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -139,14 +141,24 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+            self.newInput("Float", "Z Scale", "size3", value = 1)
```
OmarEmaraDev, line 153: I see no reason to keep this here. The user should scale the matrices manually after distribution or supply a properly scaled object. So I think this should be omitted from this node.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -201,6 +213,9 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
             else:
                 yield "matrices = self.execute_SplineVertices(spline)"
```
OmarEmaraDev, line 215: No new line here.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -18,6 +18,7 @@ modeItems = [
     ("MESH", "Mesh", "", "NONE", 3),
     ("SPIRAL", "Spiral", "", "NONE", 4),
     ("SPLINE", "Spline", "", "NONE", 5),
+    ("HEXAGON", "Hexagon", "", "NONE", 6),
```
OmarEmaraDev, line 21: Better call this "Gexagonal Grid" for clarity.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -385,6 +400,34 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         count = len(spline.points)
         return spline.getDistributedMatrices(count, 0, 1, "RESOLUTION")
 
+    def execute_HexagonGrid(self, int xDiv, int yDiv, float size1, float size2, float size3):
+        cdef Py_ssize_t i, j, index
+        cdef float x, y
```
OmarEmaraDev, line 405: Move declarations as close as possible to initializations. So those should be declared just before the loop.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -385,6 +400,34 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
         count = len(spline.points)
         return spline.getDistributedMatrices(count, 0, 1, "RESOLUTION")
 
+    def execute_HexagonGrid(self, int xDiv, int yDiv, float size1, float size2, float size3):
```
OmarEmaraDev, line 403: Same about ambiguous naming here. Simply calling them `xSize` and `ySize` would be clearer.

OmarEmaraDev (changes requested):
Can we add center options along x and y here similar to grid distribution? Don't reuse the properties of the grid distribution, create new ones.

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: Done

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: removed

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: removed

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: Renamed

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: Done

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: Done

harisreedhar:
> Can we add center options along x and y here similar to grid distribution? Don't reuse the properties of the grid distribution, create new ones.

Added center along X and Y option.

animation_nodes/nodes/matrix/distribute_matrices.pyx
OmarEmaraDev: Perhaps you misunderstood my intentions here. I don't think we should scale the matrices in the node. Just like the grid node, where the matrices are always unity scaled regardless of the distribution. It follows that the Z scale input should be removed from the node.

harisreedhar: Ok i'll remove scale from matrix. So, what about ``Step`` and ``Size`` option should i keep or remove it?

OmarEmaraDev: Keep it.

harisreedhar: > Keep it.

Can i reuse ``distanceMode`` from grid ?

OmarEmaraDev: It is better if you don't.

harisreedhar: @OmarEmaraDev I have changed the code.

animation_nodes/nodes/matrix/distribute_matrices.pyx
```diff
@@ -385,6 +411,43 @@ class DistributeMatricesNode(bpy.types.Node, AnimationNode):
...
+        cdef float x, y
+
+        for i in range(xDivisions):
+            x = i * xDistance
+
+            for j in range(yDivisions):
+                if (i % 2) == 0:
+                    y = j * yDistance
+                else:
+                    y = (j + 0.5) * yDistance
```
OmarEmaraDev, line 442: The for loops should be switched. The outer loop should be the one for `yDivisions` and the inner loop should be for `xDivisions`.

There are two reasons why you should do this:

1- y is more expensive to compute and is conditional, so it would make sense to compute it once for every x in the same row.
2. You write the matrices in a row major order. When you have `yDivisions` as the inner loop, you might jump to a different cache line for every single iteration, which means bad cache coherency. While if you have `xDivisions` as the inner loop, you always write matrices sequentially, so you get ideal spacial cache coherency.

While at it, try computing y as follows `y = (j + 0.5 * (i % 2)) * yDistance`. It might give better performance for high number of matrices. Try it maybe.

OmarEmaraDev (changes requested):

animation_nodes/nodes/matrix/distribute_matrices.pyx
harisreedhar: > The for loops should be switched. The outer loop should be the one for `yDivisions` and the inner loop should be for `xDivisions`.
> 
> There are two reasons why you should do this:
> 
> 1- y is more expensive to compute and is conditional, so it would make sense to compute it once for every x in the same row.
> 2. You write the matrices in a row major order. When you have `yDivisions` as the inner loop, you might jump to a different cache line for every single iteration, which means bad cache coherency. While if you have `xDivisions` as the inner loop, you always write matrices sequentially, so you get ideal spacial cache coherency.
> 
> While at it, try computing y as follows `y = (j + 0.5 * (i % 2)) * yDistance`. It might give better performance for high number of matrices. Try it maybe.

Can u provide a pseudocode for this?

OmarEmaraDev:
```python
for j in range(yDivisions):
    for x in range(xDivisions):
        x = i * xDistance
        y = (j + 0.5 * (i % 2)) * yDistance
```

harisreedhar: @OmarEmaraDev Updated

OmarEmaraDev (approved):
Thanks!

PR #1753, opened 2021-07-04
Add FastNoiseSIMD_neon to support arm64 macOS

marcan:
This is all that's needed to make it build and run on arm64 macOS.

### DIFF
--- a/animation_nodes/libs/FastNoiseSIMD/source/compile_macos.sh
+++ b/animation_nodes/libs/FastNoiseSIMD/source/compile_macos.sh
@@ -1,8 +1,17 @@
+#!/bin/sh
+set -e
+
 gcc -c FastNoiseSIMD.cpp -std=c++11 -fPIC -O3
 gcc -c FastNoiseSIMD_internal.cpp -std=c++11 -fPIC -O3
-gcc -c FastNoiseSIMD_sse2.cpp -std=c++11 -fPIC -O3 -msse2
-gcc -c FastNoiseSIMD_sse41.cpp -std=c++11 -fPIC -O3 -msse4.1
-gcc -c FastNoiseSIMD_avx2.cpp -std=c++11 -fPIC -O3 -march=core-avx2
+
+if [ "$(arch)" == "arm64" ]; then
+	gcc -c FastNoiseSIMD_neon.cpp -std=c++11 -fPIC -O3
+else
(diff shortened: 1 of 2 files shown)


animation_nodes/libs/FastNoiseSIMD/source/compile_macos.sh
```diff
@@ -1,8 +1,17 @@
...
-gcc -c FastNoiseSIMD_sse2.cpp -std=c++11 -fPIC -O3 -msse2
-gcc -c FastNoiseSIMD_sse41.cpp -std=c++11 -fPIC -O3 -msse4.1
-gcc -c FastNoiseSIMD_avx2.cpp -std=c++11 -fPIC -O3 -march=core-avx2
+
+if [ "$(arch)" == "arm64" ]; then
+	gcc -c FastNoiseSIMD_neon.cpp -std=c++11 -fPIC -O3
+else
+	gcc -c FastNoiseSIMD_sse2.cpp -std=c++11 -fPIC -O3
+	gcc -c FastNoiseSIMD_sse41.cpp -std=c++11 -fPIC -O3
+	gcc -c FastNoiseSIMD_avx2.cpp -std=c++11 -fPIC -O3
```
OmarEmaraDev, line 12: Why remove the compile flags `-msse2`, `-msse4.1`, and `-mavx2`? Are they not needed? And does Neon require similar flags?

OmarEmaraDev (commented):
Thanks for looking into this! I can't test this myself, so I guess we can just merge this and wait for user feedback.

animation_nodes/libs/FastNoiseSIMD/source/compile_macos.sh
OmarEmaraDev: The MacOS x86 build is failing, so I think those flags are important.

marcan: Sorry, that was some nonsense left from an earlier experiment... I hadn't tested the x86 build! Neon doesn't need any special flags here.

OmarEmaraDev (approved):
Looks good. Thanks!

Github currently doesn't have public runners for MacOS-11, so we will not be able to build for ARM as part of our automated build workflow. Once we are ready for the next release, I may ping you to compile a build so that we can publish it manually, if you are okay with that.

marcan:
Yup, no problem at all. Let me know and I'll put together a build :)

PR #1761, opened 2021-07-27
Add Clamp Falloff node

harisreedhar:
Clamp Falloff node:

clamp_falloff_preview

### DIFF
--- a/animation_nodes/nodes/falloff/clamp_falloff.pyx
+++ b/animation_nodes/nodes/falloff/clamp_falloff.pyx
@@ -1,0 +1,39 @@
+import bpy
+from ... base_types import AnimationNode
+from ... data_structures cimport CompoundFalloff, Falloff
+
+class ClampFalloffNode(bpy.types.Node, AnimationNode):
+    bl_idname = "an_ClampFalloffNode"
+    bl_label = "Clamp Falloff"
+
+    def create(self):
+        self.newInput("Falloff", "Falloff", "inFalloff")
+        self.newInput("Float", "Min", "MinValue", value = 0)
+        self.newInput("Float", "Max", "MaxValue", value = 1)
+        self.newOutput("Falloff", "Falloff", "outFalloff")
+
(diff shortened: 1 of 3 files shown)


animation_nodes/nodes/falloff/clamp_falloff.pyx
```diff
@@ -0,0 +1,39 @@
...
+
+cdef class ClampFalloff(CompoundFalloff):
+    cdef:
+        Falloff falloff
+        float minValue, maxValue
+
+    def __cinit__(self, Falloff falloff, float minValue, float maxValue):
+        self.falloff = falloff
+        self.minValue = minValue
+        self.maxValue = maxValue
```
OmarEmaraDev, line 27: Set the `clamped` value here. `        self.clamped = 0 <= min(minValue, maxValue) <= max(minValue, maxValue) <= 1`

OmarEmaraDev (changes requested):
And remember to update the CHANGELOG.

animation_nodes/nodes/falloff/clamp_falloff.pyx
harisreedhar: @OmarEmaraDev  what it does exactly?

OmarEmaraDev: @harisreedhar When a Falloff declares itself as clamped, it is essentially guaranteeing that the values it will be producing will be in the range `[0, 1]`. This guarantee can be useful to avoid superfluous clamping when there are clamping requirements.

harisreedhar: Ok

OmarEmaraDev (approved):
Thanks! It would be great if you can add that to the documentation and release notes.

harisreedhar:
> Thanks! It would be great if you can add that to the documentation and release notes.

Sure i can do that

PR #1764, opened 2021-07-28
Vectorize Point Distance Falloff node

harisreedhar:
point_distance_vectorzed

### DIFF
--- a/animation_nodes/nodes/falloff/point_distance_falloff.pyx
+++ b/animation_nodes/nodes/falloff/point_distance_falloff.pyx
@@ -1,19 +1,56 @@
 import bpy
+from bpy.props import *
 from ... data_structures cimport BaseFalloff
-from ... base_types import AnimationNode
 from ... math cimport Vector3, setVector3, distanceVec3
+from ... base_types import AnimationNode, VectorizedSocket
+
+mixListTypeItems = [
+    ("MAX", "Max", "", "NONE", 0),
+    ("ADD", "Add", "", "NONE", 1),
+]
 
 class PointDistanceFalloffNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_PointDistanceFalloffNode"
     bl_label = "Point Distance Falloff"
 
+    __annotations__ = {}
+
+    __annotations__["mixListType"] = EnumProperty(name = "Mix List Type", default = "MAX",
+        items = mixListTypeItems, update = AnimationNode.refresh)
+
+    __annotations__["useOriginList"] = VectorizedSocket.newProperty()
+
+    __annotations__["useSizeList"] = VectorizedSocket.newProperty()
+
+    __annotations__["useFalloffWidthList"] = VectorizedSocket.newProperty()
+
     def create(self):
-        self.newInput("Vector", "Origin", "origin")
-        self.newInput("Float", "Size", "size")
-        self.newInput("Float", "Falloff Width", "falloffWidth", value = 4)
+        self.newInput(VectorizedSocket("Vector", "useOriginList",
+            ("Origin", "origin"), ("Origins", "origin")))
+        self.newInput(VectorizedSocket("Float", "useSizeList",
+            ("Size", "size"), ("Sizes", "size")))
+        self.newInput(VectorizedSocket("Float", "useFalloffWidthList",
+            ("Falloff Width", "falloffWidth", dict(value = 4)),
+            ("Falloff Widths", "falloffWidth")))
+
         self.newOutput("Falloff", "Falloff", "falloff")
 
-    def execute(self, origin, size, falloffWidth):
+    def draw(self, layout):
+        if any([self.useOriginList, self.useSizeList, self.useFalloffWidthList]):
+            layout.prop(self, "mixListType", text = "")
+
+    def getExecutionCode(self, required):
+        if any([self.useOriginList, self.useSizeList, self.useFalloffWidthList]):
+            yield "_origins = VirtualVector3DList.create(origin, (0,0,0))"
+            yield "_sizes = VirtualDoubleList.create(size, 0)"
+            yield "_widths = VirtualDoubleList.create(falloffWidth, 0)"
+            yield "amount = VirtualVector3DList.getMaxRealLength(_origins, _sizes, _widths)"
+            yield "falloffs = [self.executeSingle(_origins[i], _sizes[i], _widths[i]) for i in range(amount)]"
+            yield f"falloff = AN.nodes.falloff.mix_falloffs.MixFalloffs(falloffs, '{self.mixListType}')"
+        else:
+            yield "falloff = self.executeSingle(origin, size, falloffWidth)"
+
+    def executeSingle(self, origin, size, falloffWidth):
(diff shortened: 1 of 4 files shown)


OmarEmaraDev:
Is it possible to accelerate the Max option using a KD Tree?

harisreedhar:
@OmarEmaraDev Did you mean something like this? It looks fast

comparison

OmarEmaraDev:
Yes, something like that. The Add operation can't work like this as far as I can tell. How many points until the KD Tree approach becomes faster?

harisreedhar:
@OmarEmaraDev 
Approx 150 points in my testing. For searching higher amount it is faster, but not good for small numbers.

compare_kd_point_distance_vectorize

Here is code:
````
import bpy
from ... math cimport Vector3
from mathutils import Vector, kdtree
from ... data_structures cimport BaseFalloff, Vector3DList

class KD_PointDistanceFalloffNode(bpy.types.Node, AnimationNode):
    bl_idname = "an_KD_PointDistanceFalloffNode"
    bl_label = "KD Point Distance Falloff"

    __annotations__ = {}

    __annotations__["useOriginList"] = VectorizedSocket.newProperty()

    def create(self):
        self.newInput(VectorizedSocket("Vector", "useOriginList",
            ("Origin", "origin"), ("Origins", "origin")))
        self.newInput("Float", "Size", "size")
        self.newInput("Float", "Falloff Width", "falloffWidth", value = 4)
        self.newOutput("Falloff", "Falloff", "falloff")

    def execute(self, origin, size, falloffWidth):
        vectors = origin if self.useOriginList else Vector3DList.fromValue(origin)
        amount = len(vectors)
        kdTree = kdtree.KDTree(amount)
        [kdTree.insert(vector, i) for i, vector in enumerate(vectors)]
        kdTree.balance()
        return PointDistanceFalloff(kdTree, size, falloffWidth)


cdef class PointDistanceFalloff(BaseFalloff):
    cdef:
        kdTree
        float factor
        float minDistance, maxDistance

    def __cinit__(self, kdTree, float size, float falloffWidth):
        self.kdTree = kdTree
        if falloffWidth < 0:
            size += falloffWidth
            falloffWidth = -falloffWidth
        self.minDistance = size
        self.maxDistance = size + falloffWidth

        if self.minDistance == self.maxDistance:
            self.minDistance -= 0.00001
        self.factor = 1 / (self.maxDistance - self.minDistance)

        self.dataType = "LOCATION"
        self.clamped = True

    cdef float evaluate(self, void *value, Py_ssize_t index):
        return calcDistance(self, <Vector3*>value)

    cdef void evaluateList(self, void *values, Py_ssize_t startIndex,
                           Py_ssize_t amount, float *target):
        cdef Py_ssize_t i
        cdef Vector3* vector
        for i in range(amount):
            target[i] = calcDistance(self, <Vector3*>values + i)


cdef inline float calcDistance(PointDistanceFalloff self, Vector3 *v):
    cdef float distance = self.kdTree.find(Vector((v.x, v.y, v.z)))[2]
    if distance <= self.minDistance: return 1
    if distance <= self.maxDistance: return 1 - (distance - self.minDistance) * self.factor
    return 0
````

Also i can't figure out vectorizing ``size`` and ``falloff width`` inputs.

**Comment by harisreedhar** [comment]
Another test:

compare_kd_point_distance_vectorize2

File: animation_nodes/nodes/falloff/point_distance_falloff.pyx [hunk 1]
```diff
@@ -1,19 +1,56 @@
...
+            ("Falloff Widths", "falloffWidth")))
+
         self.newOutput("Falloff", "Falloff", "falloff")
 
-    def execute(self, origin, size, falloffWidth):
+    def draw(self, layout):
+        if any([self.useOriginList, self.useSizeList, self.useFalloffWidthList]):
+            layout.prop(self, "mixListType", text = "")
+
+    def getExecutionCode(self, required):
```
OmarEmaraDev, line 42: Use `getExecutionFunctionName` instead.

animation_nodes/nodes/falloff/point_distance_falloff.pyx
```diff
@@ -1,19 +1,56 @@
...
+        if any([self.useOriginList, self.useSizeList, self.useFalloffWidthList]):
+            yield "_origins = VirtualVector3DList.create(origin, (0,0,0))"
+            yield "_sizes = VirtualDoubleList.create(size, 0)"
+            yield "_widths = VirtualDoubleList.create(falloffWidth, 0)"
```
OmarEmaraDev, line 46: Set the default value to 4 like in the socket.

animation_nodes/nodes/falloff/point_distance_falloff.pyx
```diff
@@ -1,19 +1,56 @@
...
+]
 
 class PointDistanceFalloffNode(bpy.types.Node, AnimationNode):
     bl_idname = "an_PointDistanceFalloffNode"
     bl_label = "Point Distance Falloff"
 
+    __annotations__ = {}
+
+    __annotations__["mixListType"] = EnumProperty(name = "Mix List Type", default = "MAX",
+        items = mixListTypeItems, update = AnimationNode.refresh)
```
OmarEmaraDev, line 19: Is `refresh` really necessary here?

OmarEmaraDev (changes requested):
I am tempted to say just keep the implementation as is. The KD Tree approach is not worth it in this case.

animation_nodes/nodes/falloff/point_distance_falloff.pyx
harisreedhar: @OmarEmaraDev when i used ``propertyChanged`` it dosen't update

OmarEmaraDev: @harisreedhar Update what?

harisreedhar: from changing ``Max`` to ``Add`` it dosen't work

OmarEmaraDev: @harisreedhar Ah. You are embedding/expanding the value of the enum in the execution code `'{self.mixListType}'`, which means the execution code changes and have to be recompiled every time the enum change. In cases where the execution code change, you can use `executionCodeChanged` instead of `propertyChanged`. But you can just write the code such that it doesn't change by simply writing `self.mixListType` without expanding it. But since I suggested we use `getExecutionFunctionName`, this won't be necessary either.

harisreedhar: @OmarEmaraDev  Where should i look to learn how AN execution works?

animation_nodes/nodes/falloff/point_distance_falloff.pyx
harisreedhar: done

animation_nodes/nodes/falloff/point_distance_falloff.pyx
harisreedhar: done

OmarEmaraDev (approved):
Thanks!

Lets write the documentation and release notes for those as well.

PR #1765, opened 2021-07-29
Added isUsed property for Attribute Output node

harisreedhar:


OmarEmaraDev:
What is the rational behind the change?

By the way, don't commit your changes to master. Create a new branch and commit it there. Your master should be identical to upstream master.

harisreedhar:
> What is the rational behind the change?

Sometimes we only want to set attribute only once, not every time. which saves some performance.
For example setting object's display type to bounds, which doesn't requires to execute every time. but only once.
There are many other cases as well like setting rigidbody mass, collision shape etc.

harisreedhar:
> By the way, don't commit your changes to master. Create a new branch and commit it there. Your master should be identical to upstream master.

Ok i will take care of that, do i need to create new pull request from new branch?

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@
 - Remove clamping from the inputs of the *Random Falloff* node.
 - Support FLIP particle systems in the *Particles Data* node.
 - Vectorized *Point Distance Falloff* node.
+- Added *Attribute Output* node enable/disable property for value socket.
```
OmarEmaraDev, line 62: I think a clearer statement would be:

```
Added *Is Used* property to the value socket in the *Attribute Output* node.
```

animation_nodes/nodes/object/object_attribute_output.py
```diff
@@ -44,28 +46,29 @@ def getExecutionCode(self, required):
             yield "self.setErrorMessage('Invalid Syntax', show = len(self.attribute.strip()) > 0)"
             return
 
-        yield "try:"
-        if self.useObjectList:
-            if self.useValueList:
-                yield "    _values = [None] if len(values) == 0 else values"
-                yield "    _values = itertools.cycle(_values)"
-                yield "    for object, value in zip(objects, _values):"
+        if self.inputs[1].isUsed:
```
OmarEmaraDev, line 49: Use a guard condition here.
```
if not self.inputs[1].isUsed: return
```

OmarEmaraDev (changes requested):
Okay. Sounds reasonable.

You don't have to recreate the pull request or the branch now, just make sure to do it in the future.

CHANGELOG.md
harisreedhar: ok

harisreedhar:
@OmarEmaraDev Do i need to add a release note?

OmarEmaraDev (approved):
Thanks!

Yes, just add an entry like this https://docs.animation-nodes.com/release_notes/v2_2/object/.

PR #1769, opened 2021-08-01
Add Scene Changed auto execution option and make it the default

OmarEmaraDev:
This patch adds another auto execution option *Scene Changed*. This option executes the node tree when and if and only if anything changes in the scene. Additionally, the option is now enabled by default and the *Always* option is now disabled by default.

Having *Always* default to True was always questionable as it is very inefficient and mostly superfluous. However, the concern was that *Always* is the most expected execution behavior of Animation Nodes for users, especially for novices, because all changes are immediately reflected to the scene without worrying about the ideal execution behavior.

The *Scene Changed* option is, I believe, the best compromise we can get. Because it only executes when something actually changes in the scene, and at the same time, it is inclusive to all events that the user would expect to trigger an execution.

---

I would like to refactor other execution behaviors to supplement this change, but we shall look at those refactors separately later.

### DIFF
--- a/animation_nodes/base_types/node_tree.py
+++ b/animation_nodes/base_types/node_tree.py
@@ -63,7 +63,9 @@ class AnimationNodeTree(bpy.types.NodeTree):
         if not a.enabled: return False
         if not self.hasMainExecutionUnits: return False
 
-        if "Frame" in events and (a.sceneUpdate or a.frameChanged): return True
+        if "Scene" in events and (a.always or a.sceneChanged): return True
+
+        if "Frame" in events and (a.always or a.sceneChanged or a.frameChanged): return True
 
         if not (isRendering() or isInterfaceLocked()):
             if self.timeSinceLastAutoExecution < a.minTimeDifference: return False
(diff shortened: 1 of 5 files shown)


Durman:
I'm curious which events are included in *Scene change*? Is it triggered by change in some node tree (AN, material, compositor), moving an object in object mode, editing an object, the same via python API?

OmarEmaraDev:
@Durman Everything you mentioned and more. It really just executes when **anything** changes. You can try it yourself by downloading the "Artifacts" zipt file from [here](https://github.com/JacquesLucke/animation_nodes/pull/1769/checks?check_run_id=3213171502).

JacquesLucke:
I wonder, how does this change existing files? Will files that used Always before, have nothing selected now? If that's the case, it might make sense to add some minimum versioning code somehow.

Overall I'm fine with this change. Some background: Originally, I used the scene update handler, but due to unknown reasons this handler was always executed. We only got a "real" scene changed handler reeelatively recently, using it sounds like a good idea.

Durman:
The hanlder reacts on changes of a node color but is not triggered upon link creation. The handler is still strange.

OmarEmaraDev:
@Durman Are you talking about the handler itself or this particular pull request?

Durman:
About the handler. I was thinking to use it in Sverchok but it looks like is quite problematic. Mesh viewer node updates a scene this triggers the handler and this triggers updating mesh viewer again and updates get into infinite loop. Probably some handler events could be ignored but I'm not sure what will be an elegant solution for that.

OmarEmaraDev:
@Durman As far as I know, depsgraph updates are executed in two phases. Where the first phase executes the python handlers and the second phase updates the scene based on the possible changes by the handlers, without executing the handlers. So recursive execution shouldn't be a problem when using such a handler. Therefore, I think maybe this is related to Sverchok instead.

Durman:
Oh thanks for the idea. It definitely should work in this way. It seems the thing why it does not in Sverchok is because all updates happen via a timer there. Probably for this type of update I have to make an exception and update a tree during the handler call.

OmarEmaraDev:
Merged in 69132fe.

PR #1791, opened 2021-09-24
Add Vertex Group as attribute for Mesh object

3DSinghVFX:
I have added *Vertex Group* as an attribute for the *Mesh* object. I have also added the *Get Vertex Group* and *Insert Vertex Group* nodes.


OmarEmaraDev:
Vertex groups in Blender are part of objects not meshes?

3DSinghVFX:
> Vertex groups in Blender are part of objects not meshes?

Okay. But we can access that in Geometry Nodes: 
Screenshot from 2021-09-27 14-26-55

OmarEmaraDev:
@3DSinghVFX It would still be accessible through the set and input vertex group nodes. The question is whether we really want it as part of the mesh structure.

Vertex groups seems to be supported in only three object types, Meshes, Lattices, and Grease Pencil objects. And I just realized that our vertex group nodes only support meshes, so it would be good to extend them to the rest of the types.

Actually looking at the source code, vertex groups are stored in meshes, not objects, they are just exposed in RNA through objects. So this patch looks fine design wise. We will also have the advantage of including vertex group in mesh processing for "propagation" of attribute data.

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 
 - Added support for splines in *3D Viewer* node.
 - Added built-in attributes for Mesh object.
-
+- Added Vertex Group as attribute for Mesh object.
```
OmarEmaraDev, line 7: Keep a new line between contents and headings. Some markdown processors only work with that new line.

animation_nodes/nodes/mesh/get_vertex_group.py
```diff
@@ -0,0 +1,26 @@
...
+    bl_idname = "an_GetVertexGroupNode"
+    bl_label = "Get Vertex Group"
+    errorHandlingType = "EXCEPTION"
+
+    def create(self):
+        self.newInput("Mesh", "Mesh", "mesh")
+        self.newInput("Text", "Name", "groupName")
+        self.newOutput("Float List", "Weights", "weights")
+
+    def execute(self, mesh, attributeName):
```
OmarEmaraDev, line 15: Don't use the generic attribute name, use group instead. `attributeName -> vertexWeightName`, `Vertex group name can't be empty. -> Group name can't be empty.`, `...`.

animation_nodes/nodes/mesh/mesh_object_input.py
```diff
@@ -15,6 +15,7 @@ def create(self):
         self.newInput("Boolean", "Use Modifiers", "useModifiers", value = False)
         self.newInput("Boolean", "Load UVs", "loadUVs", value = False, hide = True)
         self.newInput("Boolean", "Load Vertex Colors", "loadVertexColors", value = False, hide = True)
+        self.newInput("Boolean", "Load Vertex Weights", "loadVertexWeights", value = False, hide = True)
         self.newInput("Boolean", "Load Custom Attributes", "loadCustomAttributes", value = False, hide = True)
```
OmarEmaraDev, line 19: Not related to this patch, but we should consider making those into node options instead.

animation_nodes/nodes/mesh/mesh_object_output.py
```diff
@@ -151,6 +151,18 @@ def setMesh(self, outMesh, mesh):
         for attribute in mesh.iterVertexColorAttributes():
             vertexColorLayer = outMesh.vertex_colors.new(name = attribute.name, do_init = False)
             vertexColorLayer.data.foreach_set("color", attribute.data.asMemoryView())
+        
+        # Vertex Groups
+        vertexAmount = len(outMesh.vertices)
+        for attribute in mesh.iterVertexWeightAttributes():
+            try:
```
OmarEmaraDev, line 158: Use explicit checks instead of a try except.

animation_nodes/nodes/mesh/mesh_object_output.py
```diff
@@ -151,6 +151,18 @@ def setMesh(self, outMesh, mesh):
...
+                vertexGroup = object.vertex_groups[attribute.name]
+            except:
+                vertexGroup = object.vertex_groups.new(name = attribute.name)
+            weights = attribute.data
+            for i in range(vertexAmount):
+                vertexGroup.add([i], weights[i], "REPLACE")
```
OmarEmaraDev, line 164: Would `vertexGroup.add(list(range(vertexAmound)), weights, "REPLACE")` be faster?

animation_nodes/nodes/mesh/mesh_object_input.py
```diff
@@ -195,3 +213,32 @@ def loadCustomAttributes(self, mesh, sourceMesh, object):
...
+        vertexAmount = len(object.data.vertices)
+        weights = DoubleList(length = vertexAmount)
+        getWeight = vertexGroup.weight
+
+        for i in range(vertexAmount):
+            try: weights[i] = getWeight(i)
+            except: weights[i] = 0
+        return weights
+
+    def execute_All_WithModifiers(self, object, vertexGroup, scene):
```
OmarEmaraDev, line 227: Just remove this for now. Seems like a Blender limitation.

OmarEmaraDev (changes requested):

3DSinghVFX:
> @3DSinghVFX It would still be accessible through the set and input vertex group nodes. The question is whether we really want it as part of the mesh structure.

I think it can provide a similar workflow as we have in Geometry Nodes. 

> Vertex groups seems to be supported in only three object types, Meshes, Lattices, and Grease Pencil objects. And I just realized that our vertex group nodes only support meshes, so it would be good to extend them to the rest of the types.

Yes. In later versions of Blender, Group was not exposed in Grease Pencil. But, it is possible in Blender 3.0.

> Actually looking at the source code, vertex groups are stored in meshes, not objects, they are just exposed in RNA through objects. So this patch looks fine design wise. We will also have the advantage of including vertex group in mesh processing for "propagation" of attribute data.

That is the idea.

CHANGELOG.md
3DSinghVFX: Okay.

animation_nodes/nodes/mesh/mesh_object_input.py
3DSinghVFX: That will make the node more cleaner, and also be good for more options in the future.

animation_nodes/nodes/mesh/get_vertex_group.py
3DSinghVFX: Omar, we also have *Face Groups*, so we should use the full name *Vertex Group*, isn't it?

animation_nodes/nodes/mesh/mesh_object_output.py
3DSinghVFX: That is not possible, Vertex Weight does not have `foreach_set/foreach_get` functionality.

animation_nodes/nodes/mesh/get_vertex_group.py
OmarEmaraDev: Yes, my comment above was typo, make it vertex group.

animation_nodes/nodes/mesh/mesh_object_output.py
OmarEmaraDev: But `add` can add multiple elements together, can't it? So the above statement should work?

3DSinghVFX: It can set the single weight to multiple vertices, however, it does not set weights to individual vertices.
If I add that line, shows this error: `TypeError: VertexGroup.add(): error with argument 2, "weight" -  Function.weight expected a float type`.

animation_nodes/nodes/mesh/mesh_object_input.py
3DSinghVFX: But how we are going to handle the data length? Or should we just add zero weight?

animation_nodes/nodes/mesh/mesh_object_output.py
OmarEmaraDev: I see. Ignore this comment then.

animation_nodes/nodes/mesh/mesh_object_input.py
OmarEmaraDev: For now return zero weights. We can look into this case in more details later.

OmarEmaraDev (approved):
Thanks!

PR #1802, opened 2021-10-31
Fix float list conversion in Insert Vertex Group node

3DSinghVFX:
I have fixed the float list conversion in the Insert Vertex Group node.


OmarEmaraDev (approved):
Thanks! Make sure to update the CHANGELOG.

PR #1812, opened 2021-12-09
Use real Action class

### LINKED ISSUE #1810: All Action conversions fail with NameError
## Description
When attempting to convert an Action socket, the Node Tree will error with the message `NameError: name 'Action' is not defined`

## To reproduce
1. Create an Empty and add an Action to it
1. Create an `Action from Object` node in the Animation Nodes editor
1. Attach the `Action` output socket to a `Convert` node
1. Change the conversion type to `Action`
1. `NameError` occurs

image

I tried to work around this using a `Script` node and some custom class conversion with Python, but `Script` nodes seem to silently fail to output a valid `Action` even when the NodeTree displays no errors and the provided output variable is a valid instance of `data_structures.Action`.

2021-12-08_23-56-03_%pn

## Potential fix
Changing [`ActionSocket.correctValue()`](https://github.com/JacquesLucke/animation_nodes/blob/4efcf03272c14ac0078bb3d4564a27fe25b1eef3/animation_nodes/sockets/action.py#L22) to use `data_structures.Action` fixes both of the explicit and silent errors listed above, but a small caveat is that the `Convert` node still cannot automatically convert `bpy.types.Action` to a format animation_nodes understands. 

With fix:
2021-12-08_23-54-41_%pn

## Example file
Here is a [minimal blend file](https://dl.dropboxusercontent.com/s/jybn9grb8z9d88q/action%20selector%20script%20minimal.blend) which reproduces this bug. Three scenarios are included:
1. Converting an `Action` socket from an `Action from Object` node (fixed with above solution)
2. Converting a `bpy.types.Action` generic attribute from an `Object Attribute Input` node (conversion fails with above solution)
3. Converting an action using custom python scripting (fixed with above solution)

image

Th3Gavst3r:
Resolves #1810

### DIFF
--- a/animation_nodes/sockets/action.py
+++ b/animation_nodes/sockets/action.py
@@ -1,4 +1,5 @@
 import bpy
+from .. data_structures import Action
 from .. base_types import AnimationNodeSocket, PythonListSocket
 
 class ActionSocket(bpy.types.NodeSocket, AnimationNodeSocket):
@@ -18,7 +19,6 @@ class ActionSocket(bpy.types.NodeSocket, AnimationNodeSocket):
 
     @classmethod
     def correctValue(cls, value):
-        # TODO: use real Action class
         if isinstance(value, Action):
             return value, 0
         return cls.getDefaultValue(), 2


OmarEmaraDev (approved):
Thanks!

Issue #1813, opened 2021-12-09
Object Action Output nodes fail with SyntaxError for Action inputs with named attributes


Th3Gavst3r:
## Description
When an `Action` contains keyframes for a path with a user-named component (such as keyframes on a Custom Property), inputting it to an `Object Action Output` results in a `SyntaxError`.

## To reproduce
1. Create an Empty and give it a Custom Property with any name.
2. Create an Action on your Empty and insert a keyframe for your Custom Property.
3. Open the Animation Nodes editor and create a new NodeTree.
4. Add an `Action from Object` node and select your Empty for the input object.
5. Add an `Object Action Output` node and plug in the Action from your `Action from Object` node.
6. `SyntaxError` occurs.
```
  File "...\animation_nodes\utils\attributes.py", line 27, in getMultiAttibuteSetter
    exec(code, variables)
  File "<string>", line 2
    owner.["prop"] = values[0]
          ^
SyntaxError: invalid syntax
```

Here is a [minimal blend file](https://github.com/JacquesLucke/animation_nodes/files/7688797/animation_nodes.named.property.minimal.example.zip) reproducing the error.

PR #1816, opened 2021-12-10
Fix custom property assignment types

### LINKED ISSUE #1815: Actions with Custom Property curves always use Float values
## Description
After support for Custom Properties was added to `Actions` in #1814, the `Object Action Output` node always overwrites Custom Property types with `Float`. Doing so removes other configuration items from the Custom Property, such as `Min` and `Max` values.

image

## To reproduce
1. Create a Cube and add a Custom Property with type `Integer`.
2. Add a keyframe for your Integer property.
3. In the Animation Nodes editor, create an `Action from Object` node and select your Cube as the source object.
4. Create an `Object Action Output` node and select your Cube for the output object as well.
5. Connect the Action sockets between the two nodes.
6. Your Integer property becomes a Float.
7. Disconnect the Action sockets.
8. Select the gear icon next to your custom property and change the type back to Integer.
9. Some settings may have been destroyed, including `Mix`/`Max` values and the `Description`.

Here is a [minimal blend file](https://github.com/JacquesLucke/animation_nodes/files/7695777/animation_nodes.custom.property.type.reassignment.from.fcurves.zip) reproducing the error.

https://user-images.githubusercontent.com/9059018/145649900-5664a87e-0fc9-4151-a06d-4b06a64c4283.mp4

Th3Gavst3r:
Modifies the `Object Action Output` node's attribute setter function to perform a type cast.

## Notes
* `object.bl_rna.properties` does not contain Custom Properties, but copies of items in `bl_rna.properties` seem to exist directly on the `object` (such as `object.bl_rna.properties[location]` and `object.location`).
* Unlike `bl_rna.properties`, properties accessed directly from the `object` do not have an `array_length` attribute, so `len()` was used instead.

Resolves #1815

### DIFF
--- a/animation_nodes/utils/attributes.py
+++ b/animation_nodes/utils/attributes.py
@@ -35,6 +35,8 @@ def getAttributeSetterLines(objectName, propName, valueName):
     yield f"try: {propPath} = type({propPath})({valueName})"
     # Property does not exist; default to new float
     yield f"except: {propPath} = {valueName}"
+    # Force dependency graph update. See https://developer.blender.org/T63793#881438
+    yield f"{objectName}.update_tag()"
 
 def getPropertyPath(objectName, propName):
     if propName.startswith("["):
--- a/animation_nodes/utils/attributes.py
+++ b/animation_nodes/utils/attributes.py
@@ -21,23 +21,19 @@ def getMultiAttibuteSetter(propNames):
     code = "def setter(owner, values):\n"
     for i, prop in enumerate(propNames):
         lines = getAttributeSetterLines("owner", prop, "values[{}]".format(i))
-        code += ''.join("    " + line + "\n" for line in lines)
+        code += "".join("    " + line + "\n" for line in lines)
     code += "    pass"
     variables = {}
     exec(code, variables)
     return variables["setter"]
 
 def getAttributeSetterLines(objectName, propName, valueName):
-    if propName.startswith('['):
-        # Path is a named attribute
-        path = "{}{}".format(objectName, propName)
-    else:
-        path = "{}.{}".format(objectName, propName)
+    propPath = getPropertyPath(objectName, propName)
+    # Try to cast to existing property type
+    yield f"try: {propPath} = type({propPath})({valueName})"
+    # Property does not exist; default to new float
+    yield f"except: {propPath} = {valueName}"
 
-    return (
-        'try: {} = type({})({}) # Try to cast to existing property type'.format(path, path, valueName),
-        'except: {} = {} # Property does not exist; default to new float'.format(path, valueName)
-    )
 
 def hasEvaluableRepr(value):
     try: return eval(repr(value)) == value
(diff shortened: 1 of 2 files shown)


OmarEmaraDev:
Sorry for the delay, I have a lot on my plate but and will get to this as soon as possible.

animation_nodes/utils/attributes.py
```diff
@@ -20,19 +20,24 @@ def getAttributeGetter(propName):
...
+def getAttributeSetterLines(objectName, propName, valueName):
+    if propName.startswith('['):
         # Path is a named attribute
-        return '{}{} = {}'.format(objectName, propName, valueName)
+        path = "{}{}".format(objectName, propName)
     else:
-        return '{}.{} = {}'.format(objectName, propName, valueName)
+        path = "{}.{}".format(objectName, propName)
+
+    return (
```
OmarEmaraDev, line 37: Use `yield` instead of return. Double quotes instead of singles quotes. Comments go above statements. Use f-strings.

```python
# Try to cast to existing property type.
yield f"try: {path} = type({path})({valuename})"
# Property does not exist, default to new float.
yield f"except: {path} = {valueName}"
```

animation_nodes/utils/attributes.py
```diff
@@ -20,19 +20,24 @@ def getAttributeGetter(propName):
 def getMultiAttibuteSetter(propNames):
     code = "def setter(owner, values):\n"
     for i, prop in enumerate(propNames):
-        line = getAttributeSetterLine("owner", prop, "values[{}]".format(i))
-        code += "    " + line + "\n"
+        lines = getAttributeSetterLines("owner", prop, "values[{}]".format(i))
+        code += ''.join("    " + line + "\n" for line in lines)
```
OmarEmaraDev, line 24: Double quotes.

animation_nodes/utils/attributes.py
```diff
@@ -48,16 +53,14 @@ def pathBelongsToArray(object, dataPath):
...
-    else:
-        pathToProperty = ""
-        propertyName = dataPath
+    if not dataPath.startswith("[") and not dataPath.startswith("."):
+        dataPath = "." + dataPath
 
     try:
-        amount = eval("object{}.bl_rna.properties[{}].array_length".format(pathToProperty, repr(propertyName)))
-        return amount > 0
+        data = eval("object{}".format(dataPath))
```
OmarEmaraDev, line 60: Use f-strings.

animation_nodes/utils/attributes.py
```diff
@@ -48,16 +53,14 @@ def pathBelongsToArray(object, dataPath):
...
+        if isinstance(data, str):
+            return False # Strings have len() but aren't arrays
```
OmarEmaraDev, line 62: Comments go above statements.

OmarEmaraDev (changes requested):
Looks generally fine. Just some style comments to follow AN's code style.

Th3Gavst3r:
Sorry for the delay on my end this time, lots of holidays lately.

Going from your examples I fixed up some other `f` strings in `attributes.py`, which made made realize several other functions were using the `object.` notation that caused #1813. I haven't had those functions cause me problems yet, but I went ahead and modified them.

I also added an `object.update_tag()` to the setters because Blender has a [known issue](https://developer.blender.org/T63793#881438) where custom properties when controlled by Python don't trigger updates to the dependency graph. This means drivers controlled by the value won't work unless you tag the object for updates.

OmarEmaraDev (approved):
Looks good. Thanks!

Issue #1811, opened 2021-12-09
Add Velocity to Evaluate MIDI Track Node


fabricatedmath:
It would be great to have

OmarEmaraDev:
Maybe we should return the index of the note instead for more flexibility.

fabricatedmath:
That might work better idk, maybe the lookups will be slow?

Maybe something to keep in mind also is the new MIDI MPE standard. Typically, as you might know, pitch bend applies to the entire channel (that wobbly thing on the side of a keyboard piano). MPE has a separate note per channel, so you can get pitch bend, after touch, etc... for each note individually. 

Maybe knowing there's eventually going to be more dimensions to each midi note, it might be easier to return a "MIDI Note object" in "Evaluate MIDI Track"  instead of a note value?

fabricatedmath:
Hey, it's been a while, but been pretty busy, finally getting settled to get started on these two features. 

What do you think about having an option to have velocity multiplied into the note value? The issue would be that if you only have velocity from the node, then what do you do with velocity output as the note persists? 

Also might be more friendly to the overlapping envelopes part. You could have a really strong heavy note followed by a light note and the state from the heavy note could still dominate the note value. If it wasn't multiplied in, then you'd be forced to multiply in a small velocity value to the current note value accumulation.

What do you think?

OmarEmaraDev:
@fabricatedmath Just to clarify:
- Is the velocity of a note generally associated with the volume at which its sound is played? And consequently associated with the levels of the note's envelope?
- Regarding #1820, two notes are said to overlap if the release interval of the first overlaps with the attack interval of the second? In this case, we have two options. 1) We can re-trigger by starting from zero as soon as we enter the attack interval of the second note. 2) We can take the maximum of the two intervals such that the release is continued until the attack ramps up to the same level. Is that correct?

If both of those are correct, then yes, I think adding an option to multiply the note values with the attack is reasonable.

fabricatedmath:
I did a bit more digging realizing its probably more complicated. Apple does a good writeup here: https://support.apple.com/guide/logicpro/use-sampler-envelopes-lgsifc8612a5/mac

> 
> Vel slider: Set the sensitivity of amplitude envelope modulation in response to incoming velocity data.
> If the Vel slider is set to zero, the envelope outputs its maximum level when you strike the keys at any velocity.
> 
> At a Vel slider value of 100%, the entire dynamic range is under velocity control.
> 
[...]

PR #1832, opened 2022-03-13
Added ADSR envelope with velocity sensitivity

fabricatedmath:
Github won't let me attach a midi file...

But I set the frame rate to 30 fps, attack to 0.1, decay to 0.1, sustain to its default 0.6 and release to 0.1, First note in my file plays for 0.5s, so attack takes 3 frames, decay takes 3 frames and at frame 15, release happens for 3 frames and at frame 18 its back to 0. Seems to work for Single mode and All mode.

Velocity sensitivity seems to be working as well.

Let me know what you think!

### DIFF
--- a/animation_nodes/data_structures/midi/midi_note.py
+++ b/animation_nodes/data_structures/midi/midi_note.py
@@ -8,18 +8,34 @@ class MIDINote:
     timeOff: float = 0
     velocity: float = 0
 
-    def evaluate(self, time, attackTime, attackInterpolation, releaseTime, releaseInterpolation):
-        peakTime = self.timeOn + attackTime
-        endTime = self.timeOff + releaseTime
-        if self.timeOff >= time >= peakTime:
-            return 1
-        elif peakTime > time >= self.timeOn:
-            if attackTime == 0: return 1
-            return attackInterpolation((time - self.timeOn) / attackTime)
-        elif endTime >= time > self.timeOff:
-            if releaseTime == 0: return 1
-            return releaseInterpolation(1 - ((time - self.timeOff) / releaseTime))
-        return 0.0
+    def evaluate(self, time, attackTime, attackInterpolation, decayTime, decayInterpolation, sustainLevel, 
+        releaseTime, releaseInterpolation, velocitySensitivity):
+
+        def valueInEnvelope():
+            # find either point in time for envelope or where in envelope the timeOff happened
+            relativeTime = min(time, self.timeOff) - self.timeOn
+
+            if relativeTime <= 0.0:
+                return 0.0
+
+            if relativeTime < attackTime:
+                return attackInterpolation(relativeTime / attackTime)
+
+            relativeTime = relativeTime - attackTime
+
+            if relativeTime < decayTime:
+                decayNormalized = decayInterpolation(1 - relativeTime/ decayTime)
+                return decayNormalized * (1 - sustainLevel) + sustainLevel
+
+            return sustainLevel
+
+        value = valueInEnvelope()
+
+        if time > self.timeOff:
+            value = value * releaseInterpolation(1 - ((time - self.timeOff) / releaseTime))
+
+        # if velocity sensitivity is 25%, then take 75% of envelope and 25% of envelope with velocity
+        return (1 - velocitySensitivity) * value + velocitySensitivity * self.velocity * value
 
     def copy(self):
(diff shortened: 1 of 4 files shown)


fabricatedmath:
referencing https://github.com/JacquesLucke/animation_nodes/issues/1811

animation_nodes/data_structures/midi/midi_note.py
```diff
@@ -8,18 +8,34 @@ class MIDINote:
...
+
+            if relativeTime < decayTime:
+                decayNormalized = decayInterpolation(1 - relativeTime/ decayTime)
+                return decayNormalized * (1 - sustainLevel) + sustainLevel
+
+            return sustainLevel
+
+        value = valueInEnvelope()
+
+        if time > self.timeOff:
```
OmarEmaraDev, line 34: I see no reason to put the release case outside like this. Since the values is guaranteed to be `sustainLevel`, just do `return sustainLevel * releaseInterpolation(...)` inside.

animation_nodes/data_structures/midi/midi_note.py
```diff
@@ -8,18 +8,34 @@ class MIDINote:
...
-            if attackTime == 0: return 1
-            return attackInterpolation((time - self.timeOn) / attackTime)
-        elif endTime >= time > self.timeOff:
-            if releaseTime == 0: return 1
-            return releaseInterpolation(1 - ((time - self.timeOff) / releaseTime))
-        return 0.0
+    def evaluate(self, time, attackTime, attackInterpolation, decayTime, decayInterpolation, sustainLevel, 
+        releaseTime, releaseInterpolation, velocitySensitivity):
+
+        def valueInEnvelope():
```
OmarEmaraDev, line 14: Prefer not to have local functions. Just pull this function outside into a separate method and call it `evaluateEnvelop`.

OmarEmaraDev (changes requested):
Looks generally good, just small nitpicks. Make sure to add your change to the CHANGELOG file as well.

animation_nodes/data_structures/midi/midi_note.py
fabricatedmath: Hmm, I don't think that works, because the time off could have taken place at any of the phases in the envelope. The function only returns the point in time in the envelope or the point that a timeOff took place.

animation_nodes/nodes/sound/evaluate_midi_track.py
```diff
@@ -26,9 +26,14 @@ def create(self):
         self.newInput("Float", "Attack Time", "attackTime", value = 0.01)
         self.newInput("Interpolation", "Attack Interpolation",
                 "attackInterpolation", defaultDrawType = "PROPERTY_ONLY")
+        self.newInput("Float", "Decay Time", "decayTime", value = 0.01)
+        self.newInput("Interpolation", "Decay Interpolation",
+                "decayInterpolation", defaultDrawType = "PROPERTY_ONLY")
+        self.newInput("Float", "Sustain Level", "sustainLevel", value = 0.6).setRange(0,1)
```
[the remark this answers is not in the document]
fabricatedmath: I added a min(max(val,0),1) to the yield string, hope thats what you meant

animation_nodes/data_structures/midi/midi_note.py
fabricatedmath: done

fabricatedmath:
If these changes work for you, is it too much trouble to ask you to do a build with py3.10? I had no trouble getting this working on linux, but this windows stuff is a nightmare

fabricatedmath:
Never mind about the build, I got it working, something about defaulting to 32bit even when using the x64 developer console, idfk...

animation_nodes/data_structures/midi/midi_note.py
OmarEmaraDev: Right, my bad.

OmarEmaraDev (approved):
Looks good now. Thanks!

OmarEmaraDev:
@fabricatedmath For the future, builds for all platforms for any pull request can be download from its build artifacts page.
https://github.com/JacquesLucke/animation_nodes/pull/1832/checks

fabricatedmath:
> 

Oh wow good to know, thanks! All the stuff I downloaded and installed to build this on windows feels like malware

PR #1834, opened 2022-03-16
Vectorized the Get Spline Length node.

3DSinghVFX:
I have vectorized the **Get Spline Length** node.

### DIFF
--- a/animation_nodes/nodes/spline/get_partial_length.py
+++ b/animation_nodes/nodes/spline/get_partial_length.py
@@ -1,16 +1,32 @@
 import bpy
-from ... base_types import AnimationNode
 from . spline_evaluation_base import SplineEvaluationBase
+from ... base_types import AnimationNode, VectorizedSocket
+from ... data_structures import VirtualDoubleList, DoubleList
 
 class GetSplineLengthNode(bpy.types.Node, AnimationNode, SplineEvaluationBase):
     bl_idname = "an_GetSplineLengthNode"
     bl_label = "Get Spline Length"
 
+    useSplineList: VectorizedSocket.newProperty()
+    useStartList: VectorizedSocket.newProperty()
+    useEndList: VectorizedSocket.newProperty()
+
     def create(self):
-        self.newInput("Spline", "Spline", "spline", defaultDrawType = "PROPERTY_ONLY")
-        self.newInput("Float", "Start", "start").setRange(0, 1)
-        self.newInput("Float", "End", "end", value = 1.0).setRange(0, 1)
-        self.newOutput("Float", "Length", "length")
+        socket = self.newInput(VectorizedSocket("Spline", "useSplineList",
+            ("Spline", "spline"), ("Splines", "splines")))
+        socket.defaultDrawType = "PROPERTY_ONLY"
+        socket.dataIsModified = True
+
+        self.newInput(VectorizedSocket("Float", "useStartList",
+            ("Start", "start"),
+            ("Starts", "starts")))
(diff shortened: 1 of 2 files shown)


OmarEmaraDev (changes requested):
Can't all the vectorization variants be a single methods?
If any of the inputs is vectorized, create a virtual list from all three of them and run the loop over the maximum length like we do for the rest of the node.

3DSinghVFX:
> Can't all the vectorization variants be a single methods? 

I think that is possible. Is it possible to create a virtual list for spline? and can I use the multi-virtual here?

OmarEmaraDev:
@3DSinghVFX Yes. See `VirtualPyList`. And yes, you can use `createMultiple` for the other two.

3DSinghVFX:
> @3DSinghVFX Yes. See `VirtualPyList`. And yes, you can use `createMultiple` for the other two.

Thanks.

animation_nodes/nodes/spline/get_partial_length.py
```diff
@@ -20,7 +36,23 @@ def drawAdvanced(self, layout):
...
 
-    def execute(self, spline, start, end):
+    def getExecutionFunctionName(self):
+        if any([self.useSplineList, self.useStartList, self.useEndList]):
+            return "execute_MultipleSplines_MultipleStarts_MultipleEnds"
+        else:
+            return "execute_SingleSpline_SingleStart_SingleEnd"
+
+    def execute_MultipleSplines_MultipleStarts_MultipleEnds(self, spline, start, end):
+        splines = VirtualPyList.create(spline, None)
```
OmarEmaraDev, line 46: The default value should be an empty spline, otherwise you get an error.

animation_nodes/nodes/spline/get_partial_length.py
```diff
@@ -1,16 +1,32 @@
...
+
     def create(self):
-        self.newInput("Spline", "Spline", "spline", defaultDrawType = "PROPERTY_ONLY")
-        self.newInput("Float", "Start", "start").setRange(0, 1)
-        self.newInput("Float", "End", "end", value = 1.0).setRange(0, 1)
-        self.newOutput("Float", "Length", "length")
+        socket = self.newInput(VectorizedSocket("Spline", "useSplineList",
+            ("Spline", "spline"), ("Splines", "spline")))
+        socket.defaultDrawType = "PROPERTY_ONLY"
+        socket.dataIsModified = True
```
OmarEmaraDev, line 18: Why is this set to True?

OmarEmaraDev (changes requested):

animation_nodes/nodes/spline/get_partial_length.py
3DSinghVFX: Okay. So which one use here Poly or Bezier?

animation_nodes/nodes/spline/get_partial_length.py
3DSinghVFX: Sorry. I'll remove it.

animation_nodes/nodes/spline/get_partial_length.py
OmarEmaraDev: @3DSinghVFX  It wouldn't matter in this case, since isEvaluable will be false. So just create a Poly one.

3DSinghVFX: Okay.

OmarEmaraDev (approved):

PR #1839, opened 2022-03-26
Fix the API changes of gpencil nodes.

3DSinghVFX:
I have fixed the API changes for Stroke Line Width, Frame Number, and Change Line, now they have data type `int`.

### DIFF
--- a/animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
+++ b/animation_nodes/nodes/gpencil/gp_frame_from_strokes.py
@@ -16,7 +16,7 @@ class GPFrameFromStrokesNode(bpy.types.Node, AnimationNode):
     def create(self):
         self.newInput(VectorizedSocket("GPStroke", "useStrokeList",
             ("Stroke", "stroke"), ("Strokes", "strokes")), dataIsModified = True)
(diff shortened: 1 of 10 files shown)


OmarEmaraDev (approved):

3DSinghVFX:
Should we change the name `Stroke Thickness > Offset Line Width or Change Line Width`? This option basically offsets the line width of the strokes.

OmarEmaraDev:
@3DSinghVFX I am not familiar with the option, can you demonstrate that for me or point me in the right direction?

3DSinghVFX:
> @3DSinghVFX I am not familiar with the option, can you demonstrate that for me or point me in the right direction?

Okay. In the Stroke, we can set the _Line Width_. However, if we want to offset the line width of all the strokes per layer that we can do with _the Line Change_ option in the Layers. Is that make sense?

OmarEmaraDev:
I think we can leave it as is, one can think of it as a thickness in the space of the layer which have a thickness itself.

3DSinghVFX:
Peek 2022-03-26 17-51

3DSinghVFX:
> I think we can leave it as is, one can think of it as a thickness in the space of the layer which have a thickness itself.

Okay.

Issue #1885, opened 2022-10-20
On parsing midi files, add option to retrieve absolute note on and note off times in ticks


lgmventura:
I noticed, from using the node _Read Midi File_ and related notes and also by following the discussion on the [issue](https://github.com/JacquesLucke/animation_nodes/issues/1195) about creating these nodes, that the parser is converting the times _note on_ and _note off_ into seconds before outputting it. This means, it is dividing the absolute time in _ticks_ by the midi TPQ (**T**icks **P**er **Q**uarter note) so it gets the time in quarter notes. Then it runs through the list of tempos (quarter notes per minute) considering their starting times to calculate a given time in seconds.

### I would suggest adding following outputs:
- Node _Read MIDI File_:
  - List of tempos – each element would be a tempo (beats per minute or per second) and its starting time in **ticks** (int)
  - Ticks Per Quarter note (TPQ), which is a single integer for the MIDI file
- Node _MIDI Note Info_:
  - Note on in ticks (int)
  - Note off in ticks (int)
  - (rename current _note on_ and _note off_ to _note on in seconds_ and _note off in seconds_ or something like that)

### OR alternatively
- Node _Read MIDI File_:
  - List of tempos – each element would be a tempo (beats per minute or per second) and its starting time in **quarter notes** (float)
- Node _MIDI Note Info_:
  - Note on in unit quarter notes (float, already divided by TPQ)
  - Note off in unit quarter notes (float, already divided by TPQ)
  - (rename current _note on_ and _note off_ to _note on in seconds_ and _note off in seconds_ or something like that)

This would make it possible to create more complex animations – for example, taking the [Raw Note Information example](https://docs.animation-nodes.com/tutorials/midi/), we could use the tempos to pan the view with the speed proportional to the current tempo, without having to stretch the notes instead.

I could help you doing this, since I've already dealt with that in my [music visualization tool](https://github.com/lgmventura/MusicaVisual).

OmarEmaraDev:
Since Animation Nodes is mostly made for artists, we have to make sure that any feature we add or data we expose would be somewhat intuitive to understand and use.

Temp events, TPQ, and musical units are rather hard to understand, so I am not sure about exposing them or how we should do that. Maybe you can share an example node tree that would use those data (On dummy data just for investigation sake) just to get a feel or what a typical node tree would look like?

lgmventura:
[...]

PR #1890, opened 2022-12-07
MIDI parser retrieves additionally a list of tempo changes as well as note on and note off in quarter notes

lgmventura:
Solves #1885 

Improved flexibility when working with MIDI files. See the issue description for more info. Some UI improvements could be implemented. I will write a documentation on this.

### DIFF
--- a/animation_nodes/data_structures/__init__.py
+++ b/animation_nodes/data_structures/__init__.py
@@ -38,6 +38,7 @@ def importDataStructures():
 
     from . midi.midi_note import MIDINote
     from . midi.midi_track import MIDITrack
+    from . midi.midi_tempo_event import MIDITempoEvent
 
     from . action import (
         Action, ActionEvaluator, ActionChannel,
--- a/animation_nodes/data_structures/midi/midi_note.py
+++ b/animation_nodes/data_structures/midi/midi_note.py
@@ -31,10 +31,10 @@ class MIDINote:
     def evaluate(self, time, attackTime, attackInterpolation, decayTime, decayInterpolation, sustainLevel, 
         releaseTime, releaseInterpolation, velocitySensitivity):
 
-        value = evaluateEnvelope(time, self.timeOn, self.timeOff, attackTime, attackInterpolation, decayTime, decayInterpolation, sustainLevel)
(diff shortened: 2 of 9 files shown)


OmarEmaraDev:
Can you update the pull request with screenshots and examples for your changes?

lgmventura:
Hi, sorry for the late response here, but here we go:
- The data structure Midi Note now has timeOn and timeOff in two different units: seconds and quarter notes. Previously it was only in seconds.
- New data structure MIDITempoEvent in `midi/midi_tempo_event.py`
- ReadMIDIFileNode from `midi_file_reader.py` got a new output "MIDI Tempo Event List"
- MIDINoteInfoNode from `midi_note_info.py` got timeOn and timeOff in both seconds and quarter notes.
- New class MIDINoteInfoNode in new file in `midi_tempo_event_info.py` with following I/O:
  - Input "MIDI Tempo Event"
  - Output "Time in Quarter Note"
  - Output "Time in seconds"
  - Output "Tempo".
  - All these become a list by demand
- New classes MIDITempoEventSocket and MIDINoteListSocket in file `sockets/midi_tempo_event.py`.
- In file `utils/midi.py`, the time in quarter notes is included; the object MIDITempoEvent is added to the list of tempos instead of TempoEventRecord, which didn't include timeInQuarterNotes; a new short function timeInTicksToQuarterNotes was created to convert ticks, which is read out from the midi file, into quarter notes; the class NoteOnRecord has now time in ticks and time in seconds; when a noteOff event is found, the note's start and end time will be stored in seconds and in quarter notes. At the end, the tempos will be also returned alongside with the tracks.

lgmventura:
An example setup is shown here:

Main programme: loading the midi, calling the subprogram and using the tempos to pan the screen accordingly.
https://user-images.githubusercontent.com/16710238/197834327-08314cb4-f4e5-4568-b943-dd2b9b28f47d.png

The second image shows the subprogram only:
https://user-images.githubusercontent.com/16710238/197834710-06549164-257f-4b11-b370-0bcba604d770.png

OmarEmaraDev:
Isn't it the case that each MIDI track may have its own tempo map for MIDI format 2? In that case, the proposed design that cover that situation. Maybe this information should be part of the MIDI track instead. I also don't think we should have a separate data type and socket for tempo events, it doesn't seem worth it.

PR #1907, opened 2023-02-20
Fix Transform Vector node.

3DSinghVFX:
I have fixed the Transform Vector node for the modification of input data.


OmarEmaraDev (changes requested):
Data is not really modified as far as I can see. It is just that we use the same variable name for the input and output, so the data gets overwritten. It is best if the patch should just use unique names for both. `input_vectors` and `output_vectors`.

OmarEmaraDev (approved):

PR #1990, opened 2025-10-29
Change data path type check in attributes.py

rassvet-ii:
Object Data Path Output can't set scalar data due to wrong `pathBelongsToArray` type check.

It uses `pathBelongsToArray` to determine whether to use the index.
However, `_pathBelongsToArray` returns None if `len()` called with types such as float and int.
Since `createSetFunction` returns None when `pathBelongsToArray` returns None, it seems that it can't set Data Paths such as floats.

This PR fixes the problem by explicitly returning False for str, float, int, and bool cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Added
 
+- Added support for non-list types in *Object Data Path Output* node.
+
 ### Fixed
 
 - Fixed Sequence sockets for recent API change.
--- a/animation_nodes/utils/attributes.py
+++ b/animation_nodes/utils/attributes.py
@@ -65,7 +65,7 @@ def _pathBelongsToArray(object, dataPath):
     try:
         data = eval(f"object{dataPath}")
         if any(isinstance(data, ty) for ty in [str, float, int, bool]):
-            # Strings have len() but aren't arrays
+            # Strings and scalars shuldn't be indexed
             return False
         return len(data) > 0
     except:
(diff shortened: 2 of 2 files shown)


OmarEmaraDev (commented):
Can you update the comment under the line that was changed?

rassvet2:
Just to be clear, this PR fixes a bug that occurred in v4.5 and also work in v5.0. So I recommend merging this pr before #1991.

One concern is that `pathBelongsToArray` is also used in [animation_nodes/nodes/action/action_from_object.pyx#L27](https://github.com/JacquesLucke/animation_nodes/blob/master/animation_nodes/nodes/action/action_from_object.pyx#L27), and its behavior could change. But I don't know what the node does.
If that seems fine, this is also ready for merge.

OmarEmaraDev (changes requested):
It seems `isinstance` can already receive a tuple of types to check, so no need to use `any`.
Also, can you add an entry in the CHANGELOG.md file in the added section saying that this adds support for non-vector types in the data path node?

Can you also add the CHANGELOG entry for your other PR.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### Added
 
+- Added support for non-list types in *Object Data Path Output* node.
```
rassvet2: Initially, I wanted to fix the [`Cannot set attribute`](https://github.com/rassvet2/animation_nodes/blob/patch-1/animation_nodes/nodes/object/object_data_path_output.py#L26) error, so maybe the Fixed section is better, but it's also a feature addition……

PR #1991, opened 2025-11-01
Fix float/int sockets and sound sequences

rassvet-ii:
I'm lazy so I only fixed the errors that occurred when opening my project in 5.0. Therefore, I think there must be still many errors remaining for features I don't use.
This PR is not compatible with v4.5 or older because I use `set/get_transform` added in 5.0.
Sorry for dirty commit log. I'll squash and force push later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixed
 
 - Fixed Sequence sockets for recent API change.
+- Fixed some numerical inputs were not saved or loaded correctly.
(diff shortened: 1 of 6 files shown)


OmarEmaraDev:
Thanks for the PR. No need to support v4.5 once 5.0 is released, as we only support the latest version for now.

OmarEmaraDev:
Is this ready? It is still marked as a draft.
Don't worry about squashing, I will do it when merging.

